PR #16, opened 2014-11-21
Add xTrace and xProfile support

gitblit:
This PR exposes the `xTrace` and `xProfile` functionality to Java allowing you to instrument the SQLite engine.

http://www.sqlite.org/c3ref/profile.html

My C is pretty rusty so I'd appreciate a close inspection of the C code.

PR #19, opened 2015-01-09
Update build process to SQLite v3.8.7.4

gitblit:


xerial:
we also need to recompile native libraries. I'll work on lator

gitblit:
Hi @xerial I'm wondering what your plan is for #19 & #16.  I'd also like to see my changes from November in #13 released to MavenCentral.

BTW, SQLite is now at 3.8.8.2 so this PR is out-of-date.

xerial:
I haven't determined how to make code and native libraries synchronized. Except that there is no problem in accepting your recent changes.

Linux and win binaries can be built in Ubuntu by using cross compiler. For Mac OS, I need to start up VMs for building. 

An idea is allowing proceed commits in the develop branch without native libs, then at some timing I'll buld native libs then merge commits to mastet branch.

Any idea?

gitblit:
Here's what I'd like to see happen shortly.

1. Update to 3.8.8.2 (#21)
2. Release 3.8.8.2 with Savepoints support (#13)

Here is what I'd like to see after:

1. A thorough review of `xTrace` and `xProfile` (#16).  I think I have a leak or memory corruption in the JNI changes since it (intermittently) blows up in unit testing environments.

Longer-term I think that distribution of binary blobs needs to be reconsidered.

This JDBC driver is already 4MB (90% of that is native SQLite for different architectures) and that is without the recently FreeBSDx64 addition.  Perhaps the mainstream OS builds (Win64, OSX64, Linux64) should be bundled  into the core jar (like now) but all others (32-bit, BSDs, etc) should be distributed through add-on native library jars.

As for native lib build environments, unfortunately I think it's best for you to build all the blobs.  That makes it consistent.

Issue #25, opened 2015-03-29
createConnection shouldn't throw an exception if it isn't the appropriate driver.


jplmelanson:
According to the JDBC Driver spec:

> Attempts to make a database connection to the given URL. The driver should return "null" if it realizes it is the wrong kind of driver to connect to the given URL. This will be common, as when the JDBC driver manager is asked to connect to a given URL it passes the URL to each loaded driver in turn. The driver should throw an SQLException if it is the right driver to connect to the given URL but has trouble connecting to the database.

However, the implementation is currently not returning null if it's the wrong kind of driver:

```
public static Connection createConnection(String url, Properties prop) throws SQLException {
  if (!isValidURL(url))
    throw new SQLException("invalid database address: " + url);
  url = url.trim();
  return new SQLiteConnection(url, extractAddress(url), prop);
}
```

If this driver is loaded along other drivers inside the JVM, we stack traces like this:

> java.sql.SQLException: invalid database address: jdbc:postgresql:\/\/MYHOST\/DB at  org.sqlite.JDBC.createConnection(JDBC.java:111) at org.sqlite.JDBC.connect(JDBC.java:88) at java.sql.DriverManager.getConnection(DriverManager.java:571) at java.sql.DriverManager.getConnection(DriverManager.java:187)

xerial:
Available in sqlite-jdbc-3.8.9

PR #26, opened 2015-03-29
return null if unhandled url protocol

jplmelanson:
I don't think any additional checks are required, because the underlying methods will try to establish connection and throw a `SQLException` if it fails, and that is in accordance with the spec.


jplmelanson:
Proposed fix for #25

xerial:
Looks good. Thanks

PR #27, opened 2015-04-17
Implement the JDBC4 isValid() method

gitblit:


xerial:
Do we need to actually perform the query `select 1`?

gitblit:
That's a fair question.  I think the answer should be yes to handle the case where the db exists on a network file share and we've lost the connection to that share.

xerial:
OK. I got it.

PR #29, opened 2015-04-17
Fix case-mangling of URL parameters

gitblit:
One convenient use of the URL parameters is to specify the date format.  Unfortunately the URL parameter value was aggressively lower-cased which breaks many supplied date formats.  Now you can properly specify a case-sensitive date format.


xerial:
Looks good. Thanks

PR #41, opened 2015-07-24
Use Apache Commons FastDateFormat instead of SimpleDateFormat

sanelson2000:
SimpleDateFormat's .parse() method does not treat format Strings the same as its .format() method does.  This is a problem when dealing with dates created by ActiveRecord, which stores dates as strings with pseudo-microsecond precision in the following format: `yyyy-MM-dd HH:mm:ss.SSS000` (It's actually stored with millisecond precision, but `000` is always appended to the string.

SimpleDateFormat will properly format a date given this format string, but it fails when trying to parse a date.

The FastDateFormat utility provided by Apache Commons Lang 3.x does not have this issue.  As an added bonus, it is also faster and thread-safe.

This pull request uses FastDateFormat instead of SimpleDateFormat to get around this issue.

Unfortunately, it also has a few drawbacks.
1) make is broken because it can't find the commons-lang3-3.4.jar file.   This should be a simple fix, but I'm not familiar with make, so I'm not exactly sure where it's failing.
2) It requires the user to include commons-lang3-3.4.jar with alongside sqlite-jdbc.jar in order to work.  This may be a dealbreaker.  I'm not sure if there is a way to include the contents of the jar inside of the sqlite jar when it's built, though.  It seemed like it may be possible with Maven, but I'm not familiar with it, either.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ src/main/resources/org/ibex/nestedvm/*.o
 .idea
 *.iml
 atlassian-ide-plugin.xml
+.classpath
+.project
--- a/src/main/java/org/sqlite/core/CoreConnection.java
+++ b/src/main/java/org/sqlite/core/CoreConnection.java
@@ -10,13 +10,13 @@ import java.net.URL;
 import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
+
+import org.sqlite.date.FastDateFormat;
 
 import org.sqlite.SQLiteConfig;
 import org.sqlite.SQLiteConfig.DateClass;
@@ -56,7 +56,8 @@ public abstract class CoreConnection {
     public final DateClass dateClass;
     public final DatePrecision datePrecision; //Calendar.SECOND or Calendar.MILLISECOND
     public final long dateMultiplier;
-    public final DateFormat dateFormat;
+    public final FastDateFormat dateFormat;
+    public final String dateStringFormat;
 
     protected CoreConnection(String url, String fileName, Properties prop) throws SQLException
     {
@@ -66,7 +67,8 @@ public abstract class CoreConnection {
         SQLiteConfig config = new SQLiteConfig(prop);
         this.dateClass = config.dateClass;
         this.dateMultiplier = config.dateMultiplier;
-        this.dateFormat = new SimpleDateFormat(config.dateStringFormat);
+        this.dateFormat = FastDateFormat.getInstance(config.dateStringFormat);
+        this.dateStringFormat = config.dateStringFormat;
         this.datePrecision = config.datePrecision;
         this.transactionMode = config.getTransactionMode();
         this.openModeFlags = config.getOpenModeFlags();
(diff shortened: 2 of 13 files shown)


xerial:
@sanelson2000 
Thanks for the improvement. One of the design policy of sqlite-jdbc.jar is make it standalone-jar file. 

Could you consider the followings?:
- Could you extract a part of java files in commons-lang3-3.4 which is necessary to implement this feature, and add them to the src folder?  This usage is valid under APL2
- And also rename the package `org.apache.commons.lang3.time.FastDateFormat`  into 'org.sqlite.date` etc. so that sqlite-jdbc can be used with other libraries that use commons-lang3.

sanelson2000:
Thanks for the suggestions.  I'll look into it and update the PR when I get time.

sanelson2000:
@xerial
I've updated the PR with your suggested changes

xerial:
Looks good. Thanks for the contribution!

sanelson2000:
No problem.  Thanks for merging this!

sanelson2000:
@xerial Is there somewhere where I can download the latest build?  I'm seeing this when I try to load the driver that I compiled in my production environment on Windows... I'm guessing because I compiled it incorrectly...  Seems to be a problem with the native DLL

```
I, [2015-07-27T22:12:56.122000 #468]  INFO -- : Connecting to database: jdbc:sqlite:C:/consensus/Consensus_5.4.NIGHTLY.4953/consensus.db java.lang.UnsatisfiedLinkError: C:\Users\snelson\AppData\Local\Temp\sqlite-3.8.10.1-6038616b-b2e7-4e53-befe-d75da7b081fa-sqlitejdbc.dll: Can't find dependent libraries
E, [2015-07-27T22:12:56.247000 #468] ERROR -- : org.sqlite.core.NativeDB._open(Ljava/lang/String;I)V (Java::JavaLang::UnsatisfiedLinkError)
org.sqlite.core.NativeDB._open(Native Method)
org.sqlite.core.DB.open(org/sqlite/core/DB.java:161)
```

xerial:
Just released 3.8.10.2. Try this version.

2015/07/27 (月) 7:21午後 Scott Nelson <notifications@github.com>:

> @xerial <https://github.com/xerial> Is there somewhere where I can
> download the latest build? I'm seeing this when I try to load the driver
> that I compiled in my production environment on Windows... I'm guessing
> because I compiled it incorrectly... Seems to be a problem with the native
> DLL
>
> I, [2015-07-27T22:12:56.122000 #468]  INFO -- : Connecting to database: jdbc:sqlite:C:/consensus/Consensus_5.4.NIGHTLY.4953/consensus.db java.lang.UnsatisfiedLinkError: C:\Users\snelson\AppData\Local\Temp\sqlite-3.8.10.1-6038616b-b2e7-4e53-befe-d75da7b081fa-sqlitejdbc.dll: Can't find dependent libraries
> E, [2015-07-27T22:12:56.247000 #468] ERROR -- : org.sqlite.core.NativeDB._open(Ljava/lang/String;I)V (Java::JavaLang::UnsatisfiedLinkError)
> org.sqlite.core.NativeDB._open(Native Method)
> org.sqlite.core.DB.open(org/sqlite/core/DB.java:161)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/xerial/sqlite-jdbc/pull/41#issuecomment-125409684>.
>
-- 
Taro L. Saito, Ph.D.
<leo@treasure-data.com>
Software Engineer, Treasure Data, Inc.
http://xerial.org/leo

sanelson2000:
Awesome.  That works.  Thanks!

PR #43, opened 2015-07-29
Add a cache for ResultSet#findColumn

jcgay:
This is a performance boost when using #findColumn a lot.


xerial:
Looks good. Thanks for the improvement!

jcgay:
Wow that was fast, thanks !

Issue #49, opened 2015-09-06
Raspberry pi load broken 3.8.11.1


aschmois:
Java error dump on initialization
```` 
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGILL (0x4) at pc=0xac6f8cb0, pid=5678, tid=3056833648
#
# JRE version: Java(TM) SE Runtime Environment (8.0-b132) (build 1.8.0-b132)
# Java VM: Java HotSpot(TM) Client VM (25.0-b70 mixed mode linux-arm )
# Problematic frame:
# C  [sqlite-3.8.11.1-7a260e63-0de6-4553-b0c4-4b87818e1de0-libsqlitejdbc.so+0x5cb0]  _init+0x290b
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.sun.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#

---------------  T H R E A D  ---------------

Current thread (0x00b5a7d8):  JavaThread "main" [_thread_in_native, id=5679, stack(0xb62ea000,0xb633a000)]

siginfo:si_signo=SIGILL: si_errno=0, si_code=1 (ILL_ILLOPC), si_addr=0xac6f8cb0

Registers:
  r0  = 0x00000008
  r1  = 0xbee4b814
  r2  = 0x0000015c
  r3  = 0x00000000
  r4  = 0x00000008
  r5  = 0xbee4b814
  r6  = 0x00d6ecb8
  r7  = 0x00bbaca8
  r8  = 0xbee4b814
  r9  = 0xb6f16048
  r10 = 0x00000008
  fp  = 0xb63372fc
  r12 = 0xb63374d0
  sp  = 0xb6337240
  lr  = 0xac6f63ab
  pc  = 0xac6f8cb0
  cpsr = 0xa0000030

Top of Stack: (sp=0xb6337240)
0xb6337240:   ac6f63a5 b6eff210 00bbaca8 00000001
0xb6337250:   00000000 00d6ecb8 00bbaca8 bee4b814
0xb6337260:   00000008 b6eff36c 00d6ecb8 b6f16560
0xb6337270:   00d6ee14 b6f16048 00000003 00000003
0xb6337280:   00000001 b6f033b8 00000001 b6f16094
[...]

PR #55, opened 2015-10-02
added ARM native build file for raspberry pi 2

yonisolo:
as asked in #49 please find the native extension for armv7l


xerial:
Thanks for the contribution!

PR #64, opened 2015-10-26
Fix the build on 32-bit Windows

mkauf:
Bugfix for this issue: https://bitbucket.org/xerial/sqlite-jdbc/issues/191/problem-with-native-compile-on-windows-x86


xerial:
Merged. Thanks

PR #65, opened 2015-10-26
Fix JDBC3ResultSet.getBinaryStream()

mkauf:
Bugfix for this issue: https://bitbucket.org/xerial/sqlite-jdbc/issues/163/null-pointer-exception-on-null-input


xerial:
Good catch. Thanks!

Issue #61, opened 2015-10-22
Wrong UTF-8 decoding of Unicode Code Points higher then u+FFFF


0r1t0m:
When using UTF-8 encoding and insert a character where the unicode point is higher then u+FFFF. The Java ResultSet#getString() method return for this characters the same incorrect character. The Byte Representation seems to be correct.

I guess that that bug need to be in the C++ JNI Implementation https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java/org/sqlite/core/NativeDB.c#L503

## How to reproduce
* Linux x64
* org.xerial:sqlite-jdbc:3.8.11.2
* java version 1.8.0_45 - Java HotSpot(TM) 64-Bit Server VM (build 25.45-b02, mixed mode)

```bash
# HINTS:
# - LANG=en_US.UTF-8

echo "CREATE TABLE TEST (id INTEGER PRIMARY KEY, name CHARSET);" | sqlite3 /tmp/sqlite.db -batch
echo "PRAGMA encoding = \"UTF-8\";" | sqlite3 /tmp/sqlite.db -batch

# LATIN CAPITAL LETTER A
# http://unicode-table.com/de/0041/
# http://www.fileformat.info/info/unicode/char/0041/index.htm
echo -e "INSERT INTO TEST (name) VALUES ('\x41');" | sqlite3 /tmp/sqlite.db -batch

# Miao Letter Archaic Ma
# http://unicode-table.com/de/16F06/
# http://www.fileformat.info/info/unicode/char/16F06/index.htm
echo -e "INSERT INTO TEST (name) VALUES ('\xF0\x96\xBC\x86');" | sqlite3 /tmp/sqlite.db -batch
```

```java
Properties properties = new Properties();
properties.setProperty("characterEncoding", "UTF-8");
properties.setProperty("encoding", "\"UTF-8\"");
try (Connection connection = DriverManager.getConnection("jdbc:sqlite:/tmp/sqlite.db", properties)){
    Statement statement = connection.createStatement();
    ResultSet rs = statement.executeQuery("select name from TEST");
    while (rs.next()) {
        byte[] b = rs.getBytes("name");
        System.out.println("VALUE A = " + new String(b, java.nio.charset.StandardCharsets.UTF_8));

        String value = rs.getString("name");
        System.out.println("VALUE B = " + value);
    }
}
```

## Hints
[...]

PR #66, opened 2015-10-26
Fix the handling of supplementary characters (characters > U+FFFF)

mkauf:
JNI uses a modified UTF-8 encoding. For supplementary characters, an invalid UTF-8 sequence was written to the database, which resulted in interoperability problems. The solution is to avoid UTF-8 in the native code and use the UTF-16 functions of SQLite (where possible). SQLite will then convert the UTF-16 to standards-compliant, unmodified UTF-8.

This also fixes related bugs in JDBC3PreparedStatement and improves the "out of memory" handling in the native code.

Fixed Issues:
- https://bitbucket.org/xerial/sqlite-jdbc/issues/200/wrong-utf-8-decoding-of-unicode-code , same as https://github.com/xerial/sqlite-jdbc/issues/61
- https://bitbucket.org/xerial/sqlite-jdbc/issues/144/nativedbexec-throws-an-exception-without
- https://bitbucket.org/xerial/sqlite-jdbc/issues/84/bug-in-nativedbc-bind_1blob
- https://bitbucket.org/xerial/sqlite-jdbc/issues/70/setting-a-blob-in-prepstmt

### DIFF
--- a/src/main/java/org/sqlite/core/NativeDB.c
+++ b/src/main/java/org/sqlite/core/NativeDB.c
@@ -49,7 +49,7 @@ static void throwex(JNIEnv *env, jobject this)
     (*env)->CallVoidMethod(env, this, mth_throwex);
 }
 
-static void throw_errorcode(JNIEnv *env, jobject this, int errorCode)
+static void throwex_errorcode(JNIEnv *env, jobject this, int errorCode)
 {
     static jmethodID mth_throwex = 0;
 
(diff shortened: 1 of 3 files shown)


mkauf:
I think that the continuous integration system has not compiled the native library, but used an old version of the native library instead. A new unit test that I have written fails because the old native library has been used.

xerial:
OK. I will prepare native binaries rebuilt with this fix.

xerial:
Thanks for the proper error handling and the improvement of the query execution by using new API of SQLite.

mkauf:
Thank you for merging!

jberkel:
Just ran into this bug, glad it's already been fixed. However there doesn't seem to be a 3.9.1-SNAPSHOT version on sonatype, I only found 3.9.0.

PR #68, opened 2015-10-28
support JDBC 4 unwrap-ing

kares:
mostly useful on Connection instances but the rest is :scissors: as well


xerial:
Merged. Thanks!

PR #69, opened 2015-11-03
support closeOnCompletion

kares:
+ missed unwrapping test for previous PR

### DIFF
--- a/src/main/java/org/sqlite/jdbc4/JDBC4Statement.java
+++ b/src/main/java/org/sqlite/jdbc4/JDBC4Statement.java
@@ -32,6 +32,18 @@ public class JDBC4Statement extends JDBC3Statement implements Statement {
         return closed;
     }
 
+    boolean closeOnCompletion;
+
+    public void closeOnCompletion() throws SQLException {
+        if (closed) throw new SQLException("statement is closed");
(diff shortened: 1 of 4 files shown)


src/main/java/org/sqlite/jdbc4/JDBC4Statement.java
```diff
@@ -20,9 +20,28 @@ public boolean isWrapperFor(Class<?> iface) {
...
+        closed = true; // isClosed() should only return true when close() happened
+    }
+
+    public boolean isClosed() {
+        return closed;
+    }
+
+    boolean closeOnCompletion;
+
+    public void closeOnCompletion() throws SQLException {
```
xerial: Please add `@Override` to closeOnCompletion() and isCloseOnCompletion()

xerial:
Calling closeOnCompletion etc. breaks jdk6 compatibility. We may need to use reflection for calling these methods.

xerial:
Fixed the test code so that it can work in JDK6. 
https://github.com/xerial/sqlite-jdbc/commit/3123878737d095cbf0478f35a9caea53268913b0

Thanks for the contribution!

kares:
right, missed that - thanks!

PR #72, opened 2015-11-14
Test and fix bug in execute(String sql)

batterseapower:
Right now, if execute(String) fails when a transaction is open, then you aren't able to ever roll back the transaction and start again. Replacing "execute" with "executeUpdate" makes the problem go away.

I'm not 100% convinced my fix is right here, but it does make the test pass!


batterseapower:
Actually the problem affected prepared statements as well: I've pushed a fix to that too

xerial:
Looks good. Thanks for the fix!

PR #79, opened 2015-12-03
3.8.11.2 branch

gasup:
- "select cast(sum(×××) as decimal)"
- getColumnType return a INT type


gitblit:
Could you add a unit test for your column type change?

gitblit:
@gasup There are too many unrelated changes in this PR.  If you'd like to contribute your code, please open a PR for each change.  Thanks!

Issue #78, opened 2015-11-30
PreparedStatement setBytes()


danap:
Hello,

   Somewhere between Release 3.8.10.2 and current code the PreparedStatement setBytes()
has been changed that results in what appears to be twice the byte size being stored in the
database. I'm not sure if getBytes() is the issue. This results in corrupt BLOB type being
stored.

   A cursory look appears it maybe around the changes in the Kaufmann, (character > U+FFFF)
commit with the JDBC3PreparedStatement. I suppose it might also have to do with the native
library. Currently I'm using the Linux/64bit one. I have not tested on Windows or 32 bit machines.

danap.

xerial:
Thanks for reporting. This sounds serious. Could you give me a sample query to reproduce this issue? Thanks.

danap:
Hello,

   The test case code I have created to provide is not duplicating the behavior like the application
code. The only difference that i can determine at this time is the temporary HashMap that is used
to store the bytes for the user before election to dump to a file.

   Without changing anything other then the build jars, 3.8.10.1 to the current exempts the
behavior described above. I will not be able to investigate further for a day or two. It may be
a problem with the current code jar build created on my end. The environment is JRE 7.

   I would not classify this as a BUG at this time without further work on my end.

danap.

danap:
Hello,

   After further analysis, I have determined this is not related to the
PreparedStatements setBytes(). Rather in seems the ResultSet getBytes()
retrieval.

   I have created a test case that demonstrates the issue. I do not
believe this would be defined as a BUG, but a new inconsistency introduced
after build 3.8.10.1/2 in retrieving the field data more then once using
getString() then getBytes().

   The application uses what has been a general safe getString() to 
determine if the content is NULL. If not normally processing takes
place with reading a second time the content of the field with getBytes().
In so doing the returned bytes are 2X the size of the stored BLOB.

   I have ran into a similar issue with Oracle, HyperSQL, and MSAccess.
MSAccess only allows ResultSets to be read once, then they become null
I believe. None of the other databases my application supports restrict
or deform the data in retrieving a ResultSet more then once.

   The application has been updated to insure this is no longer an
issue, but I would say in the sqlite-jdbc it could snare other users
who may also read ResultSet more then once. I learned long ago what
you don't think the user will do in, will often be done.

Test Case: http://dandymadeproductions.com/temp/
SQLite_JDBC.java

   Thank you for your work with the sqlite-jdbc driver. By the way the
[...]

PR #82, opened 2015-12-12
ResultSet.getBytes(): Return strings in UTF-8

mkauf:
Fixes issue https://github.com/xerial/sqlite-jdbc/issues/78


mkauf:
I recommend pull request #84 instead of this one.

PR #84, opened 2015-12-16
Use UTF-8 functions of SQLite to avoid unexpected data type conversions

mkauf:
Also don't use modified UTF-8, see https://github.com/xerial/sqlite-jdbc/issues/61
Convert Java Strings to "unmodified" UTF-8 and vice versa.
Improve out-of-memory handling.

Fixes https://github.com/xerial/sqlite-jdbc/issues/78

### DIFF
--- a/src/main/java/org/sqlite/core/NativeDB.java
+++ b/src/main/java/org/sqlite/core/NativeDB.java
@@ -16,6 +16,7 @@
 
 package org.sqlite.core;
 
+import java.io.UnsupportedEncodingException;
 import java.sql.SQLException;
 
 import org.sqlite.Function;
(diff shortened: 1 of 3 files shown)


gitblit:
@mkauf Would you be willing to rebase this on master and provide some tests to confirm your code works like you think it should?

mkauf:
@gitblit I have updated the pull request and written a test.

gitblit:
@mkauf Sorry, about this but would you be willing to rebase this one more time?

mkauf:
@gitblit OK, done

src/main/java/org/sqlite/core/NativeDB.java
```diff
@@ -396,4 +397,22 @@ public native synchronized int restore(String dbName, String sourceFileName, Pro
     static void throwex(String msg) throws SQLException {
         throw new SQLException(msg);
     }
+
```
gitblit: These string util methods should move to `org.sqlite.util.StringUtils`.

gitblit: Ah.  Well that complicates the native implementation, doesn't it.

mkauf:
Thank you for merging!

PR #89, opened 2016-02-02
Building for new (released in 2016) and old SQLite versions

fatalmind:
Please look at the commit messages—they include plenty of explanation.


xerial:
Nice fix! Thanks!

PR #99, opened 2016-03-17
Implemented JDBC3ResultSet.isSigned(int).

RobertZenz:
JDBC3ResultSet.isSigned(int) does now return true if the type of the
column is either "NUMERIC", "INTEGER" or "REAL". In all other cases it
returns false.


gitblit:
:+1: Looks good.  Merging on behalf of @xerial.

PR #110, opened 2016-05-25
Fixed date/time handling

agalanin:
SQLite's internal date and time functions use UTC time zone. So we need to adjust date/time formatting and scanning in library to always use UTC.

### DIFF
--- a/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
@@ -17,6 +17,7 @@ import java.sql.Types;
 import java.util.Calendar;
 import java.util.GregorianCalendar;
 import java.util.Locale;
+import java.util.TimeZone;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.sqlite.core.CoreResultSet;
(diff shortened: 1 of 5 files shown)


src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
```diff
@@ -311,46 +312,21 @@ public Date getDate(int col) throws SQLException {
      * @see java.sql.ResultSet#getDate(int, java.util.Calendar)
      */
     public Date getDate(int col, Calendar cal) throws SQLException {
```
gitblit: It seems like this moves the driver away from the JDBC spec by ignoring the user supplied Calendar/Timezone information.

src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
```diff
@@ -311,46 +312,21 @@ public Date getDate(int col) throws SQLException {
...
     public Date getDate(String col) throws SQLException {
-        return getDate(findColumn(col), Calendar.getInstance());
+        return getDate(findColumn(col));
     }
 
     /**
      * @see java.sql.ResultSet#getDate(java.lang.String, java.util.Calendar)
      */
     public Date getDate(String col, Calendar cal) throws SQLException {
-        return getDate(findColumn(col), cal);
```
gitblit: (Duplicate) It seems like this moves the driver away from the JDBC spec by ignoring the user supplied Calendar/Timezone information.

src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
```diff
@@ -474,32 +450,7 @@ public Time getTime(int col) throws SQLException {
      * @see java.sql.ResultSet#getTime(int, java.util.Calendar)
      */
     public Time getTime(int col, Calendar cal) throws SQLException {
-        checkCalendar(cal);
```
gitblit: (Duplicate) It seems like this moves the driver away from the JDBC spec by ignoring the user supplied Calendar/Timezone information.

src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
```diff
@@ -547,35 +498,7 @@ public Timestamp getTimestamp(int col) throws SQLException {
      * @see java.sql.ResultSet#getTimestamp(int, java.util.Calendar)
      */
     public Timestamp getTimestamp(int col, Calendar cal) throws SQLException {
-        if (cal == null) {
```
gitblit: (Duplicate) It seems like this moves the driver away from the JDBC spec by ignoring the user supplied Calendar/Timezone information.

src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
```diff
@@ -589,7 +512,7 @@ public Timestamp getTimestamp(String col) throws SQLException {
      * @see java.sql.ResultSet#getTimestamp(java.lang.String, java.util.Calendar)
      */
     public Timestamp getTimestamp(String c, Calendar ca) throws SQLException {
-        return getTimestamp(findColumn(c), ca);
```
gitblit: (Duplicate) It seems like this moves the driver away from the JDBC spec by ignoring the user supplied Calendar/Timezone information.

gitblit:
I like the idea of wrangling the date/time type.  Looks like the driver is storing text formatted dates using the local timezone.  That was probably a mistake.  I'm concerned about backwards compatibility with this change.  Need more eyeballs on this review.

TuomasKiviaho:
I'm getting +2 hours with Europe/Helsinki when using `date_class=text`

gitblit:
@TuomasKiviaho A quick search tells me Helsinki is GMT+3 right now with DST.  Are you saying your dates are actually +5 compared to GMT (and therefore +2 compared to what you expected)?  Or are you saying they are GMT+2 ?  Or some other combination?  Fun with timezones...

TuomasKiviaho:
Hi, 14:01 saved as string became 16:01 when I applied the UTC timezone fix. I didn't investigate it any further than that. Works OK with machines set in UTC timezone. Below is the snipped that I'm using.

```
 case SQLITE_TEXT:
                try {
                    return new Timestamp(stmt.conn.timestampFormat.parse(db.column_text(stmt.pointer, markCol(col))).getTime());
                }
                catch (Exception e) {
                    SQLException error = new SQLException("Error parsing time stamp");
                    error.initCause(e);

                    throw error;
                }
```

> Fun with timezones...

Using org.threeten could be slightly less fun, but that's another story.

gitblit:
> Hi, 14:01 saved as string became 16:01 when I applied the UTC timezone fix.

Yeah, problems along those lines are what I was expecting.  Hmmm...

src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
agalanin: Unfortunately there are no sqlite datatype like 'timestamp with timezone' in postgresql, so timezone information is lost during saving in sqlite database.

agalanin:
Any news?

PR #111, opened 2016-05-26
Add QNX check to OSInfo

renjinghai:
for QNX, the so is not compatible  between different version. As far as I know, 6.5 and 6.6 are different. So add os.version to differentiate for QNX.


gitblit:
I don't see that QNX is a supported target in the `Makefile`.

gitblit:
Re-open if you'd like to provide QNX binaries.

PR #114, opened 2016-06-06
Added detection of the TIMESTAMP datatype.

RobertZenz:


gitblit:
This seems like a good idea.  Can you provide a unit test which demonstrates it's utility?

RobertZenz:
I've added the unit test.

This conflicts with #115, however I will happily rebase/rewrite the pull which is pulled last.

PR #118, opened 2016-06-10
Provide the native FreeBSD library also under `x86_64`.

### LINKED ISSUE #106: FreeBSD amd64 native code is ignored when FreeBSD x86_64 detected
When I try to use version `3.8.11.2` on 64-bit FreeBSD, I get the following error: `No native library is found for os.name=FreeBSD and os.arch=x86_64`. However, if I copy `org/sqlite/native/FreeBSD/amd64/libsqlitejdbc.so` to `org/sqlite/native/FreeBSD/x86_64/libsqlitejdbc.so` and repack the jar file, the library works.

I'd be happy to submit a PR that duplicates that binary to both paths, but I suspect the right answer is something in OSInfo.java, as noted in <https://bitbucket.org/xerial/sqlite-jdbc/issues/206/native-lib-for-freebsd-cant-be-resolved>.

jan0sch:
Using the jdbc driver under FreeBSD on an amd64 plattform results in the
following error:

    No native library is found for os.name=FreeBSD and os.arch=x86_64

A bug description and workaround are available in issue https://github.com/xerial/sqlite-jdbc/issues/106

This patch provides the native library under `x86_64` as well as under `amd64` and will fix #106.


gitblit:
Thanks for the PR @jan0sch, but the preferred fix was to move the binary, not duplicate it.  This was merged in #124.

PR #119, opened 2016-06-13
Use extended error codes

magnusr:
Enable extended error code reporting in sqlite, and new error enum values in SQLiteErrorCode.java. The extended error code is held in a new member of SQLiteErrorCode, so that the code member retains its old behaviour, so that  existing clients which depend on that are not affected. The message returned by toString does however change for errors for which there now are more detailed descriptions.

### DIFF
--- a/src/main/java/org/sqlite/SQLiteErrorCode.java
+++ b/src/main/java/org/sqlite/SQLiteErrorCode.java
@@ -24,6 +24,8 @@
 //--------------------------------------
 package org.sqlite;
 
+import java.sql.SQLException;
+
 /**
  * SQLite3 error code
  * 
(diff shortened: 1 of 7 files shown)


magnusr:
It looks like the TravisCI builds do not recompile NativeDB.c. Is that correct? If so, how can I get the PR builds passing?

magnusr:
The TravisCI build fix in PR #120 should make this one pass.

gitblit:
Looks like you'll need to rebase this PR on master in order for Travis to be happy.

src/main/java/org/sqlite/SQLiteErrorCode.java
```diff
@@ -63,9 +63,56 @@
...
+    SQLITE_IOERR_SHMSIZE(4874, " I/O error within xShmMap while trying to resize an existing shared memory segment"),
+    SQLITE_IOERR_SHMMAP(5386, " I/O error within xShmMap while trying to map a shared memory segment"),
+    SQLITE_IOERR_SEEK(5642, " I/O error while trying to seek a file descriptor"),
+    SQLITE_IOERR_DELETE_NOENT(5898, " The file being deleted does not exist"),
+    SQLITE_IOERR_MMAP(6154, " I/O error while trying to map or unmap part of the database file"),
+    SQLITE_IOERR_GETTEMPPATH(6410, " Unable to determine a suitable directory in which to place temporary files"),
+    SQLITE_IOERR_CONVPATH(6666, " cygwin_conv_path() system call failed");
 
     public final int code;
+    public final int extended;
```
gitblit: Can you explain why we need to maintain both `code` and `extended`?

magnusr: This was an incomplete attempt at maintaining backwards compatibility. I'll rework it, hopefully by Monday.

Just to make sure: we still want user code doing `SQLiteErrorCode.getErrorCode(ex.getErrorCode())` on an exception caused by an extended code like `SQLITE_BUSY_RECOVERY` get the non-extended instance (in this case`SQLITE_BUSY`) by default, right? Otherwise code that does that and compares the result with an instance of `SQLiteErrorCode` to determine how to handle the exceåtion would suddenly fail, so we should have some kind of setting for getting the extended code instances instaed, correct?

gitblit: Ok.  My understanding was incomplete.  I think your solution is probably fine.

magnusr: I have an updated version that adds an internal pragma for enabling/disabling use of extended result codes as SQLException vendor codes (but still uses the extended result code to build the message text for the exception).
Would you like it as a force-push that replaces the commit in this PR, or as a new commit on top of this broken one?

gitblit:
Go ahead and force-push to this PR so we can keep the conversation together.

I'm not sure it needs to be as complicated as you've implemented with a new PRAGMA.  Seems like we could always set the extended result code and have accessor methods to return the basic result code and extended result code.  Or have I missed something important?

magnusr:
We could, but changing what we use as vendor code could be seen as a breaking API change. If that's fine, then it's a lot simpler: just use extended result codes all the time. If not, then using extended result codes instead of legacy error codes as the vendor code must be opt-in.

So: does sqlite-jdbc  guarantee stability of vendor codes of SQLExceptions?

gitblit:
> does sqlite-jdbc guarantee stability of vendor codes of SQLExceptions?

I don't see that `sqlite-jdbc` has specified any stability guarantees, but that doesn't mean it shouldn't.  (Fixing #110 could be painful, though.)

What if we implemented a `SQLiteException` subclass of `SQLException` which returned the standard result code for `getErrorCode` and offered a `getExtendedCode` accessor for the more granular error?  Maybe this subclass should have the `SQLiteErrorCode` instance?

Or maybe the answer is using proper `SQLState` codes in conjunction with the extended result codes?

magnusr:
Something like this then?

I believe SQLState is a different thing. We should provide the exact sqlite result code to applications that are OK with database-specific code for figuring out exactly what went wrong. We should probably also provide SQLState codes for applications that are OK with general descriptions of what happened as long as they don't have to add database-specific code.

gitblit:
Yes, that looks good to me. @xerial Do you have an opinion?

gitblit:
Nice.  Thanks!

PR #120, opened 2016-06-13
Also build the JNI code in Travis builds

magnusr:
Makes it possible to get Travis builds for PRs that change NativeDB.c pass.


gitblit:
Super! Thanks for the fix.

Issue #103, opened 2016-04-01
Null query causes JVM to segfault


cdinger:
I ran into this in a Clojure program, but it looks like it's easily reproducible in Java. I'm not sure if this is at all related to [this BitBucket issue](https://bitbucket.org/xerial/sqlite-jdbc/issues/80/485-segmentation-fault-when-loading).

Big warning: I'm not a Java person. If any additional info is helpful, let me know.

`executeQuery(null)` example:
```java
import java.sql.*;

public class SegfaultOnNullQuery {
  public static void main(String args[]) throws Exception {
    DriverManager.getConnection("jdbc:sqlite::memory:").createStatement().executeQuery(null);
  }
}
```

run with:

```
$ javac SegfaultOnNullQuery.java
$ java -classpath ".:sqlite-jdbc-3.8.11.2.jar" SegfaultOnNullQuery
```

results in this segfault:

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x000000010e51b8c2, pid=63256, tid=5891
#
# JRE version: Java(TM) SE Runtime Environment (8.0_77-b03) (build 1.8.0_77-b03)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.77-b03 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# V  [libjvm.dylib+0x31b8c2]  jni_GetStringUTFChars+0x66
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/ding0057/play/sqlite-jdbc-segfault/hs_err_pid63256.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
#
Abort trap: 6
```

I'm running:

```
$ uname -a
Darwin Christians-MBP 15.3.0 Darwin Kernel Version 15.3.0: Thu Dec 10 18:40:58 PST 2015; root:xnu-3248.30.4~1/RELEASE_X86_64 x86_64
$ java -version
java version "1.8.0_77"
Java(TM) SE Runtime Environment (build 1.8.0_77-b03)
Java HotSpot(TM) 64-Bit Server VM (build 25.77-b03, mixed mode)
```

gitblit:
Fix merged.

PR #121, opened 2016-06-13
Don't crash the JVM on null queries

magnusr:
Issue #103


gitblit:
:+1: Looks good.  Merging on behalf of @xerial.

Issue #96, opened 2016-03-04
clearParameters() discards batch


dirk-olmes:
I'm trying to use the JDBC batch API in conjunction with PreparedStatements. The code looks like this:

``` java
stmt.clearParameters();
stmt.setInt(1, i);
stmt.addBatch(); 
```

But this approach throws

```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 2
    at org.sqlite.core.CorePreparedStatement.batch(CorePreparedStatement.java:110)
    at org.sqlite.jdbc3.JDBC3PreparedStatement.setInt(JDBC3PreparedStatement.java:291)
```

Debugging into this issue a bit, I find that the implementation of the clearParameters() method [discards the current batch](https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java/org/sqlite/jdbc3/JDBC3PreparedStatement.java#L39).

While I can change my code to just omit the call to clearParameters, I'd rather have the sqlite jdbc driver do the right thing. The same code works flawlessly against other JDBC drivers.

gitblit:
Would you be willing to open a PR with an appropriate unit test?

dirk-olmes:
Sure, let me look into this. Will take a few days, though as I'm quite busy right now.

gotson:
Closing as it seems resolved. Comment if necessary.

PR #129, opened 2016-07-04
Unit test to demo #96 - clearParameters() discards batch

dirk-olmes:
The test creates a table and inserts data using the batch API. In its current state, the test fails with the NPE described in issue #96.


xerial:
Closes this in favor of #323. Thanks

PR #136, opened 2016-07-12
Fix crash on addBatch with zero-parameter queries

### LINKED ISSUE #48: Using of prepared batches leads to NPE
I'm getting NPE when using `addBatch` because internal `batch` variable at `JDBC3PreparedStatement` is still `null` at this point. 

``` java
    public void addBatch() throws SQLException {
        checkOpen();
        batchPos += paramCount;
        if (batchPos + paramCount > batch.length) {
            Object[] nb = new Object[batch.length * 2];
            System.arraycopy(batch, 0, nb, 0, batch.length);
            batch = nb;
        }
        System.arraycopy(batch, batchPos - paramCount, batch, batchPos, paramCount);
    }
```
I snooped around `JDBC3Statement` and there the non-prepared version has this case already covered.

``` java
 public void addBatch(String sql) throws SQLException {
        internalClose();
        if (batch == null || batchPos + 1 >= batch.length) {
            Object[] nb = new Object[Math.max(10, batchPos * 2)];
            if (batch != null)
                System.arraycopy(batch, 0, nb, 0, batch.length);
            batch = nb;
        }
        batch[batchPos++] = sql;
    }
```

peterholak:
Currently, calling `addBatch()` on a prepared statement where the query has no parameters causes a NullPointerException. This PR fixes that.

Of course calling `addBatch()` this way is not very useful when using JDBC directly, but can occur in higher-level wrapper libraries that don't special-case the zero parameters case.

Other JDBC drivers (I checked MySQL and PostgreSQL) work just fine even when there are no parameters in the query. The JDBC spec doesn't mention this case specifically.

This probably also fixes #48. There is still no error thrown when some parameters are missing (and there probably should be), but that was the case before as well.

### DIFF
--- a/src/main/java/org/sqlite/core/CorePreparedStatement.java
+++ b/src/main/java/org/sqlite/core/CorePreparedStatement.java
@@ -26,6 +26,7 @@ public abstract class CorePreparedStatement extends JDBC4Statement
 {
     protected int columnCount;
     protected int paramCount;
+    protected int batchQueryCount;
 
     /**
      * Constructs a prepared statement on a provided connection.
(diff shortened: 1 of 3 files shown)


gitblit:
Great!  Thank you.

PR #149, opened 2016-09-22
Fix build problem for SQLite 3.14.0 onwards

fatalmind:
The preprocessor variable _SQLITEINT_H_ was renamed to SQLITEINT_H
with SQLite 3.14.0 (http://www.sqlite.org/src/info/5471aca0158851d3).

Only if neither is defined attempt to include sqliteInt.h.


gitblit:
Thanks for fixing that!

Issue #141, opened 2016-08-13
Exception in thread "main" java.lang.UnsatisfiedLinkError: org.sqlite.core.NativeDB._open(Ljava/lang/String;I)V


GKLuke:
`Exception in thread "main" java.lang.UnsatisfiedLinkError: org.sqlite.core.NativeDB._open(Ljava/lang/String;I)V
        at org.sqlite.core.NativeDB._open(Native Method)
        at org.sqlite.core.DB.open(DB.java:161)
        at org.sqlite.core.CoreConnection.open(CoreConnection.java:220)
        at org.sqlite.core.CoreConnection.<init>(CoreConnection.java:76)
        at org.sqlite.jdbc3.JDBC3Connection.<init>(JDBC3Connection.java:24)
        at org.sqlite.jdbc4.JDBC4Connection.<init>(JDBC4Connection.java:23)
        at org.sqlite.SQLiteConnection.<init>(SQLiteConnection.java:45)
        at org.sqlite.JDBC.createConnection(JDBC.java:114)
        at org.sqlite.JDBC.connect(JDBC.java:88)
        at java.sql.DriverManager.getConnection(DriverManager.java:664)
        at java.sql.DriverManager.getConnection(DriverManager.java:270)
        at SQLiteJDBC.main(SQLiteJDBC.java:11)
`
 I use  **a raspberry pi 3**, OS  raspbian (based on Debian 8 jessie), jdk and java VM below 
![qq 20160813123130](https://cloud.githubusercontent.com/assets/8219798/17641430/e45910c6-6151-11e6-8fb4-2367fa579b13.jpg)
Also tested openjdk8, and failed.Cry
I use **sqlite-jdbc-3.8.11.2.jar**.I tested this jar on my Mac(use jdk 1.7) and Windows(also use jdk 1.7), it worked fine.
my test code below

import java.sql.*;
public class SQLiteJDBC
{
  public static void main( String args[] )
  {
    Connection con = null;
    try {
      Class.forName("org.sqlite.JDBC");
      con = DriverManager.getConnection("jdbc:sqlite:/home/pi/Documents/arp.db");
    } catch ( Exception e ) {
      System.err.println( e.getClass().getName() + ": " + e.getMessage() );
      System.exit(0);
    }
    System.out.println("Opened database successfully");
  }
}
 What can I do next?

kwantumWare:
I'm having the same problem, but I'm using a Raspberry Pi vB+.

[...]

Issue #58, opened 2015-10-11
Raspberry pi load broken 3.8.11.2


DavidYYC:
This picks up from issue #49 (Raspberry pi load broken 3.8.11.1). I see the same error metioned in #49 using sqlite-jdbc versions 3.8.11.1 and 3.8.11.2.

I assume that the Raspberry Pi is using the armhf/libsqlitejdbc.so. That library is the same in 3.8.11.1 and 3.8.11.2. Was it a simple deployment glitch?

I built my own sqlite-jdbc-3.8.11.3-SNAPSHOT.jar on my Raspberry Pi (Model B 2nd Revision) and it works. No changes to the source code. I simply ran a build outlined in the READ_BUILD.md file. The armhf/libsqlitejdbc.so generated is very different from 3.8.11.1 and 3.8.11.2. It also works.

 $ cat /proc/cpuinfo
processor       : 0
model name      : ARMv6-compatible processor rev 7 (v6l)
BogoMIPS        : 2.00
Features        : half thumb fastmult vfp edsp java tls
CPU implementer : 0x41
CPU architecture: 7
CPU variant     : 0x0
CPU part        : 0xb76
CPU revision    : 7

Hardware        : BCM2708
Revision        : 000e
Serial          : 00000000bd950b8a
$ uname -a
Linux raspberrypi 4.1.7+ #817 PREEMPT Sat Sep 19 15:25:36 BST 2015 armv6l GNU/Linux

I'm happy to share my sqlite-jdbc-3.8.11.3-SNAPSHOT.jar if that would help.

yonisolo:
Hi thanks for your job, Can you share it please via filetea.me for example, i would like to test it on my pi 2 on raspbian.

DavidCGG:
Please see https://filetea.me/t1sxiMx3tzGTLeVTvOg1LdJUw for sqlite-jdbc-3.8.11.3-SNAPSHOT.jar.

yonisolo:
Sorry it doesn't work (blank page)

DavidCGG:
My fault. Try it now.

https://filetea.me/t1ss1E4Xc3qTdWKIXMO0Do7kg

yonisolo:
sorry, blank again, filetea seems to fail with jars...

yonisolo:
can you zip it with password and send it to this mail: lamri.yoni@gmail.com please, I know this method works because I already did it. Thx

yonisolo:
[...]

PR #151, opened 2016-09-22
Add docker based build of armv6 for Raspberry Pi

xerial:
For resolving UnsatisfiedLinkError #58, #141


xerial:
@gitblit Tried to build Raspberry Pi by using docker. Please merge this PR if you get any success report with this. I don't have Raspberry Pi, so have no way to test it by myself.

xerial:
Binaries for testing this is available from here: https://oss.sonatype.org/content/repositories/snapshots/org/xerial/sqlite-jdbc/3.14.2-SNAPSHOT/sqlite-jdbc-3.14.2-20160922.165637-1.jar

gitblit:
@xerial Awesome!  Btw, when I committed the update for 3.14.2 it was purely for changes to the build system.  We still need Linux binaries compiled against the older glibc.  I can't recall what base you were using.

xerial:
For linux x86 and X86_64, I'm using an old CentOS 5 (glibc 2.5) for supporting as many versions of Linux as possible since glibc is guaranteed to be backward compatible.

xerial:
@gitblit Actually I found there is no Linux x86(_64) binary difference with the ones I built 6 months ago https://github.com/xerial/sqlite-jdbc/commit/eb8fd076cf29f0a21dd76917587c0c7f5f6c8ede

gitblit:
@xerial How can that be??  Way different SQLite version!

xerial:
@gitblit Actually these native libraries are built against 3.14.2 :)

xerial:
I'm now double checking...

xerial:
It seemed my VMWare's file cache was broken. I pushed the new native libs. Sorry for the confusion

xerial:
Added scripts for building arvm5 (soft float), and armv6 (hardfloat) libs.

xerial:
Got a success report on Raspberry Pi. I'll merge it.

PR #153, opened 2016-09-23
Build windows librariles with docker cross compiler image

xerial:
- Confirmed library built with this works in Windows 8 :)


xerial:
@gitblit Let me merge this.

PR #154, opened 2016-09-24
Remove obsolete pure java SQLite class file

mkauf:


xerial:
Merged. Thanks for the cleanup.

PR #159, opened 2016-09-27
fixed NPE in getIndexInfo() for expression index

kassak:
If expression is used instead of column, then null-name is returned by sqlite


xerial:
Merged. Thanks!

PR #162, opened 2016-09-29
seems to make the latest code faster by reducing the back-n-forth between JNI and Java.

patcheng:
re: https://github.com/xerial/sqlite-jdbc/issues/160

On my MacBook Pro,
**3.8.11.2**
> Time to create & insert: 6872 ms
> Time to select: 8963 ms
> Time to select & get resultset values: 10279 ms

**3.14.2**
> Time to create & insert: 6654 ms
> Time to select: 11362 ms
> Time to select & get resultset values: 14068 ms
> 

**3.14.2.1-SNAPSHOT**
> Time to create & insert: 6527 ms
> Time to select: 8942 ms
> Time to select & get resultset values: 10936 ms
>

PR #165, opened 2016-10-01
Reduce the back-n-forth between JNI and Java even more

mkauf:
Pass Strings as UTF-8 byte arrays between the Java code and the native library
to improve the performance.

Also remove the declaration of value_bytes(), because this function has never
been implemented in the native code.

See https://github.com/xerial/sqlite-jdbc/pull/162


xerial:
@mkauf Thanks for the improvement. String.getBytes(UTF-8) has JVM specific optimization so it should be fast enough.

mkauf:
Thank you for merging!

Issue #164, opened 2016-10-01
Error handling in JDBC3PreparedStatement#execute()


jberkel:
I'm currently in the process of uprading from 3.9.1-SNAPSHOT to 3.14.2 which caused some tests failures in my application.

The failing tests verify some error conditions like writing to a read-only database. Instead of the expected error I get "The prepared statement has been finalized", raised in `NativeDB#reset()`, called from the finally block of `JDBC3PreparedStatement#execute()`

```java
try {
     resultsWaiting = db.execute(this, batch);
     success = true;
     return columnCount != 0;
 } finally {
     if (!success) db.reset(pointer);
 }
```

This got added in https://github.com/xerial/sqlite-jdbc/commit/df97e9163f504e3e2e7db4fc7918fa59bb71c6f4.

I suspect it should check if the reset call could fail before calling through to `db.reset`, or wrapping in another `try..catch` to avoid failing in the finally block.

Dunemaster:
I ran into the same problem. An example test

` @Test
    public void testExceptionCode() throws SQLException {
        // Arrange
        final String INSERT_INTO_PERSON_VALUES = "insert into person values(?, ?);";

        // Create table with unique constraint and insert a row
        stat.executeUpdate("create table person (id integer, name string, CONSTRAINT U_ID UNIQUE (id)); "
                +  "insert into person values(1, 'leo');");

        try
        {
            // try to insert a row with duplicate id
            try (PreparedStatement statement = conn.prepareStatement(INSERT_INTO_PERSON_VALUES))
            {
                statement.setInt(1, 1);
                statement.setString(2, "irma");
                statement.execute();
            }

        }
        catch (SQLException ex)
        {
            ex.printStackTrace();
            assertEquals(SQLITE_CONSTRAINT, ex.getErrorCode());
        }
    }`

jberkel:
@Dunemaster thanks for the testcase! I worked it into PR #166

jberkel:
closed w/ #166

PR #166, opened 2016-10-01
Check if pointer is valid before call to reset()

jberkel:
Test case and proposed fix for #164


xerial:
Nice catch and fix. Thanks!

jberkel:
@xerial You're welcome! I couldn't find any current snapshots, are they still being built?

xerial:
Just uploaded a snapshot here: https://oss.sonatype.org/content/repositories/snapshots/org/xerial/sqlite-jdbc/3.14.2.2-SNAPSHOT

jberkel:
@xerial thanks. with that snapshot i still get errors, addressed in #169

PR #167, opened 2016-10-02
Increase "Maximum Number Of Host Parameters In A Single SQL Statement"

jberkel:
Default is 999, Debian/Ubuntu and homebrew use 250000
This is a compile time setting which can only be lowered but not increased on runtime.

https://sqlite.org/limits.html#max_variable_number


xerial:
Thanks for the information. LGTM

PR #169, opened 2016-10-04
Check if pointer is valid before call to reset()

jberkel:
Similar to #166, but for `executeUpdate()`


xerial:
OK. Thanks for the fix.

xerial:
Set up a build step to automatically rebuild the jar after PR is merged. You can find the latest snapshot version from here:
https://oss.sonatype.org/content/repositories/snapshots/org/xerial/sqlite-jdbc/3.14.2.2-SNAPSHOT/

cc: @gitblit

jberkel:
@xerial great to have automatic snapshots. with this last PR all my tests pass now.

Issue #80, opened 2015-12-04
Library creates endless number of dll libraries in temp and they are not getting deleted


compengi:
Hello there,

I noticed this day that my Windows system drive is out of space after running a java program I wrote which imported 500k different files into a database.

Inspecting the issue using Windows's "Disk Cleanup" shows around 129GB of data in user's temp folder, which turns out to be this sqlite jdbc dll driver, creating 500k different jdbc dlls on each call of the java my java program taking, each one taking 720kb of disk space.

The file name looks like this: **sqlite-3.8.11.2-00a5dca3-f7cb-44b9-9014-de47cf377bd2-sqlitejdbc.dll** where **00a5dca3-f7cb-44b9-9014-de47cf377bd2** part is always different.

After some analysis:
The DeleteOnExitHook is called but the library is not unloaded (I checked using ProcessExplorer) so it never gets deleted.

Couple of questions though. Why do you create a random UUID each time a process is created, it's obvious that you can have multiple instances of an application and you can have multiple reads from the same library as well since the library is loading in read-only mode. You can always have several sqlite native libraries as well since they are identified by the version number, so why the random UUID?

OS: Windows 7 64-bit
Java: 1.8.0_66 Java(TM) SE Runtime Environment (build 1.8.0_66-b18)
sqlite-jdbc: 3.8.11.2

dfex55:
Any updates here?
We are experiencing the same problems with disks running out of space.
Scenario is a cron job using sqlite-jdbc.

compengi:
Hi,

I feel you, but till xerial figures it out, there are couple of things you can do.

1. Create your own class for extracting the library to the temp folder
2. Set both `System.setProperty` org.sqlite.lib.path to the temp folder and org.sqlite.lib.name to the extracted library name before you open a connection

This is how I'm doing it for now.

gitblit:
This is a duplicate of #87 and is an issue on Windows.  *NIX should allow deleting the shared library without a problem, AFAIK.

LeoGerber:
Could you please be  more precisely about that "*NIX" thing? Thnx

gitblit:
[...]

PR #173, opened 2016-10-11
Added Lck-File that is correctly deleted on JVM-Exit.

DaKaLKa:
I'm using sqlite-jdbc with windows and I have a big problem with #80.
As this is a bug in JVM (not able to delete On Exit dlls that were loaded) I added another file beside the extracted dll with the same name and the suffix ".lck". 
The file is correctly deleted on JVM exit. So now I can very easy create a cleanup method looking for all sqlite dll with missing lck-file and delete them in custom code.

Other ways (like looking if file is locked) was not always solving the problem (with multiple JVM-processes (e.g. 20)). 
(Maybe a cleanup method should also be added)

Maybe this would be useful for other people.
What do you think.

### DIFF
--- a/src/main/java/org/sqlite/SQLiteJDBCLoader.java
+++ b/src/main/java/org/sqlite/SQLiteJDBCLoader.java
@@ -180,7 +180,38 @@ public class SQLiteJDBCLoader {
                 // Delete the extracted lib file on JVM exit.
                 extractedLibFile.deleteOnExit();
 				extractedLckFile.deleteOnExit();
+				
+				// Add a Shutdown-Hook to remove other unused binaries (dll in temp without lck-file)
+				Runtime.getRuntime().addShutdownHook(new Thread()
+					{
+						@Override
+						public void run()
+						{
+						System.gc();
+						final String ver = org.sqlite.SQLiteJDBCLoader.getVersion();
+						
+						String tempFolder = new  	java.io.File(System.getProperty("java.io.tmpdir")).getAbsolutePath();					
+							java.io.File dir = new java.io.File(tempFolder);
+							java.io.File [] files = dir.listFiles(new java.io.FilenameFilter() {
 
+								@Override
+								public boolean accept(java.io.File dir, String name) 
+								{
+									return name.startsWith("sqlite-" + ver);
+								}
+							});
+							
+							for (java.io.File tempFile : files) 
+							{
+								if(tempFile.getName().indexOf(".lck") > 0) { continue; }
+								java.io.File lckFile = new java.io.File(tempFile.getName() + ".lck");
+								if(!lckFile.exists()){
+									tempFile.deleteOnExit();
+								}
+							}
+						   
+						}
+					}); 
                 if(writer != null) {
                     writer.close();
                 }
--- a/src/main/java/org/sqlite/SQLiteJDBCLoader.java
+++ b/src/main/java/org/sqlite/SQLiteJDBCLoader.java
@@ -29,6 +29,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
+import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


gitblit:
I think adding the `lck` file as a flag is a step forward, but now we need the corresponding search & delete functionality to make the `lck` file worthwhile.

DaKaLKa:
Ok, I've added some logic for automatic cleanup. The cleanup is made on exit of the Java-VM, so it has no impact of the startup-time.

src/main/java/org/sqlite/SQLiteJDBCLoader.java
```diff
@@ -170,9 +174,44 @@ private static boolean extractAndLoadLibraryFile(String libFolderForCurrentOS, S
...
-
+				extractedLckFile.deleteOnExit();
+				
+				// Add a Shutdown-Hook to remove other unused binaries (dll in temp without lck-file)
+				Runtime.getRuntime().addShutdownHook(new Thread()
+					{
+						@Override
+						public void run()
+						{
+						System.gc();
```
xerial: Why do you need to call `System.gc` here?

xerial:
Btw does it actually solve your problem? JVM has no guarantee to call the cleanup shutdown hook.

gitblit:
My initial thought had been to automatically cleanup on startup. Not sure what is best but there have been several reports of massive disk usage because the libs are not properly deleted.

xerial:
@gitblit Agreed. The cleanup phase should be on startup, rather than relying on shutdown hook.

src/main/java/org/sqlite/SQLiteJDBCLoader.java
DaKaLKa: Sorry, this was from an earlier test.

DaKaLKa:
Now, it's called on startup, before extracting the native library. 
For me, it's working lime a charm

xerial:
@DaKaLKa ok. Thanks. First, could you fix the indentation or remove unnecessary blank lines?

src/main/java/org/sqlite/SQLiteJDBCLoader.java
```diff
@@ -62,9 +62,46 @@
      * @return True if SQLite native library is successfully loaded; false otherwise.
      */
     public static boolean initialize() throws Exception {
-        loadSQLiteNativeLibrary();
-        return extracted;
+        // only cleanup before first extract
+		if(!extracted) {
```
xerial: weird indentation

src/main/java/org/sqlite/SQLiteJDBCLoader.java
```diff
@@ -62,9 +62,46 @@
...
+			cleanup();
+		}
+		loadSQLiteNativeLibrary();
+		return extracted;
     }
+	
+	
+	/**
+	* Deleted old nativ libraries e.g. on Windows this are not removed on VM-Exit (bug #80)
```
xerial: typo: native

src/main/java/org/sqlite/SQLiteJDBCLoader.java
```diff
@@ -62,9 +62,46 @@
...
+	*/
+	static void cleanup(){
+
```
xerial: remove this line

src/main/java/org/sqlite/SQLiteJDBCLoader.java
```diff
@@ -62,9 +62,46 @@
...
+		final String ver = org.sqlite.SQLiteJDBCLoader.getVersion();
```
xerial: unnecessary final and remove `org.sqlite.SQLiteJDBCLoder`

src/main/java/org/sqlite/SQLiteJDBCLoader.java
```diff
@@ -62,9 +62,46 @@
...
+		String tempFolder = new	java.io.File(System.getProperty("java.io.tmpdir")).getAbsolutePath();
```
xerial: Just `import java.io.File`

src/main/java/org/sqlite/SQLiteJDBCLoader.java
```diff
@@ -62,9 +62,46 @@
...
+	*/
+	static void cleanup(){
+
+		final String ver = org.sqlite.SQLiteJDBCLoader.getVersion();
+		String tempFolder = new	java.io.File(System.getProperty("java.io.tmpdir")).getAbsolutePath();					
+		java.io.File dir = new java.io.File(tempFolder);
+		java.io.File [] files = dir.listFiles(new java.io.FilenameFilter() {
+			@Override
+			public boolean accept(java.io.File dir, String name) {
+				return name.startsWith("sqlite-" + ver);
```
xerial: Use private variable e.g., `private final String prefix ="sqlite-" + ver` to minimize String generation cost.

src/main/java/org/sqlite/SQLiteJDBCLoader.java
```diff
@@ -62,9 +62,46 @@
...
+			}
+		});
+		
+		for (java.io.File tempFile : files) {
```
xerial: listFiles can return `null` so this might throw NullPointerException

src/main/java/org/sqlite/SQLiteJDBCLoader.java
```diff
@@ -62,9 +62,46 @@
...
+			if(tempFile.getName().indexOf(".lck") > 0) { continue; }
+			java.io.File lckFile = new java.io.File(tempFile.getName() + ".lck");
+			if(!lckFile.exists()) {
+				try{
+					tempFile.delete();
+				}
+				catch(SecurityException ex){
+
```
xerial: Swallowing exception is a bad habit. At least we should show the error message here.

DaKaLKa:
Hi xerial, I modified the code with your review wishes. 
Are there any code formatting specification that you are using. I tried the code formatting from eclipse (eclipde neon) but in doesn't seems to fit 100% to the code formatting you are using. Any settings I should be aware of in further code commits?

xerial:
Thanks. Currently we are not enforcing any specific format. But it will worth a try so that you can find the expected coding style. 

I'll check the code later.

xerial:
Merged to master. Thanks

PR #176, opened 2016-10-28
ResultSetMetaData::getTableName now returns empty string

### LINKED ISSUE #116: JDBC spec incompatibility: ResultSetMetaData::getTableName should not return null
I've noticed an inconsistency between the JDBC standard and your implementation.
JDBC spec for ResultSetMetaData::getTableName

```
/ * Gets the designated column's table name. 
* @param column the first column is 1, the second is 2, ... 
* @return table name or "" if not applicable * 
* @throws SQLException if a database access error occurs 
*/
```
A query, that contains a scalar subquery will return null for getTableName.

Example set up script and query
```
CREATE TABLE test (initial INT);
INSERT INTO test VALUES (1);

SELECT (1+initial) value FROM test
```

doe300:
Fixed #116
Signed-off-by:doe300 <stadeldani@web.de>


gitblit:
Super!  Thanks for helping us improve JDBC conformance.

PR #184, opened 2016-11-26
TypeMap implimentation added

geoolekom:
Added simpliest implimentation of typeMap getter and setter in order to avoid senseless exceptions. When some methods in snaq.db need to get typeMap to clean it up in connection pool, they call getTypeMap method. And it causes senseless crash, because they don't use content of this map, they just destroy it.


xerial:
OK. LGTM

PR #188, opened 2016-12-30
Add compile android arm native lib

psini:
This pull add the compile of sqlitejdbc for android arm. 
Thank you

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ NATIVE_TARGET_DIR:=$(TARGET)/classes/org/sqlite/native/$(OS_NAME)/$(OS_ARCH)
 NATIVE_DLL:=$(NATIVE_DIR)/$(LIBNAME)
 
 # For cross-compilation, install docker. See also https://github.com/dockcross/dockcross
-native-all: native win32 win64 linux32 linux64 linux-arm linux-armhf
+native-all: native win32 win64 linux32 linux64 linux-arm linux-armhf android-arm
 
 native: $(SQLITE_UNPACKED) $(NATIVE_DLL)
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -98,6 +98,9 @@ $(NATIVE_DLL): $(SQLITE_OUT)/$(LIBNAME)
 	@mkdir -p $(NATIVE_TARGET_DIR)
 	cp $< $(NATIVE_TARGET_DIR)/$(LIBNAME)
 
+android-arm: $(SQLITE_UNPACKED) jni-header
+	./docker/dockcross-android bash -c 'make clean-native native CROSS_PREFIX=/usr/arm-linux-androideabi/bin/arm-linux-androideabi- OS_NAME=Android OS_ARCH=arm'
```
xerial: A concern here is OS_NAME (Android) should match JVM system property `os.name` of DalvikVM for Android.

xerial:
To check the system is android or not, we need to run `System.getProperty("java.runtime.name", "").toLowerCase().contains("android")` within OSInfo.java

xerial:
@psini If you can fix OSInfo.java to add support for android OS, it would be great. But if you can wait for a week, I can do it on behalf of you.

psini:
I will try to make myself

Il 31 Dic 2016 06:59, "Taro L. Saito" <notifications@github.com> ha scritto:

psini:
Done make me know if i have to do other change

src/main/java/org/sqlite/util/OSInfo.java
```diff
@@ -162,6 +162,9 @@ else if (osName.contains("Linux")) {
         else if (osName.contains("AIX")) {
             return "AIX";
         }
+        else if (osName.contains("Android")) {
```
xerial: @psini Did you check the actual value of `os.name` JVM property under Android?
I think it's just Linux, so this code wouldn't work as expected

xerial:
I made changes in #193. Let me close this PR. Thanks.

PR #193, opened 2017-01-06
Add ARM v6, v7, android support

xerial:


xerial:
@psini I applied your change here.

Issue #74, opened 2015-11-25
Prepared statements close/finalize on constraint violations.


zurgug:
Use case:
We are creating a prepared statement outside a loop. Inside the loop we bind the values to each parameter and then execute the prepared statement. The loop is running over data being read from an external source.

Issue:
If the prepared statement has an exception (e.g. constraint violation) the rest of the loop produces "statement is not executing" errors.

Cause:
This is caused by the statement being finalized in DB.java

   execute(CoreStatement stmt, Object[] vals)

when step returns SQLITE_CONSTRAINT.

Workaround:
If an exception is thrown then recreate the prepared statement. This is not ideal as it means you can not easily use lambda's or try with resource.

LukeWoodward:
[This StackOverflow question](http://stackoverflow.com/questions/40184674/sqlite-preparedstatement-finalized/40185846#40185846) has what looks to be a similar issue (if not the same), in that it also relates to constraint violations and closing/finalization of prepared statements.  The workaround I suggested in my answer to that question was to downgrade sqlite-jdbc to 3.8.11.2, which correctly reports constraint violation exceptions.

LukeWoodward:
OK, so here's a repro for the exception I mentioned.  Only dependency is that the SQLite JDBC driver must be on the classpath:

```
package com.example;

import java.sql.*;

public class Main {
    public static void main(String[] args) throws Exception {
        DriverManager.registerDriver(new org.sqlite.JDBC());
        Connection conn = DriverManager.getConnection("jdbc:sqlite:");
        conn.createStatement().executeUpdate("CREATE TABLE test (a INT NOT NULL, PRIMARY KEY (a))");
        PreparedStatement stmt = conn.prepareStatement("INSERT INTO test (a) VALUES (1)");
        stmt.execute();
        stmt.execute();
    }
}
```

**Expected output:**
```
Exception in thread "main" java.sql.SQLException: [SQLITE_CONSTRAINT]  Abort due to constraint violation (UNIQUE constraint failed: test.a)
    at org.sqlite.core.DB.newSQLException(DB.java:890)
    at org.sqlite.core.DB.newSQLException(DB.java:901)
    at org.sqlite.core.DB.execute(DB.java:810)
    at org.sqlite.jdbc3.JDBC3PreparedStatement.execute(JDBC3PreparedStatement.java:50)
    at com.example.Main.minimalRepro(Main.java:36)
    at com.example.Main.main(Main.java:10)
```
[...]

PR #195, opened 2017-01-10
issue 74 add check valid pointer before reset

psini:
complete fix fro issue 74 add check valid pointer before reset in JDBC3PreparedStatement.java


xerial:
LGTM. Merged.

xerial:
This would solve #74

PR #203, opened 2017-02-08
Avoid any hardcoded file permissions

nguillaumin:
On Unix systems hardcoded file permissions should be avoided: When
creating files the umask of the current user should apply. Hardcoding
permissions prevents SQLite files to be created group-writable for
instance (umask = 0002)

To achieve this, configure the default SQLite permissions to 0666. This
grants read & write to all, then the umask applies, e.g. 0002, resulting
in permissions 0664.

Similarly, removed forcing permissions on the extracted library since
they don't account for the umask, and I think they're unnecessary anyway
(the library doesn't need to be executable for Java to load it).

Note that I'm not sure how that affect Windows & Mac. I think for
Windows it should be ok as it's using a different permission scheme, and
for Mac it should behave the same as Unix (except perhaps the removal of
the executable flag for the library, which I can't test).


nguillaumin:
There's some background [here](http://unix.stackexchange.com/questions/102075/why-are-666-the-default-file-creation-permissions) explaining why some/most basic standard Unix utilities sets 0666/0777 file permissions by default rather than hardcoding permissions.

xerial:
I have experienced a problem in loadind native library file if it misses x flag. That might be a bug of older versions of JVM, but that is the only reason why this code exists.

nguillaumin:
I guess I can revert the executable bit change if you want to keep it, although I really believe that's unnecessary. What matters to me more is the other change to do with the permissions of the actual DB file.

xerial:
@nguillaumin OK. Please keep only `-DSQLITE_DEFAULT_FILE_PERMISSIONS=0666` change. I don't want to include the change for the native library permission.

nguillaumin:
Ok, done.

xerial:
Thanks! Merged.

nguillaumin:
Thanks!

PR #220, opened 2017-05-03
makefile: use ssl for downloads from sqlite.org

mercutiodesign:
Since the sqlite.org supports ssl it would make sense to use that when downloading the zip archives.


xerial:
This change was no brainer. Thanks!

PR #221, opened 2017-05-03
BusyHandlerTest: use assertEquals instead of assertTrue

mercutiodesign:
Sometimes this test failed randomly on my machine. Since it used `assertEquals` it was not possible to determine what the difference was.


xerial:
Good fix. Thanks!

Issue #197, opened 2017-01-14
java.lang.UnsatisfiedLinkError: org.sqlite.core.NativeDB._open_utf8([BI)V


Eternity-Yarr:
I'm still getting this error with latest (3.16.1) version on ubuntu server with correct java tmpdir property.

```
java.lang.UnsatisfiedLinkError: org.sqlite.core.NativeDB._open_utf8([BI)V
	at org.sqlite.core.NativeDB._open_utf8(Native Method) ~[sqlite-jdbc-3.16.1.jar!/:na]
	at org.sqlite.core.NativeDB._open(NativeDB.java:71) ~[sqlite-jdbc-3.16.1.jar!/:na]
	at org.sqlite.core.DB.open(DB.java:174) ~[sqlite-jdbc-3.16.1.jar!/:na]
	at org.sqlite.core.CoreConnection.open(CoreConnection.java:220) ~[sqlite-jdbc-3.16.1.jar!/:na]
	at org.sqlite.core.CoreConnection.<init>(CoreConnection.java:76) ~[sqlite-jdbc-3.16.1.jar!/:na]
	at org.sqlite.jdbc3.JDBC3Connection.<init>(JDBC3Connection.java:26) ~[sqlite-jdbc-3.16.1.jar!/:na]
	at org.sqlite.jdbc4.JDBC4Connection.<init>(JDBC4Connection.java:24) ~[sqlite-jdbc-3.16.1.jar!/:na]
	at org.sqlite.SQLiteConnection.<init>(SQLiteConnection.java:45) ~[sqlite-jdbc-3.16.1.jar!/:na]
	at org.sqlite.JDBC.createConnection(JDBC.java:114) ~[sqlite-jdbc-3.16.1.jar!/:na]
	at org.sqlite.JDBC.connect(JDBC.java:88) ~[sqlite-jdbc-3.16.1.jar!/:na]
	at org.apache.tomcat.jdbc.pool.PooledConnection.connectUsingDriver(PooledConnection.java:310) ~[tomcat-jdbc-8.5.6.jar!/:na]
	at org.apache.tomcat.jdbc.pool.PooledConnection.connect(PooledConnection.java:203) ~[tomcat-jdbc-8.5.6.jar!/:na]
	at org.apache.tomcat.jdbc.pool.ConnectionPool.createConnection(ConnectionPool.java:718) ~[tomcat-jdbc-8.5.6.jar!/:na]
[...]

PR #233, opened 2017-06-30
Fix issue #197: create lock before lib file, remove lib files by absolute path

niktsar:
We faced this problem running a few instances of app that use sqlite. Creating lock-file before lib allows to avoid concurrent creating/deleting on initializing. Test taken from @bertmaher 's repo, reproduces on MacOS 10.12.5. Extra tested on Win10

### DIFF
--- a/src/test/java/org/sqlite/SQLiteJDBCLoaderTest.java
+++ b/src/test/java/org/sqlite/SQLiteJDBCLoaderTest.java
@@ -33,8 +33,12 @@ import java.sql.DriverManager;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


src/test/java/org/sqlite/SQLiteJDBCLoaderTest.java
```diff
@@ -111,4 +115,31 @@ public void version()
     // System.out.println(SQLiteJDBCLoader.getVersion());
     }
 
+    @Test
+    public void test() throws Throwable {
+        ExecutorService pool = Executors.newFixedThreadPool(32);
+        for (int i = 0; i < 32; i++) {
+            final String connStr = "jdbc:sqlite:sample-" + i + ".db";
```
xerial: Could you use `target/sample-` path to avoid messing up the source code root?

xerial:
Merged. Thanks for the fix.

Issue #219, opened 2017-04-28
primary keys from DatabaseMetaData.getPrimaryKeys  are case-insensitive 


downgoon:
## a table in ``default.db``

``` sql
CREATE TABLE `FRIEND` (
    `USER_id`   INTEGER NOT NULL,
    `FRIE_id`   INTEGER NOT NULL,
    `degree`    INTEGER,
    primary key ( `USER_id`, `FRIE_id` )
);
```

## testing code

``` java
import java.sql.Connection;
import java.sql.DatabaseMetaData;
import java.sql.DriverManager;
import java.sql.ResultSet;

public class SqliteJdbcAutoIncr {

	public static void main(String[] args) throws Exception {
		Class.forName("org.sqlite.JDBC");
		// Connection conn = DriverManager.getConnection("jdbc:sqlite:auto_incr.db");
		Connection conn = DriverManager.getConnection("jdbc:sqlite:default.db");

		DatabaseMetaData dbmeta = conn.getMetaData();

		// Listing Tables
		String catalog = null;
		String schemaPattern = null;
		String tableNamePattern = null;
		String[] types = null;

		ResultSet tables = dbmeta.getTables(catalog, schemaPattern, tableNamePattern, types);
		while (tables.next()) { // for each table

			final String tableName = tables.getString(3);

			// Listing Columns in a Table
			String columnNamePattern = null;
			ResultSet columns = dbmeta.getColumns(catalog, schemaPattern, tableName, columnNamePattern);

			System.out.println("table: " + tableName);
			System.out.println("----------");
			
			while (columns.next()) {
				String columnName = columns.getString(4);
				// int columnType = columns.getInt(5);

				boolean isAutoIncrement = columns.getBoolean("IS_AUTOINCREMENT");
				String autoIncString = columns.getString("IS_AUTOINCREMENT");

				System.out.println(
						String.format("\tcol: %s isAutoIncrement: %s %s", columnName, isAutoIncrement, autoIncString));
			}

			System.out.println();
			
			String schema = null;
			ResultSet keys = dbmeta.getPrimaryKeys(catalog, schema, tableName);
			while (keys.next()) {
				String keyName = keys.getString(4); // COLUMN_NAME
				int keySeq = keys.getInt("KEY_SEQ");
				
				System.out.println("primary key: " + keyName + ", " + keySeq);
			}


		}

	}

}

```

## output result

```
table: FRIEND
----------
	col: USER_id isAutoIncrement: false NO
[...]

PR #246, opened 2017-08-06
Fixes case-sensitivity of primary keys - see issue #219

sualeh:
Removed unnecessary `toLowerCase()` when obtaining primary keys, and also added unit tests for mixed case primary key names.


sualeh:
See issue #219.

xerial:
OK. Thanks for the fix.

Issue #267, opened 2017-09-06
Driver does not allow proper (class) garbage collection


trohwer:
In deployments with dynamic classloaders (e.g. web container), the sqlite driver is not properly garbage collected. Even if one deregisters the driver explicitly from the DriverManager, the global jni references (see NativeDB.c)

static jclass dbclass = 0;
static jclass  fclass = 0;
static jclass  aclass = 0;
static jclass pclass = 0;
static jclass phandleclass = 0;

prevent the classes from being garbage collected. This can be a problem, since that keeps the whole classloader and all associated classes in memory. Another minor leak is the data associated with the deleteOnExit() calls on temporary files, that keep accumulating.

An approach to fix the first problem might be:
- provide native method to clear the static references
- add finalize method or register phantomreference to call this clear method, once sqlite JDBC Driver instance is garbage collected
- reference sqlite JDBC Driver instance from connections to prevent this before all connections are garbage collected

If there is interest in getting this fixed this way, I might work on it.

xerial:
Basically I ageee with your suggested direction. I'd appreciate a PR. Thanks

trohwer:
I found a less intrusive, more elegant solution (see pull request).

trohwer:
One issue with my originally suggested approach is for example, that the driver class org.sqlite.JDBC has a public constructor, so even if the original instance, that is registered with the DriverManager, is gced, users might just create another one. With the references cleared, this will yield an error on the second instance.

yigit:
the PR seems to fix the issue reported here as well:
https://issuetracker.google.com/issues/62473121

trohwer:
Ok, good to know.
Note, that there is still be a minor leak with the deleteOnExit. In practise this should not matter, as this adds only two Strings (for the file paths) per initialization (see DeleteOnExitHook). I have an idea to fix that (try to delete the files right after loadNativeLibrary call; on Unix this should delete the files; on Windows it won't delete the library, but the deleteOnExit apparently does not do this either).

yigit:
@xerial is there an ETA for a release w/ this fix? 
The impact of this issue is rather significant for people using Android's new persistence library.
We can possibly ship a binary w/ the modified Xerial build but we would prefer to have it as a maven dependency to avoid conflicts with other libs that might depend on Xerial.

yigit:
[...]

PR #268, opened 2017-09-09
Fix for class gc memory leak

trohwer:
see issue #267 ; I found a more elegant, less intrusive solution.
I change the global refs to weak global refs. Since these references are
all to class objects, they will be referenced by the classloader
and not go away too early. But they won't keep the classes from becoming unreachable.
I verified, that this solves the problem
with this test:
[test.txt](https://github.com/xerial/sqlite-jdbc/files/1289555/test.txt)
Before the change, I get 10 instances of each sqlite related class in memory, after I get none.
The local tests during build (32bit Linux only) yield no errors.

### DIFF
--- a/src/main/java/org/sqlite/SQLiteJDBCLoader.java
+++ b/src/main/java/org/sqlite/SQLiteJDBCLoader.java
@@ -194,8 +194,8 @@ public class SQLiteJDBCLoader {
         String extractedLibFileName = String.format("sqlite-%s-%s-%s", getVersion(), uuid, libraryFileName);
         String extractedLckFileName = extractedLibFileName + ".lck";
 
-        File extractedLckFile = new File(targetFolder, extractedLckFileName);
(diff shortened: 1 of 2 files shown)


trohwer:
The check failure on Oracle8 does not seem to be related to the change:

Makefile.common:4: Set JAVA_HOME environment variable
rm -rf target/sqlite-3.20.0-Linux-aarch64
mkdir: cannot create directory 'target/sqlite-3.20.0-Linux-aarch64': No space left on device
Makefile:57: recipe for target 'target/sqlite-3.20.0-Linux-aarch64/sqlite3.o' failed
make: *** [target/sqlite-3.20.0-Linux-aarch64/sqlite3.o] Error 1
make: *** [linux-arm64] Error 2
The command "./scripts/travis-deploy.sh" exited with 2.
Done. Your build exited with 1.
grep: write error: No space left on device

xerial:
Thank for the fix. It seems test error is unrelated to the code change.

Issue #269, opened 2017-09-09
Racy Initialization in SQLiteJDBCLoader


trohwer:
If one creates a number of connections concurrently initially, one gets errors like

Exception in thread "Thread-15" java.lang.UnsatisfiedLinkError: org.sqlite.core.NativeDB._open_utf8([BI)V
        at org.sqlite.core.NativeDB._open_utf8(Native Method)
        at org.sqlite.core.NativeDB._open(NativeDB.java:71)
        at org.sqlite.core.DB.open(DB.java:174)
        at org.sqlite.core.CoreConnection.open(CoreConnection.java:220)
        at org.sqlite.core.CoreConnection.<init>(CoreConnection.java:76)
        at org.sqlite.jdbc3.JDBC3Connection.<init>(JDBC3Connection.java:26)
        at org.sqlite.jdbc4.JDBC4Connection.<init>(JDBC4Connection.java:24)
        at org.sqlite.SQLiteConnection.<init>(SQLiteConnection.java:45)
        at org.sqlite.JDBC.createConnection(JDBC.java:114)
        at org.sqlite.JDBC.connect(JDBC.java:88)
        at java.sql.DriverManager.getConnection(DriverManager.java:664)
        at java.sql.DriverManager.getConnection(DriverManager.java:270)
        at Test3.run(Test3.java:26)
        at java.lang.Thread.run(Thread.java:748)

The (- or at least - one) problem is, that the method

initialize() in SQLiteJDBCLoader

can be entered concurrently (only loadNativeLibrary is covered  by synchonized)
and then one running instance can delete the extracted files of the other running instance.
The mechanism with the lck files is not reliable; the lock file is created after the dll file.

Here is a test program:

`import java.sql.*;

public class Test3 implements Runnable {
  static int N= 20;
  static java.util.concurrent.CyclicBarrier cb=
    new java.util.concurrent.CyclicBarrier(N);
  public static void main(String[] args) throws Exception {
    Class.forName("org.sqlite.JDBC");
    for (int i=0; i<N; i++) {
      (new Thread(new Test3(i))).start();
    }
  }

  int d;
  public Test3(int d) {
    this.d= d;
  }

  public void run() {
    String url= "jdbc:sqlite::memory:";
    try{
      cb.await();
      Thread.sleep(0,100*d);
      DriverManager.getConnection(url).close();
    } catch(Exception e) {
      throw(new RuntimeException(e));
    }
  }
}
`
[...]

PR #270, opened 2017-09-09
Fix for issue #269

trohwer:
see remarks in issue #269


xerial:
OK. The fix itself looks good to me. Let me retry the CI build.

Issue #145, opened 2016-09-05
Usage of sqlite-jdbc together with the relocation feature of the Maven shade plugin


fmannhardt:
I would like to use the sqlite in a project where I have to use the relocation feature of the Maven shade plugin to avoid clashes with different versions of the same library being loaded. The Maven shade plugin can relocate all Java classes to a different package name (e.g., org.sqlite to test.org.sqlite). 

Unfortunately, sqlite-jdbc seems to load some of its classes dynamically at runtime in such a way that it is impossible for the Maven shade plugin to adjust the references accordingly. Therefore, all projects that use sqlilte, for example [cqengine](https://github.com/npgall/cqengine/issues/86), cannot be "shaded" entirely.

I created a small standalone test that demonstrates the issue. The test runs the example code of sqlite-jdbc. The JUnit test runs successfull, but when executing the shaded artifact: sqlite[...]-shaded.jar I get the following error:

> java -jar target/sqlite-shade-test-0.0.1-SNAPSHOT-shaded.jar
> java.sql.SQLException: No suitable driver found for jdbc:sqlite:C:\Users[...]AppData\Loca\Temp\test_8871362628708427569.db
>        at java.sql.DriverManager.getConnection(Unknown Source)
>        at java.sql.DriverManager.getConnection(Unknown Source)
>        at test.SQLite.open(SQLite.java:21)
>        at test.SQLite.main(SQLite.java:28)`

I tried to load the correct shaded driver with:

```
        try {
            Class.forName("test.org.sqlite.JDBC");
        } catch (ClassNotFoundException e) {
        }
```

but that leads to a ClassNotFoundException.

Would it be possible to support this use case? 
[sqlite-shade-test.zip](https://github.com/xerial/sqlite-jdbc/files/454772/sqlite-shade-test.zip)

xerial:
@fmannhardt JDBC driver name `org.sqlite.JDBC` is stored in https://github.com/xerial/sqlite-jdbc/blob/master/src/main/resources/java.sql.Driver, 
so if you can overwrite java.sql.Driver file in the shade jar, it would work.

fmannhardt:
Thanks for the answer. I almost forgot about this issue since I found a different solution that does not require shading the JAR at all. 

I tried to overwrite the java.sql.Driver file in the JAR, but that did not work. Then, the next class that is not found is org.sqlite.core.NativeDB:

[...]

PR #272, opened 2017-09-18
update README - Hint for maven-shade-plugin

paulvi:
ref #145


xerial:
Good information. Thanks!

PR #275, opened 2017-09-25
fix regarding https://github.com/xerial/sqlite-jdbc/pull/184

mschroeder-dfki:
Based on the comment of illumi in the closed issue https://github.com/xerial/sqlite-jdbc/pull/184 I changed the code and added a test.


xerial:
OK. Thanks for the fix.

PR #294, opened 2017-11-09
Add support for encrypted databases

woehrl01:
This PR adds support for encrypted databases. 
Most encrypted database drivers configure there password via ```pragma key='abc'``` or ```pragma hexkey='AF..'```.

This change calls the ```pragma hexkey``` and a following verification statement, if a password is provided.

This is usefull if you provide a custom libary via:
```
-Dorg.sqlite.lib.path=.
-Dorg.sqlite.lib.name=sqlite_cryption_support.dll
```

### DIFF
--- a/src/main/java/org/sqlite/SQLiteConfig.java
+++ b/src/main/java/org/sqlite/SQLiteConfig.java
@@ -119,16 +119,26 @@ public class SQLiteConfig
         pragmaParams.remove(Pragma.DATE_PRECISION.pragmaName);
         pragmaParams.remove(Pragma.DATE_CLASS.pragmaName);
         pragmaParams.remove(Pragma.DATE_STRING_FORMAT.pragmaName);
+        pragmaParams.remove(Pragma.PASSWORD.pragmaName);
+        pragmaParams.remove(Pragma.HEXKEY_MODE.pragmaName);
 
         Statement stat = conn.createStatement();
         try {
             if(pragmaTable.containsKey(Pragma.PASSWORD.pragmaName)) {
                 String password = pragmaTable.getProperty(Pragma.PASSWORD.pragmaName);
                 if(password != null && !password.isEmpty()) {
-                    stat.execute(String.format("pragma key = '%s'", password.replace("'", "''")));
+                    String hexkeyMode = pragmaTable.getProperty(Pragma.HEXKEY_MODE.pragmaName);
+                    String passwordPragma; 
+                    if(HexKeyMode.SSE.name().equalsIgnoreCase(hexkeyMode)) {
+                        passwordPragma = "pragma hexkey = '%s'";
+                    }else if(HexKeyMode.SQLCIPHER.name().equalsIgnoreCase(hexkeyMode)) {
+                        passwordPragma = "pragma key = \"x'%s'\"";
+                    } else {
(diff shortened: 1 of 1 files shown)


developernotes:
Hi @woehrl01 

Any thoughts to adjusting your pull request to support both `PRAGMA key` and `PRAGMA hexkey` by chance? Both SQLite SEE and SQLCipher support keying through the `PRAGMA key` command whereas the `PRAGMA hexkey` is limited to a binary key in SEE. SQLCipher utilized `PRAGMA key` with a special binary format to distiguish a hex key.

woehrl01:
@developernotes I'd be happy to adjust this PR to have the best fit across most SQLite libraries with crypt support. So I think supporting both formats should be possible, but I'm not sure what would be a good way to decide which way to use. 

1. We could use a define like ```-Dorg.sqlite.lib.hexkey=true```
2. Use an additional value on the the file url ```jdbc:sqlite:database.db?hexkey=true```

My initial decision to use ```hexkey``` is to not worry about escaping and possible non ascii characters.

What are your thoughts about this?

developernotes:
Hi @woehrl01 

Thanks for your follow-up, this is a nice improvement. As an additional thought, what would you think of adjusting the following:

```
PASSWORD("password", null);
```

to something like:

```
KEY("key", null),
HEXKEY("hexkey", null);
```

With this approach, you would still be able to support keying a database with both SEE and SQLCipher using string-based password material. You would then have a separate branch for the `pragmaTable.containsKey(Pragma.HEXKEY.pragmaName)` specific to hex keys in SEE. SQLCipher databases would still utilize the `Pragma.KEY.pragmaName` for binary key material as we encode a prefix and suffix format around the string. For example, SQLCipher can use a hex key such as this:

```
PRAGMA key = "x'2DD29CA851E7B56E4697B0E1F08507293D761A05CE4D1B628663F411A8086D99'";
```

Thoughts?

woehrl01:
@developernotes Thanks for your idea! 

I'm not sure if this is could lead to a possible bad sql syntax or injection.

What I would definetly like to keep is the `password` property as this allows us to easy integerate in different other utilities.

I would like to even go a more different approach for the `hexkey`:

Something like `hexkey= false (default) | sse | cipher` which would take the password and use that as the key:

`hexkey = false` + `password=helloworld`  = `pragma key = 'helloworld'`

`hexkey = sse` + `password=2DD29CA851` = `pragma hexkey = '2DD29CA851'`

`hexkey = cipher` + `password=2DD29CA851` = `pragma key = "x'2DD29CA851'"`

Maybe these should be named a little bit differently. 
This could be easily extended for possible other libraries. 

How does this sound to you?

xerial:
@woehrl01 Thanks for the contribution. Personally I have no experience on using encrypted SQLite version, but the PR itself looks good to me. 

Let me know if you find a reasonable PRAGMA format. I'll merge the PR.

woehrl01:
Thanks @xerial ! I updated the PR according to my proposal @developernotes

woehrl01:
The failing travis seems unrelated

xerial:
Merged. When you have a time, adding a short usage note (in Usage.md file) for this feature would be good to memorize the usage and configuration.

PR #301, opened 2017-11-21
Fixes #300

### LINKED ISSUE #300: Implement getSQLKeywords() correctly in DatabaseMetaData
Implement getSQLKeywords() correctly in DatabaseMetaData

sualeh:


xerial:
Should we add a test case? Just a sanity test (using this function and check some keyword containment) is fine.

xerial:
@sualeh Could you add a more meaningful title to PR next time so that we don't need to move around linked tickets?

PR #304, opened 2017-11-21
Fixes #303

### LINKED ISSUE #303: getPrimaryKeys() returns columns with quoted names
Example database:

```sql
PRAGMA foreign_keys = ON;

CREATE TABLE "Deal" (
  "deal_source" TEXT NOT NULL,
  "deal_id" TEXT NOT NULL,
  "fo_deal_id" TEXT NOT NULL,
  PRIMARY KEY ("deal_source", "deal_id")
);

CREATE TABLE "DealCost" (
  "deal_source" TEXT NOT NULL,
  "deal_id" TEXT NOT NULL,
  "deal_cost_no" INTEGER NOT NULL,
  "fee_type" TEXT NOT NULL,
  PRIMARY KEY ("deal_source", "deal_id", "deal_cost_no"),
  FOREIGN KEY ("deal_source", "deal_id") REFERENCES "Deal" ("deal_source", "deal_id")
);
```

sualeh:


src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -2243,4 +2236,24 @@ public String toString() {
...
+    
+    /**
+     * Follow rules in <a href="https://sqlite.org/lang_keywords.html">SQLite Keywords</a>
+     * @param name Identifier name
+     * @return Unquoted identifier
+     */
+    private String unquoteIdentifier(String name) {	
+    	if (name == null) return name;
+    	name = name.trim();
+        if (name.length() > 2 && (
```
xerial: You can use a regular expression like `^(['"\[])(.*)\1$`, then extract group 2, but the current code might be easier to read.

xerial:
@sualeh Please add a simple test case to check the behavior so as not to break the functionality in upcoming releases.

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
xerial: So it's ok to leave this as is

sualeh:
@xerial - sorry, I forgot to commit the tests. I have pushed them to this pull request now.

xerial:
Thanks. Merged

Issue #317, opened 2018-01-09
Feature Request: Exposing C function sqlite3_load_extension to Java


Andy-2639:
I need to load a SQLite extension.
With sqlite-jdbc, I can enable the SQL function load_extension to achieve this. However, this is not recommended as it increases the attack surface when a SQL injection is possible.
(SQL injections should be avoided anyway but humans make mistakes ...)

SQLite provides the C function
int sqlite3_load_extension(sqlite3 *db, const char *zFile, const char *zProc,  char **pzErrMsg);
https://sqlite.org/c3ref/load_extension.html
to load extensions. This could be used to load extensions from Java but not from a SQL statement.

Loading of extensions can be enabled by C API and at the same time disabled by SQL queries.
https://www.sqlite.org/c3ref/enable_load_extension.html
https://www.sqlite.org/c3ref/c_dbconfig_enable_fkey.html (SQLITE_DBCONFIG_ENABLE_LOAD_EXTENSION)

xerial:
Although the direct call of sqlite3_load_extension is not supported, you can register some UDFs as shown in the examples in https://github.com/xerial/sqlite-jdbc/blob/master/src/test/java/org/sqlite/UDFTest.java#L309

And also we are pre-loading some extension functions (math related stuffs) around here:
https://github.com/xerial/sqlite-jdbc/blob/5105677bb66e0245bf76fc1175d0be3f30833c8e/Makefile#L58-L62

xerial:
I think we have two options:
 - If what you need to load is a commonly usable one, we can embed these extension 
 - Add your extension code loading step in the above Makefile, then build your own sqlite-jdbc
 - Adding a JNI interface to call sqlite3_load_extension. 

Anyway I don't have much interest in adding these options by myself, so PR based contribution is better.

Andy-2639:
I'm going to look into adding a JNI interface.

The Extension I want to load is a slightly modified Version of sqlite3_icu. (I renamed the function names so that the original functions are not overridden.)
I don't think that it should be liked statically with sqlite-jdbc because it needs ICU which has a data library which is several MB big.

xerial:
OK. I once tried to link ICU statically, but gave it up because of that reason you mentioned and  thedifficulty of building ICU for various architectures. If load_extention works through JNI, it would be good for your purpose.

Andy-2639:
#319

Andy-2639:
I created PR #319 .
[...]

PR #319, opened 2018-01-28
Load SQLite extensions via SQLite C-API

Andy-2639:
This allows to load SQLite extensions without enabling the SQL function load_extension.


bradh:
Is there a way forward on this PR?

StevenLooman:
I am interested in this as well. What is the way forward?

alun:
I found it's relatively easy to do without this change:

1. Enable load extension in the connection properties like here https://github.com/xerial/sqlite-jdbc/blob/f948552fa06d475c4f1e72fdc2f72cbb18269217/src/test/java/org/sqlite/JDBCTest.java#L26

Also possible in DBeaver:
Screenshot 2020-05-23 at 16 12 20

2. Use SQL to load the extension e.g. `SELECT load_extension('/path/to/extension/md5', 'sqlite3_extension_init');`

Note the shared library extension must be omitted.

P. S. I've compiled md5 from this source https://github.com/moisseev/sqlite-md5/blob/master/md5.c on macOS using clang CLI: 
```
cc -c md5.c
cc -dynamiclib md5.o -o md5.dylib
```

Andy-2639:
> I found it's relatively easy to do without this change:

Yes, it is. It just allows extensions - arbitrary code - to be loaded into your process by normal SQL queries or SQL injection attacks.
I mean, you have to take care of SQL injections anyway (prepared statements, ...) but humans make errors and enabling `enable_load_extension` increases the ability of an attacker in case of an possible injection attack IMO.

alun:
@Andy-2639 cool, I see you point. I'd say those parts of a program calling load_extension stuff shouldn't be exposed to a non-authorised user's input data. And adding more native calls perhaps enlarges the attacking surface by itself.
What about this function here https://github.com/xerial/sqlite-jdbc/blob/d4b19c11e1b569cdabce82823a2fad5927df983a/src/main/java/org/sqlite/core/NativeDB.java#L109
can it be used to disable the feature once all required extensions are loaded? Although it might be added after your PR.

Andy-2639:
@alun Yeah,
https://github.com/xerial/sqlite-jdbc/blob/d4b19c11e1b569cdabce82823a2fad5927df983a/src/main/java/org/sqlite/core/NativeDB.java#L109
should work. However, I don't see it exposed.

So I guess when
https://github.com/xerial/sqlite-jdbc/blob/d4b19c11e1b569cdabce82823a2fad5927df983a/src/main/java/org/sqlite/SQLiteConnection.java#L25
exposes `enable_load_extension` and
https://github.com/xerial/sqlite-jdbc/blob/d4b19c11e1b569cdabce82823a2fad5927df983a/src/main/java/org/sqlite/SQLiteConfig.java#L100
would return a `SQLiteConnection`, the following would be possible:

1. Create SQLite connection
2. Enable extension loading
3. Load extension
4. Disable extension loading
5. Run queries with user provided data (still properly escaping/quoting it - prepared statements(!) - to protect from sql injections)

Andy-2639:
I sketched a possible solution to disable extension loading after loading them with the `enable_load_extension` method: https://github.com/xerial/sqlite-jdbc/issues/317#issuecomment-922339859

gotson:
@Andy-2639 given this is 4 years old, it needs to be updated to the latest version. If you can do that i can review the code, i think it would be a good addition to this project.

Andy-2639:
Sorry @gotson, as this PR needs the native libs rebuilt and I currently have no Linux system available to do this, caring about this PR takes too much time for me.

I'm fine with the workaround described in https://github.com/xerial/sqlite-jdbc/issues/317#issuecomment-922339859 despite the nasty downcast.
In this PR, I really don't like the serialization I needed to store the extensions to load in the properties ...

gotson:
> this PR needs the native libs rebuilt and I currently have no Linux system available to do this,

Almost everything is built via Dockcross now, should work well on WSL on Windows too.

---

I will mark this as needing rework, if someone wants to pickup your work

PortDm:
Hi. Help me please. I use Hibernate with SQLite. I need using function lower with Russian symbols. Standard JDBC driver is not support one. I created SQLite ICU:

clang -DSQLITE_ENABLE_ICU -DSQLITE_ENABLE_MATH_FUNCTIONS -I/opt/homebrew/opt/icu4c/include sqlite3.c shell.c -o sqlite3 -L/opt/homebrew/opt/icu4c/lib -licuuc -licui18n.

But build JDBC driver didn't work out. Can you write me full instruction how me solution this problem?
 I don`t understand too how include extension with ICU in my hibernate project. Please help me!

PortDm:
I tried build JDBC driver next command:
make native-all SQLITE_OBJ/path/to/sqlite3-icu.dylib SQLITE_HEADER=/path/to/sqlite-amalgamation-3490000-3/sqlite3.h

magicprinc:
Example of amazing SQLite extension set: https://github.com/nalgeon/sqlean

PR #322, opened 2018-02-08
change compiler source version to 1.6

joshimoo:
since SQLFeatureNotSupportedException is documented to only be available since 1.6
https://docs.oracle.com/javase/7/docs/api/java/sql/SQLFeatureNotSupportedException.html


xerial:
OK. LGTM

PR #323, opened 2018-02-08
Fix prepared statement parameter handling

joshimoo:
please have a look at the commit messages for explanation
the current implementation does not verify that all parameters for the prepared
statement are set

I also front loaded the check into addBatch since once a statement/command is added it becomes immutable so therefore it doesn't make any sense to wait till execution for that check.

this allows to keep the memory of the bitset constant == paramCount since we only need to validate
the current statement/command since previous statements/commands are guaranteed ot be valid


joshimoo:
This fixes #96 and #129 is no longer needed
@xerial is there an issue with the pull request or just time constrained at the moment?
Thank you for your work.

xerial:
LGTM. Sorry about the delay and thanks for addressing the long standing issue!

Issue #277, opened 2017-10-13
ArrayIndexOutofBoundsException in Prepared Statements with Multiple SELECT Queries


necouchman:
I'm using this JDBC driver in the Apache Incubator Guacamole project, with MyBatis and Guice, to try to make the SQLite database available as an authentication back-end for the product.  In a couple of places in the code ([1], [2]), there are prepared statements passed to the JDBC drivers that contain multiple select statements from different tables, each terminated by a semicolon.  This allows the Java code to make a single method call that runs both queries and returns a single result map, and keeps the code simple.  It works perfectly fine for MySQL, PostgreSQL, and SQL Server, but generates this ArrayIndexOutOfBoundsException when used with this SQLite JDBC driver.

When the batch array is allocated in the CorePreparedStatement class, it's allocated based on the number of parameters.  However, when multiple select statements are used that expect the same parameters to be passed, there will be x * y number of positions, where x is the parameter count and y is the number of SELECT statements, which is what results in the index out of bounds exception.

I'm not sure if batch needs to be dynamically allocated or resized, or if the size that it needs to be can be computed from other available parameters to get the total position count, but the current method does not allow for multiple statements to be prepared properly.

[1] [https://github.com/apache/incubator-guacamole-client/blob/b8abcd6755f9d44bf9b32daaf240c1361811751d/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/resources/org/apache/guacamole/auth/jdbc/connection/ConnectionMapper.xml#L87](url)
[2] [https://github.com/apache/incubator-guacamole-client/blob/b8abcd6755f9d44bf9b32daaf240c1361811751d/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/resources/org/apache/guacamole/auth/jdbc/connectiongroup/ConnectionGroupMapper.xml#L90](url)

gotson:
Is this still happening on the latest version?

gotson:
Just checked this a bit more, it's not supported in the driver.

It seems [it would be possible](https://sqlite.org/forum/forumpost/f47ae41b4d336239) from the SQLite C API, which can return the unused portion of the SQL query.

PR #325, opened 2018-02-12
Allocate batch array to deal with multiple queries in a single prepared statement.

necouchman:
Took a stab in the dark at a fix for #277 - this attempts to correctly deal with the issue where multiple queries are used in a single prepared statement, and the `batch` array is not allocated to handle the parameters, resulting in the `ArrayIndexOutofBoundsException`.


necouchman:
Nevermind...clearly not quite working, yet.

mathcoder23:
Object[] => LinkedArrayList<Object>

PR #328, opened 2018-02-22
Increase SQLITE_MAX_MMAP_SIZE to 1Tb by default.

### LINKED ISSUE #327: Please set SQLITE_MAX_MMAP_SIZE to something Very Large on x86_64
It's currently 2Gb, which is basically nothing for a 64 bit system. We would like to mmap databases >20Gb. Given that this feature has to be explicitly enabled using a pragma mmap_size, there's no risk to non-mmap users from just bumping the limit up to 1Tb or something in a 64-bit address space. It's the responsibility of the user calling pragma mmap_size not to do something Really Dumb(tm).

shevek:
Fixes #327 by setting to 1Tb. Checks performed:
a) sqlite will never set the mmap size of a file to more than the file size unless advised of a potential growth.
b) sqlite stores the config parameter value as an i64 on all platforms, including 32-bit.
c) sqlite will restrict the mmap size to 31 bits if running on a platform where sizeof(size_t) < 8
So I think this is safe for 32 and 64-bit platforms. Practical testing on a large 64-bit platform will follow later today.


xerial:
@shevek Sorry about the late response. OK. Let's try this configuration.

xerial:
It seems this option is causing performance issue. Although I'm not 100% confident about this.

xerial:
Added this option again as this change was not the cause.

PR #342, opened 2018-05-30
Fixed exception during finalize when statement AND db is already closed

### LINKED ISSUE #336: Exception during finalize of Statement/PreparedStatement
Hi,
I have code that basically does this:
- Open DB Connection
- Create prepared statements and use them
- Close prepared statements
- Close DB Connection

However, I'm always getting exceptions when the next GC run occurs:
```
org.sqlite.SQLiteException: [SQLITE_ERROR] SQL error or missing database (Connection is closed)
```
The reason seems to be quite simple: Finalize of the prepared statements calls close(), which in turn calls CoreStatement.internalClose().
Internalclose will then check if the DB is already closed, and then throw the exception.

The simplest  (and IMO cleanest) solution would just be to switch these two checks in CoreStatement.java, method internalClose():
```
if (db.conn.isClosed())
            throw DB.newSQLException(SQLITE_ERROR, "Connection is closed");

if (pointer == 0)
            return;
```

b3nn0:
Fixes #336

### DIFF
--- a/src/main/java/org/sqlite/core/CoreStatement.java
+++ b/src/main/java/org/sqlite/core/CoreStatement.java
@@ -116,11 +116,10 @@ public abstract class CoreStatement implements Codes
     }
 
     protected void internalClose() throws SQLException {
-        if (conn.isClosed())
-            throw DB.newSQLException(SQLITE_ERROR, "Connection is closed");
-
         if (pointer == 0)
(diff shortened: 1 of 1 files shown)


xerial:
Checking the presence of pinter at the beginning seems no harm. LGTM.

PR #343, opened 2018-05-30
Don't close prepared statements on constraint violation (Fixes #74)

### LINKED ISSUE #74: Prepared statements close/finalize on constraint violations.

b3nn0:
Fixes issue #74 .

### DIFF
--- a/src/main/java/org/sqlite/core/DB.java
+++ b/src/main/java/org/sqlite/core/DB.java
@@ -839,7 +839,7 @@ public abstract class DB implements Codes
             }
         }
 
-        int statusCode = step(stmt.pointer);
+        int statusCode = step(stmt.pointer) & 0xFF;
         switch (statusCode) {
         case SQLITE_DONE:
             reset(stmt.pointer);
@@ -850,6 +850,7 @@ public abstract class DB implements Codes
         case SQLITE_BUSY:
         case SQLITE_LOCKED:
         case SQLITE_MISUSE:
+        case SQLITE_CONSTRAINT:
             throw newSQLException(statusCode);
         default:
             finalize(stmt);


xerial:
Merged. Thanks!

patricklucas:
Apologies for commenting on this years-old change, but I ran into a problem upgrading from 3.23.1 to 3.31.1 and traced it here.

In my application, I am catching the `SQLiteException` thrown at the end of this switch statement and looking at its `resultCode` field, matching it against `SQLiteErrorCode.SQLITE_CONSTRAINT_PRIMARYKEY` to see if it represents a primary key uniqueness violation.

However, `SQLiteErrorCode.SQLITE_CONSTRAINT_PRIMARYKEY` is an "extended" error code, meaning when this change truncates the value to the lower byte with `& 0xFF`, the information in the exception is reduced to the basic error code `SQLiteErrorCode.SQLITE_CONSTRAINT`, breaking my error handling.

@b3nn0 / @xerial: Can either of you comment on whether this was intentional/necessary? Since `SQLiteException` supports extended error codes, I would imagine it's more useful to pass the non-truncated result code to it. Basically, my question is, would it be acceptable to change:

```java
        int statusCode = step(stmt.pointer) & 0xFF;
        switch (statusCode) {
```

to

```java
        int statusCode = step(stmt.pointer);
        switch (statusCode & 0xFF) {
```

?

b3nn0:
From my perspective that makes sense, yes. Don't think there is a real reason to truncate the error code for the exception - only for deciding if the statement should be closed.
If you rely on the status code, I think you should just make a pull request with your recommended change.

patricklucas:
Thanks for your input @b3nn0.

PR #350, opened 2018-06-14
Add support for update/commit/rollback event notifications

### LINKED ISSUE #98: Add support for data change notification callbacks
SQLite data change notification callbacks are functions registered on a connection that are invoked whenever a specific table row is modified.

See feature description here: http://www.sqlite.org/c3ref/update_hook.html

This is a valuable feature that should be fairly straightforward to expose at the Java layer.

Sketch of how this could be done:
- Define callback interface, with method that takes the rowid and mutation type (insert, delete, update)
- Add methods to `SQLiteConnection` to add/remove callback registrations on that connection
- Each `SQLiteConnection` maintains an internal bi-directional mapping from the `void *` handles passed to `update_hook.html()`, which could just be an incrementing counter value cast to `void *`, and the callback instance(s) registered on the corresponding rowid.

moxie0:
Add support for update/commit/rollback event callbacks. I'm not including the updated shared libraries in this PR, since I figured you all would want to build them yourselves.

Fixes #98

### DIFF
--- a/src/main/java/org/sqlite/SQLiteCommitListener.java
+++ b/src/main/java/org/sqlite/SQLiteCommitListener.java
@@ -1,0 +1,12 @@
+package org.sqlite;
+
+/**
+ * https://www.sqlite.org/c3ref/commit_hook.html
+ */
+
+public interface SQLiteCommitListener {
+
+  void onCommit();
+  void onRollback();
+
+}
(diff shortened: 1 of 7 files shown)


lalithsuresh:
I was about to start coding this feature up myself. Any chance we can get this merged?

xerial:
ok. Although we can improve the test code (e.g., wrapping with try ... catch to ensure removing event hooks at the end), it's better to add this feature first. Let me merge this.

PR #353, opened 2018-07-24
Fixed typo: feqture -> feature

Miha-x64:


xerial:
Thanks for the fix. Merged

PR #215, opened 2017-04-14
Fix concurrent issue in CoreResultSet closing

Yuhta:
fix #183

PR #358, opened 2018-08-03
Skip synchronized block in close() if ResultSet is already closed

### LINKED ISSUE #349: SQLite ResultSet close() blocks Finalizer Queue
Hi, I've noticed an issue with the SQLite ResultSet close() that was added as part of https://github.com/xerial/sqlite-jdbc/pull/215

I'm using xerial-sqlite 3.20.1. 

I have a thread that's running and taking a long time (looks like its not going to return which is its own issue).  However, while it's doing this `CoreResultSet.close()` call finalizer queue is blocked on the `db` object's lock because it's executing the query.  Code here: https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java/org/sqlite/core/CoreResultSet.java#L130

We have a number of different threads which each own their own SQLite connection.  This synchronized lock causes the whole service to stop finalizing any other objects, which eventually causes the whole service to die.

```
Thread 25408: (state = IN_NATIVE)
 - org.sqlite.core.NativeDB.step(long) @bci=0 (Compiled frame; information may be imprecise)
 - org.sqlite.core.DB.execute(org.sqlite.core.CoreStatement, java.lang.Object[]) @bci=108, line=811 (Compiled frame)
 - org.sqlite.core.CoreStatement.exec() @bci=47, line=75 (Compiled frame)
 - org.sqlite.jdbc3.JDBC3Statement.execute(java.lang.String) @bci=39, line=61 (Compiled frame)
```

```
Thread 8205: (state = BLOCKED)
 - org.sqlite.core.CoreResultSet.close() @bci=49, line=122 (Compiled frame)
 - org.sqlite.jdbc4.JDBC4ResultSet.close() @bci=6, line=34 (Compiled frame)
 - org.sqlite.jdbc3.JDBC3ResultSet.finalize() @bci=1, line=163 (Compiled frame)
 - java.lang.System$2.invokeFinalize(java.lang.Object) @bci=1, line=1270 (Compiled frame)
 - java.lang.ref.Finalizer.runFinalizer(sun.misc.JavaLangAccess) @bci=46, line=98 (Compiled frame)
 - java.lang.ref.Finalizer.access$100(java.lang.ref.Finalizer, sun.misc.JavaLangAccess) @bci=2, line=34 (Compiled frame)
 - java.lang.ref.Finalizer$FinalizerThread.run() @bci=45, line=213 (Compiled frame)
```

Is there a recommended way to fix this?  One thought is that I can try to adjust my code to avoid this issue by keeping all the ResultSet objects in scope until the end of the thread's life, so `finalize` will never be run in this scenario.

Thanks much for your thoughts on this!

tildedave:
Fixes #349

From profiling our app in production we noticed that our finalizer thread was periodically waiting on the synchronized lock on the DB (introduced in #215).  By the time our ResultSet hits the finalizer queue the ResultSet has already been closed, so there's no need to acquire the lock to reset the statement pointer.  As part of this I've removed what I believe to be dead code around checking for the nullability of the stmt member on the ResultSet.

### DIFF
--- a/src/main/java/org/sqlite/core/CoreResultSet.java
+++ b/src/main/java/org/sqlite/core/CoreResultSet.java
@@ -120,19 +120,18 @@ public abstract class CoreResultSet implements Codes
         cols = null;
         colsMeta = null;
         meta = null;
-        open = false;
         limitRows = 0;
         row = 0;
         lastCol = -1;
         columnNameToIndex = null;
 
+        if (!open) {
+            return;
+        }
+
         DB db = stmt.getDatbase();
         synchronized (db) {
-            if (stmt == null) {
-                return;
-            }
-
-            if (stmt != null && stmt.pointer != 0) {
+            if (stmt.pointer != 0) {
                 db.reset(stmt.pointer);
 
                 if (closeStmt) {
@@ -141,6 +140,8 @@ public abstract class CoreResultSet implements Codes
                 }
             }
         }
+
+        open = false;
     }
 
     protected Integer findColumnIndexInCache(String col) {
--- a/src/test/java/org/sqlite/ResultSetTest.java
+++ b/src/test/java/org/sqlite/ResultSetTest.java
@@ -140,4 +140,18 @@ public class ResultSetTest {
         resultSet.findColumn("test.id");
     }
 
+    @Test
+    public void testCloseStatement()
+        throws SQLException {
+        ResultSet resultSet = stat.executeQuery("select test.id from test");
+
+        stat.close();
+
+        assertTrue(stat.isClosed());
+        assertTrue(resultSet.isClosed());
+
+        resultSet.close();
+
+        assertTrue(resultSet.isClosed());
+    }
 }


tildedave:
Hi, I opened #349 a few months ago and found an "okay" work-around for the time being (details in ticket).  However we were still getting a large number of objects awaiting finalization (around the range of ~30k-60k objects) which was causing memory issues, especially around times which had higher than average load in our system.

I deployed this patch on release 3.20.1 to our production boxes and since then we've had less stalls on both the garbage collector and far fewer objects awaiting finalization, so I figured this patch might be helpful for others.

I added a test for the close behavior but I didn't get too specific on the particular scenario I was solving for because it requires a lot of access to class innards.  Let me know if there's anything I can do to make this easier for you.  Thanks much!

tildedave:
Actually pause on this, discovered some issues in production with this patch.  I'll update or close.

tildedave:
Okay, never mind, this patch is good.  (Issues are unrelated)

xerial:
Thanks for the fix. LGTM. Merged.

PR #364, opened 2018-09-05
Prevent null dates in prepared statements

### LINKED ISSUE #363: Null date crashes in prepared statement
Hello,
This is my first contribution on github.

Since commit 309a7d2167552376fb9a6c2157bde1234a690499 (20 Dec 2017), the prepared statements throw a `NullPointerException` when defining a `null` date, time or timestamp as parameter, while all other setters such as `setObject(int, Object)` or `setString(int, String)` allow `null` values.

Actually, the `org.sqlite.jdbc3.JDBC3PreparedStatement.setDate`, `setTime` and `setTimestamp` methods call `Date.getTime()` on the argument, even if it's `null`.
These changes where intended to take a custom timezone into account.

In fact, the `setDate(int, Date, Calendar)`, `setTime(int, Time, Calendar)` and `setTimestamp(int, Timestamp, Calendar)` methods should call `setObject(parameterIndex, null)` in case of a `null` value.

Here is a piece of code to reproduce the issue :
```
import java.sql.Connection;
import java.sql.SQLException;
import java.sql.Statement;

public class PreparedStatementIssue {

	public static void main (String[] args) throws ClassNotFoundException, SQLException {
		Class.forName("org.sqlite.JDBC");
		try (Connection connection = DriverManager.getConnection("jdbc:sqlite:");
				Statement statement = connection.createStatement()) {
			
			// Tablecreation
			statement.executeUpdate(
					"CREATE TABLE issue ("
					+ "date1 DATE,"
					+ "date2 DATETIME,"
					+ "date3 DATETIME"
					+ ")");
			
			// Prepared statement with issue
			try (PreparedStatement prepared = connection.prepareStatement(
					"INSERT INTO issue(date1, date2, date3) VALUES (?,?,?)")) {
				
				// Each of these parameter definitions is broken
				prepared.setDate(1, null);	// Null date shouldn't throw NPE
				prepared.setTime(2, null);	// Null time shouldn't throw NPE
				prepared.setTimestamp(3, null);	// Null timestamp shouldn't throw NPE
				
				prepared.executeUpdate();
			}
		}
	}
	
}
```

zachee54:
Fix #363 and add corresponding unit tests.

### DIFF
--- a/src/main/java/org/sqlite/jdbc3/JDBC3PreparedStatement.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3PreparedStatement.java
@@ -450,7 +450,11 @@ public abstract class JDBC3PreparedStatement extends CorePreparedStatement {
      * @see java.sql.PreparedStatement#setDate(int, java.sql.Date, java.util.Calendar)
      */
     public void setDate(int pos, Date x, Calendar cal) throws SQLException {
-        setDateByMilliseconds(pos, x.getTime(), cal);
+        if (x == null) {
+            setObject(pos, null);
+        } else {
+            setDateByMilliseconds(pos, x.getTime(), cal);
+        }
     }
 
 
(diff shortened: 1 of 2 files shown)


xerial:
Merged. Thanks

PR #366, opened 2018-09-06
Fix prep stmt test

### LINKED ISSUE #365: Exact floating point comparison test fails randomly
This unit test failed occasionnally :
```
Running org.sqlite.PrepStmtTest
Tests run: 33, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.047 sec <<< FAILURE!
getObject(org.sqlite.PrepStmtTest)  Time elapsed: 0.016 sec  <<< ERROR!
java.lang.NoSuchMethodError: org.junit.Assert.assertEquals(FFF)V
	at org.sqlite.PrepStmtTest.getObject(PrepStmtTest.java:285)
	...
```
This is due to the following line :
`        assertEquals(rs.getDouble(3), Double.MAX_VALUE, 0d);`
A double comparison should always give a non-zero delta, since double may be computed with approximated values.

I will open a PR to fix it.

zachee54:
Fix #365  and remove an unused local variable in PrepStmtTest.


xerial:
Thanks for the fix. Merged

PR #370, opened 2018-09-26
Implement Clob.getSubString() according to the jdbc specification

tseylerd:
The position was used as the starting index and the length as the end index. It's wrong according to the specification. The first argument is the start position. It starts with 1. The second argument is the length. The length must be greater than -1.


xerial:
Thanks for the fix. Merged.

PR #373, opened 2018-09-27
Fix build with current version of dockcross

urzds:
Compilers moved outside of $PATH, thus CROSS_PREFIX has to be adjusted.


xerial:
Thanks for the fix!

PR #375, opened 2018-10-02
Fix NPE in JDBC3ResultSet.getCharacterStream

### LINKED ISSUE #332: getCharacterStream doesn't implement JDBC correctly, throws NullPointerException instead of returning null
The Java Docs describe the return of getCharacterStream(int columnIndex) as:
<samp>
Returns:
a java.io.Reader object that contains the column value; if the value is SQL NULL, the value returned is null in the Java programming language.
</samp>
The code for JDBC3ResultSet.getCharacterStream gets a String and returns a StringReader:
https://github.com/xerial/sqlite-jdbc/blob/5105677bb66e0245bf76fc1175d0be3f30833c8e/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java#L269-L274
If the value of the column is SQL NULL, the StringReader will throw a NullPointerException. It should return a null value

snago:
Fixes #332


xerial:
Merged. Thanks for the fix

PR #376, opened 2018-10-06
Fix spelling of TransactionMode.DEFERRED (#316)

### LINKED ISSUE #316: Wrong enum value for TransactionMode DEFERRED
https://github.com/xerial/sqlite-jdbc/blob/5105677bb66e0245bf76fc1175d0be3f30833c8e/src/main/java/org/sqlite/SQLiteConfig.java#L770

should probably be `DEFERRED` instead of `DEFFERED`

snago:
The misspelled DEFFERED is still accepted for backwards compatibility.

Fixes #316


xerial:
oops... Thanks for the fix.

PR #377, opened 2018-10-06
Fix some tests

snago:
* BusyHandlerTest should now be deterministic.
* Removed BusyHandlerTest.testInterrupt as it probably didn't test
  what it was supposed to. The other two tests should be enough.
  (The "expected" SQLException was a "query does not return ResultSet"
  error because of executeQuery for insert instead of execute.)
* More verification in SQLiteJDBCLoaderTest.
* Make tests faster (run all tests in less than 2s instead of 20s).


xerial:
Nice. Unstable BusyHandlerTest has been annoying for me. Thanks!

PR #384, opened 2018-11-13
Add support for user-defined window functions

grddev:
This exposes another base class Function.Window alongside Function.Aggregate that allows implementing a user-defined window function using the new support introduced in SQLite 3.25


xerial:
Sorry for the delay. I've just merged this PR. Thanks for the great contribution!

Issue #385, opened 2018-11-18
since 3.23.1: java.sql.SQLException: assertion failure: param count (3) != value count (48)


qtxo:
In 3.23 some checks were added to the prepared statements parameters. 

https://github.com/xerial/sqlite-jdbc/commit/fd40cf60914d4063b889f0a08fdad21a4e957070#diff-9f9639750b991bb537666d8b5550d4e7

The problem is one of the commits instead of clearing the array instead created a mask as a Bitset paramValid and leave trailing null values in the array that are not going to be used except for a check that throws the exception:

```
if (params != vals.length) {
                throw new SQLException("assertion failure: param count (" + params + ") != value count (" + vals.length
                        + ")");
            }

for (int i = 0; i < params; i++) {
```

Note that the next for loop only makes use of the actual "params" so any extra null values in the array have no negative effect.

This is happening in complex batches of multiple insert, updates and selects. If I comment the throw, everything works fine like in 3.21 and lower versions.

gotson:
Is this still happening on the latest version?

gotson:
Was actually fixed in #386

PR #386, opened 2018-11-19
#385 Fix the params check when vals array is reused

qtxo:
https://github.com/xerial/sqlite-jdbc/issues/385


xerial:
Merged. Thanks for the fix.

PR #392, opened 2018-12-20
ENABLE_UPDATE_DELETE_LIMIT is not an option when using the amalgamation build

### LINKED ISSUE #391: Incorrect setting of ENABLE_UPDATE_DELETE_LIMIT compile time option in Makefile
The ENABLE_UPDATE_DELETE_LIMIT compile time options is set in the Makefile of the jdbc.
This should enable the use of LIMIT in update and delete statements.

Unfortunately this doesn't appear to work. Or more importantly, using 
`PRAGMA compile_options` makes SQLite report that it does, but it actually doesn't.

It's not a SQLite issue either (See [ENABLE_UPDATE_DELETE_LIMIT option](https://www.sqlite.org/compile.html#enable_update_delete_limit))

It states that:
>  [...] this option may only be used when the library is built from source, not from the amalgamation or from the collection of pre-packaged C files [..].

TLDR: When building from amalgamation, the ENABLE_UPDATE_DELETE_LIMIT option may not be set.

Not really a big deal, unless you actually want to support the compile option. But it confused me for a minute or two. Stumbled upon a stack overflow question about it too. So easy fix to avoid that in the future.

witmoca:
Simple single line update fixes #391 by removing the (unused) ENABLE_UPDATE_DELETE_LIMIT compile option.


xerial:
Yeah. I think this compile option is unnecessary. Thanks

magicprinc:
Please! Add limit support to delete!
https://www.sqlite.org/lang_delete.html

PR #396, opened 2019-01-24
Support temporary databases with no file name

zachee54:
When no database is specified, SQLite is supposed to create a [temporary database](https://www.sqlite.org/inmemorydb.html#temp_db), i.e. residing in memory but able to be flushed to disk under
memory pressure.
Actually, the driver forces SQLite to open an in-memory database
instead. These databases cannot be flushed to disk and might come out of
memory.


xerial:
I didn't know this functionality. Thanks for this PR!

Issue #400, opened 2019-02-14
FK_NAME in MetaData.getImportedPrimiaryKeys 


m1imueller:
If i have a table with foreign keys over more then one column. The result seams to be wrong. In this case the FK_NAME is empty.

So the following test will fail:

```java

    @Test 
    public void getImportedKeysMultipleColumns()throws SQLException{
        ResultSet importedKeys;
        stat.executeUpdate("CREATE TABLE PONE ( ID_FIRST_ONE INTEGER NOT NULL, "
                + "ID_SECOND_ONE INTEGER NOT NULL, "
                + "ADDITIONAL_ONE TEXT, "
                + "CONSTRAINT PONE_PK PRIMARY KEY (ID_FIRST_ONE, ID_SECOND_ONE) )");
        stat.executeUpdate("CREATE TABLE PTWO ( ID_FIRST_TWO INTEGER NOT NULL, "
                + "ID_SECOND_TWO INTEGER NOT NULL, ADDITIONAL_TWO TEXT, "
                + "CONSTRAINT PTWO_PK PRIMARY KEY (ID_FIRST_TWO, ID_SECOND_TWO) )");
        stat.executeUpdate("CREATE TABLE CHILD ( ID_CHILD INTEGER NOT NULL, "
                + "ID_CHILD_TO_ONE INTEGER NOT NULL, "
                + "ID_CHILD_TO_TWO INTEGER NOT NULL, "
                + "CONSTRAINT CHILD_PK PRIMARY KEY (ID_CHILD, ID_CHILD_TO_ONE), "
                + "CONSTRAINT PONE_FK01 FOREIGN KEY (ID_CHILD, ID_CHILD_TO_ONE) "
                + "REFERENCES PONE (ID_FIRST_ONE, ID_SECOND_ONE) ON DELETE CASCADE, "
                + "CONSTRAINT PTWO_FK02 FOREIGN KEY (ID_CHILD, ID_CHILD_TO_TWO) "
                + "REFERENCES PTWO (ID_FIRST_TWO, ID_SECOND_TWO) )");
        
        importedKeys = meta.getImportedKeys(null, null, "CHILD");
        
        assertTrue(importedKeys.next());
        assertEquals("PONE", importedKeys.getString("PKTABLE_NAME"));
        assertEquals("ID_FIRST_ONE", importedKeys.getString("PKCOLUMN_NAME"));
        assertEquals("PONE_PK", importedKeys.getString("PK_NAME"));
[...]

PR #401, opened 2019-02-14
Add support for getImportedKeys with ForeignKeys over more then one (#400)

m1imueller:
column (#400)

- Modification in JDBC3DatabaseMetaData
- Adding a UnitTest to DBMetaDataTest


xerial:
LGTM. Thanks for the fix!

PR #404, opened 2019-02-17
Corrects type affinity (#403)

### LINKED ISSUE #403: Wrong implementation of type affinity?
I'm using sqlite with hibernate, using https://github.com/gwenn/sqlite-dialect as dialect. I stumbled across an exception generated by hibernate `org.hibernate.MappingException: No Dialect mapping for JDBC type: 0` when running a select query on a table. JDBC type 0 is defined by `java.sql.Types.NULL`.

I reported this [issue](https://github.com/gwenn/sqlite-dialect/issues/14) to the owner of the dialect, including a fix for this issue on the dialect (I found it in a bug report at a fork of the dialect) and they pointed out that it's an driver issue and rejected my fix:

> Don't fix the dialect, fix the [driver](https://github.com/xerial/sqlite-jdbc/blob/a170d6d31db864316b042004040e356590358355/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java#L834).
> 
> https://docs.oracle.com/javase/6/docs/api/java/sql/ResultSetMetaData.html#getColumnType(int)
> 
> > Retrieves the designated column's SQL type.
> 
> Not the value type...
> 
> And https://sqlite.org/datatype3.html
> 
> > In order to maximize compatibility between SQLite and other database engines, and so that the example above will work on SQLite as it does on other SQL database engines, SQLite supports the concept of "type affinity" on columns. The type affinity of a column is the recommended type for data stored in that column.
> 
> 'NULL' is not listed among type affinities: 'TEXT', 'NUMERIC', 'INTEGER', 'REAL', 'BLOB'.

Source: https://github.com/gwenn/sqlite-dialect/issues/14#issuecomment-464126757

I currently applied the fix I found on the dialect, but I'm not happy to work this issue around.
Type affinity is great, but is the driver broken or the dialect?

witmoca:
Fixes #403 by defaulting the type affinity to NUMERIC instead of NULL

### DIFF
--- a/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
@@ -831,7 +831,7 @@ public abstract class JDBC3ResultSet extends CoreResultSet {
             }
         }
 
-        return Types.NULL;
+        return Types.NUMERIC;
     }
 
     /**
@@ -857,11 +857,11 @@ public abstract class JDBC3ResultSet extends CoreResultSet {
         case SQLITE_BLOB:
             return "BLOB";
         case SQLITE_NULL:
-            return "NULL";
+            return "NUMERIC";
         case SQLITE_TEXT:
         	return "TEXT";
         default:
-            return "NULL";
+            return "NUMERIC";
         }
     }
 


xerial:
Thanks for the fix!

Issue #382, opened 2018-10-25
Wrong error message when open fails


theogimonde:
As noted in #354, when `NativeDB._open_utf8` fails to open the database, a misleading error message is thrown ("The database has been closed"). A simple repeat is to try to open a path that points to a directory.

The underlying reason is that the `errmsg` call triggered by `throwex_errorcode` requires a valid database handle to be set, which is not true for the call in `_open_utf8`. A simple fix would be to always set the handle after opening the database and simply clear it just after the `throwex_errorcode` call on failures.

Attaching a patch with this fix and a test that fails without it: [patch.txt](https://github.com/xerial/sqlite-jdbc/files/2516482/patch.txt)

theogimonde:
@xerial Have you had the time to analyze this issue and the corresponding pull request? I would love to get the fix included eventually, but am unsure how I could contribute to that happening beyond what I've done so far.

PR #417, opened 2019-04-08
Fix issue #382: Wrong error message when open fails

theogimonde:
Creating a pull request of the patch for #382 in the hope of increasing its priority.

The fix always sets the database handle immediately after opening the database and simply clears it just after the `throwex_errorcode` call on failures, avoiding the issue that the `errmsg` call triggered by `throwex_errorcode` requires a valid database handle to be set, which is not true for the call in `_open_utf8` without this change.

Additionally a test is added that verifies that the correct more specific SQLite error message is returned when opening the database fails.


xerial:
Merged. Sorry about the late response

PR #428, opened 2019-05-16
427: wrong masking of sqlite system tables

Orfanik:
Changed sqlite system table mask.


xerial:
Thanks for the fix

Orfanik:
Hi, thx for accept it. :) 
I'll delete the fork.

PR #434, opened 2019-07-02
Update build process for JDK 10 (javah removal)

fatalmind:
javah has been superseded by the -h option added to javac in JDK 8.
The tool javah.exe was removed in Java 10.
(https://stackoverflow.com/questions/50352098/javah-missing-after-jdk-install)

Also clean output files generated by javac -h.

With this patch I was able to build with JDK 8 and JDK 11.

CI fails for JDK 7 — is it worth supporting that?


xerial:
We can drop java7 build later. Thanks for the fix

PR #441, opened 2019-07-23
Remove commented out code

changlinli:
Just general code hygiene since you're already using a VCS to keep code around.


xerial:
Good. Thanks for the fix

PR #451, opened 2019-09-10
Create libsqlitejdbc.so

### LINKED ISSUE #450: no native library for os.name=Linux and os.arch=ppc64le
when start the app on a IBM power9 server with centos 7.6,  an error occurs like:

no native library is found for os.name=Linux and os.arch=ppc64le

please see the attached picture
sqlite-jdbc-native-error

I see the newest jar ball has already include os.name=Linux and os.arch=ppc64 under the path "sqlite-jdbc/org/sqlite/native/Linux",  but miss a "ppc64le" directory.

maybe just make a directory called "ppc64le", then copy the lib file from "ppc64" directory to "ppc64le" directory. will this works ?

shenhuicc:
fix #450 for os.name=Linux and os.arch = "ppc64le"


rashmi-ibm:
@here any info when would this PR be merged and latest jar be released for power support.

xerial:
@rashmi-ibm Rather than merging a binary, we should use Makefile and OSInfo.java so that we can rebuild the native library for upcoming sqltie release. The current Makefile already supports ppc64 (with little endian), but os.arch name is just "ppc64". So we need to update OSInfo.java to map "ppc64le" arch type to "ppc64"

xerial:
To support ppc64le, we need to add a configuration to Makefiles and change OSInfo.java to detect little endian architectures.  I'll close this PR as we will not merge it.

PR #456, opened 2019-09-26
Improvements to custom functions (allow specifying nArgs and fix crash on aggregate functions for empty table)

### LINKED ISSUE #455: Allow specifying nArgs when registering a custom SQLite function
It would be slightly nicer to allow people to use SQLite's built-in argument check rather than expecting their functions to do the validation themselves.  Currently all SQLite functions get registered with a # arguments count of "-1".

tildedave:
This will allow creating and destroying custom functions with non-variable arguments.  Fixes #455.


tildedave:
I built the native libraries locally, should I include them here?

xerial:
Although test cases are missing, it's worth adding.

tildedave:
Thanks, I'll make sure I add these in the future.

PR #474, opened 2020-01-22
Fixed Maven URLs to be HTTPS

jonmclean:
Maven repository now requires HTTPS.  Fixed Maven URLs to be HTTPS


xerial:
Thanks for the fix!

PR #476, opened 2020-01-29
Remove sparse index checks

### LINKED ISSUE #393: Incorrect values not bound to statement
```
java.sql.SQLException: Values not bound to statement
	at org.sqlite.core.CorePreparedStatement.checkParameters(CorePreparedStatement.java:71)
	at org.sqlite.jdbc3.JDBC3PreparedStatement.executeQuery(JDBC3PreparedStatement.java:76)
```

Occurs when I have SQL like:

```
SELECT *
FROM test
WHERE column = ?2
```

at runtime I should be able to `bindLong(2, someValue)` then execute the statement but I get the above exception because it thinks there are 2 total arguments.

Could just be expected behavior, but at least using the native C api the above works

benasher44:
This fixes #393. This is correct/safe for a few reasons:

- Un-bound parameters are okay. sqlite interprets unbound parameters as `null` (near the bottom here: https://www.sqlite.org/c3ref/bind_blob.html).
- With this update, un-bound parameters in this library will be bound as null, but this is also okay because (see above) that's equivalent to binding null.
- This matches the implementations of other drivers:
-- requery/sqlite-android: https://github.com/requery/sqlite-android/blob/da119c1aa926975330ff727067fedc93e49da5f2/sqlite-android/src/main/java/io/requery/android/database/sqlite/SQLiteConnection.java#L1028
-- Android SQLite: https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/database/sqlite/SQLiteConnection.java#1088

Worst case, this defers to sqlite to error, and then an exception comes through that way.

FWIW, we've been using an internal build with this patch since I posted in #393 without issue.

Note: This contribution is made by Autodesk Inc. under the terms of the Apache 2.0 license.

### DIFF
--- a/src/main/java/org/sqlite/core/CorePreparedStatement.java
+++ b/src/main/java/org/sqlite/core/CorePreparedStatement.java
@@ -23,7 +23,6 @@ import org.sqlite.jdbc4.JDBC4Statement;
 
 import java.sql.Date;
 import java.sql.SQLException;
-import java.util.BitSet;
 import java.util.Calendar;
 
 public abstract class CorePreparedStatement extends JDBC4Statement
@@ -31,7 +30,6 @@ public abstract class CorePreparedStatement extends JDBC4Statement
     protected int columnCount;
     protected int paramCount;
     protected int batchQueryCount;
-    protected BitSet paramValid;
 
     /**
      * Constructs a prepared statement on a provided connection.
@@ -48,7 +46,6 @@ public abstract class CorePreparedStatement extends JDBC4Statement
         rs.colsMeta = db.column_names(pointer);
         columnCount = db.column_count(pointer);
         paramCount = db.bind_parameter_count(pointer);
-        paramValid = new BitSet(paramCount);
         batchQueryCount = 0;
         batch = null;
         batchPos = 0;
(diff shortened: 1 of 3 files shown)


benasher44:
Test need to be updated and added, but I'd like to get some maintainer buy in first, including suggestions for test coverage that is expected for this one 🙏

benasher44:
I updated some of the tests to account for the patch. Maybe that's sufficient test-wise?

benasher44:
Updated with this note in description (want to be explicit):
> Note: This contribution is made by Autodesk Inc. under the terms of the Apache 2.0 license.

benasher44:
@xerial any chance this could make it for one of the sqlite 3.31.x releases? Thanks for your consideration!

benasher44:
Thank you!!

xerial:
Thank you for waiting for a long time. Recently I'm not actively looking at PRs. It would be easy to imagine if you know that I've been maintaining this project for more than 10 years just as voluntary work 😅 For me, sqlite-jdbc just works for my small use cases. 

Now that GItHub Actions is widely available, it would be possible to automate the release process on CI just by merging a PR or adding a git tag, and asking people like you to review necessary PRs for making sqlite-jdbc usable. 

Let me know if you (or your friends) are interested in being a maintainer of sqlite-jdbc.

PR #479, opened 2020-02-18
Pass SQLITE_ENABLE_STAT4

mattsills:
SQLITE_ENABLE_STAT2 is deprecated (and is a no-op). Pass SQLITE_ENABLE_STAT4 instead.


xerial:
Thanks for letting me know that.

PR #480, opened 2020-03-10
Add Alpine support

johannespostler:
Please have a look at our extension that make the resulting jar files also work in an Alpine Linux container.

Changes:
- Changed to name of the compile SQLite binary when compiling for Alpine (as not to conflict with normal x86_64 linux)
- Added a small function in OSInfo.java that determines whether the code is running in an Alpine environment and uses the corresponding SQLite binary if this is the case.

 Looking forward to your comments!


xerial:
It looks great. Thanks for the contribution! I'll integrate alpine native lib.

PR #507, opened 2020-06-23
 Create dependabot.yml

thomas-nilsson-irfu:
Create a new dependabot.yml to help keep dependencies updated.

https://help.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates


gotson:
@thomas-nilsson-irfu would that generate the PRs for all deps, or only security deps ? Or just raise dependabot alerts ?

thomas-nilsson-irfu:
> @thomas-nilsson-irfu would that generate the PRs for all deps, or only security deps ? Or just raise dependabot alerts ?

Dear @gotson, adding this dependabot file will only result in the bot @dependabot opening new MR for all the dependencies listed in https://github.com/xerial/sqlite-jdbc/blob/master/pom.xml, if they have been updated from the specified version.

In MR from @dependabot there is also some extracted information from the dependencies, if they are providing it in things like release notes. I have seen some MR opened by @dependabot with release notes referencing security issues (CVE's etc), but typically @dependabot is indifferent in regards to why a new upstream dependent package was released...

The opened MR can be controlled via simple comments to `@dependabot`.

---

One example can be seen in https://github.com/irfu/irfu-matlab/pull/94.

gotson:
Given the project has zero dependencies, only test deps, i would prefer that it checks only for [security updates](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates).

We do however use `github-actions` ecosystem, and i would see some value in having those updates coming along.

PR #510, opened 2020-07-07
Add missing calls to "escape"

jwaataja:
There are two instances of Strings being directly inserted into queries executed on the database without escaping. This can potentially lead to vulnerabilities or crashes. Adds the two necessary `escape` calls.


xerial:
Thanks for the fix!

PR #512, opened 2020-07-08
Add support for platform Linux-Alpine/x86 (32bit)

stolzem:
I had to use sqlite-jdbc on a 32bit Alpine linux and it wasn't supported.
So I added support and then it worked fine.

Thought I'd share the Linux-Alpine/x86 support work - if you want to include it in the main lib.

To build it properly you'd have to create the docker image for Linux-Alpine/x86 via
`make docker-alpine-linux32`


gotson:
@stolzem i think that would be a good addition to the native libs. A lot has changed since your PR was made, if you are willing to update it I would happily merge that.

Note that we have changed the `OS_NAME` to `Linux-Musl`.

PR #526, opened 2020-07-30
Preserve extended result codes in SQLiteException

patricklucas:
Prior to 37278d1, extended result codes were preserved when throwing a
SQLiteException. This restores that behavior.

Refs #343


xerial:
Thanks for the fix! Merged.

PR #527, opened 2020-08-03
Added limit handling

BastouP411:
There are now new pragmas that makes the user able to set SQLite limits.
There parameters are:
  - limit_length: The maximum size of any string or BLOB or table row, in bytes.
  - limit_sql_length: The maximum length of an SQL statement, in bytes.
  - limit_column: The maximum number of columns in a table definition or in the result set of a SELECT or the maximum number of columns in an index or in an ORDER BY or GROUP BY clause.
  - limit_expr_depth: The maximum depth of the parse tree on any expression.
  - limit_compound_select: The maximum number of terms in a compound SELECT statement.
  - limit_vdbe_op: The maximum number of instructions in a virtual machine program used to implement an SQL statement. If sqlite3_prepare_v2() or the equivalent tries to allocate space for more than this many opcodes in a single prepared statement, an SQLITE_NOMEM error is returned.
  - limit_function_arg: The maximum number of arguments on a function.
  - limit_attached: The maximum number of attached databases.
  - limit_like_pattern_length: The maximum length of the pattern argument to the LIKE or GLOB operators.
  - limit_variable_number: The maximum index number of any parameter in an SQL statement.
  - limit_trigger_depth: The maximum depth of recursion for triggers.
  - limit_worker_threads: The maximum number of auxiliary worker threads that a single prepared statement may start.


pom.xml
```diff
@@ -91,8 +91,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.1</version>
         <configuration>
-          <source>1.6</source>
-          <target>1.6</target>
+          <source>1.8</source>
```
xerial, line 94: Should we change the target JVM version in this PR? If not, remove this change.

src/main/java/org/sqlite/SQLiteLimits.java
```diff
@@ -0,0 +1,27 @@
...
+    SQLITE_LIMIT_FUNCTION_ARG(6),
+    SQLITE_LIMIT_ATTACHED(7),
+    SQLITE_LIMIT_LIKE_PATTERN_LENGTH(8),
+    SQLITE_LIMIT_VARIABLE_NUMBER(9),
+    SQLITE_LIMIT_TRIGGER_DEPTH(10),
+    SQLITE_LIMIT_WORKER_THREADS(11);
+
+    private final int id;
+
+    SQLiteLimits(int id) {
```
xerial, line 20: If there is no specific reason, we can make the constructor private.

pom.xml
BastouP411: Compiling did not work for me under 1.6. If that’s not needed it can be removed

xerial: Understood. Anyway I have no intention to support below 1.8, we can make 1.8 as the default targe.t

xerial:
Overall it looks good to me. Thanks for the enhancement.

src/main/java/org/sqlite/SQLiteLimits.java
BastouP411: Yep it can be private. No reason it couldn’t.

BastouP411:
Thanks I really needed something like that ^^
So here it is !

src/main/java/org/sqlite/SQLiteLimits.java
xerial: ok. Please do that. I'll merge this PR after that

BastouP411: That's done ^^

PR #528, opened 2020-08-11
Remove Override OBject.finalize in Statement and ResultSet

### LINKED ISSUE #217: Why does Statement and ResultSet classes implement finalize method?
Recently I encountered a GC problem with sqlite-jdbc which was caused by the `finalize` methods in `ResultSet` and `Statement` implementation classes. In my case, the incoming requests varies a lot so that we couldn't fit them into several prepared statement, we had to prepare a statement for each request. 
In java, all objects who implement the `finalize` method will be linked into a `finalize queue` and there `finalize` method will be executed in a serialized way when GC happens. If there's too many objects in the queue, the gc time increase incredibly. And in common java practice, implementing the `finalize` method is NOT a good choice.
So my questions are:
1. What's the purpose of sqlite-jdbc to implement `finalize` method? Is is necessary?
2. If it's necessary, is there a better way for my scenario？

Here comes my testing code:

```java
import java.io.File;
import java.lang.management.ManagementFactory;
import java.sql.Connection;
import java.sql.ResultSet;
import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.ThreadLocalRandom;

/**
 * @author qiuxiafei
 * @date 2017/4/22
 */
public class Query2 {
    private static final int THREAD_NUM = 4;
    private static final int QUERY_PER_THREAD = 4000000;

    public static void main(String[] args) throws InterruptedException {
        System.out.println(ManagementFactory.getRuntimeMXBean().getName());
        long start = System.currentTimeMillis();
        List<Thread> threads = new ArrayList<>();
        for (int i = 0; i < THREAD_NUM; i++) {
            Thread t = new Thread(() -> {
                try {
                    System.err.println("Thread start: " + Thread.currentThread().getId());
                    long queryStart = System.currentTimeMillis();
                    String url = new File("./db.sqlite").getCanonicalFile().toURI().toURL().toString()
                        + "";
                    Connection conn = DriverManager.getConnection("jdbc:sqlite:" + url);


                    for (int j = 0; j < QUERY_PER_THREAD; j++) {
                        PreparedStatement stmt = conn.prepareStatement("select * from test_t where a = ?;");
                        stmt.setInt(1, ThreadLocalRandom.current().nextInt(10000000));
                        ResultSet rs = stmt.executeQuery();
                        while (rs.next()) {
                            rs.getInt(1);
                        }
                        stmt.close();
                        rs.close();
                    }
                    long time = System.currentTimeMillis() - queryStart;
                    double rt = 1.0 * time / QUERY_PER_THREAD;

                    System.out.println(String
                        .format("Thread: %d finish, cost: %d, rt: %.2f", Thread.currentThread().getId(), time, rt));

                } catch (Exception e) {
                    e.printStackTrace();
                }

            });
            threads.add(t);
            t.start();
        }

        for (Thread t : threads) {
            t.join();
        }

        System.err.println("Total cost: " + (System.currentTimeMillis() - start));
    }
}
```
The testing table is created by the following and has 10000000 records:

```sql
create table test_t (a integer primary key, b integer, c integer, d string);
```

szetojeff:
This PR is to fix #217.  Overriding finalize() on these 3 classes actually causes memory leak since the finalize() method adds a reference to the JVM Finalizer and this prevents the objects from garbage collection until the System finalization is invoked by the JVM.  So it is actually counter productive of having these method overridden for attempting to clean up resources; and not to say that, it's an anti-pattern in Java of overriding finalize() for the exact reason it's stated.

Also, these classes already have the close() method implemented, so callers should call close() to release resource and that's how JDBC [Statement](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#close--) and [ResultSet](https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html#close--) interfaces are designed to operate.

I tested this fix in a heavy loaded env with 50 threads and each makes 10+ per second db calls.  Without this fix, the JVM (Xmx=16GB) ran out of memory and hang in a relatively short period of time (15 mins).  By using a Java profiler, I observed over a few GBs of the PreparedStatement and ResultSet objects held by the java.lang.ref.Finalizer and garbage collector could not reclaim these heap memory due to a reference to the Finalizer.  Once I applied this fix, the JVM ran just fine without the memory issue and GC successfully reclaimed the heap resources.

### DIFF
--- a/src/main/java/org/sqlite/core/CorePreparedStatement.java
+++ b/src/main/java/org/sqlite/core/CorePreparedStatement.java
@@ -52,14 +52,6 @@ public abstract class CorePreparedStatement extends JDBC4Statement
     }
 
     /**
-     * @see org.sqlite.core.CoreStatement#finalize()
-     */
-    @Override
-    protected void finalize() throws SQLException {
-        close();
-    }
-
-    /**
      * @see org.sqlite.jdbc3.JDBC3Statement#executeBatch()
      */
     @Override
--- a/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
@@ -157,14 +157,6 @@ public abstract class JDBC3ResultSet extends CoreResultSet {
     }
 
     /**
-     * @see java.lang.Object#finalize()
-     */
-    @Override
-    protected void finalize() throws SQLException {
-        close();
-    }
-
-    /**
      * @see java.sql.ResultSet#getRow()
      */
     public int getRow() throws SQLException {
--- a/src/main/java/org/sqlite/jdbc3/JDBC3Statement.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3Statement.java
@@ -28,13 +28,6 @@ public abstract class JDBC3Statement extends CoreStatement {
     }
 
     /**
-     * @see java.lang.Object#finalize()
-     */
-    protected void finalize() throws SQLException {
-        close();
-    }
-
-    /**
      * @see java.sql.Statement#execute(java.lang.String)
      */
     public boolean execute(String sql) throws SQLException {


xerial:
Nice catch and thanks for the fix! This code was there even before I start maintain sqlite-jdbc here.

tildedave:
Hey @szetojeff just wanted send you a big thanks for doing this.  We run SQLite-JDBC in a somewhat high traffic system and deployed this change to our fleet last week.  It ended up up reducing our max memory consumption by a lot - overall things look much healthier.  You rock!

szetojeff:
> Hey @szetojeff just wanted send you a big thanks for doing this. We run SQLite-JDBC in a somewhat high traffic system and deployed this change to our fleet last week. It ended up up reducing our max memory consumption by a lot - overall things look much healthier. You rock!

You are very welcome.  Great to hear that the fix helps!

guangwei-qiao:
> This PR is to fix #217. Overriding finalize() on these 3 classes actually causes memory leak since the finalize() method adds a reference to the JVM Finalizer and this prevents the objects from garbage collection until the System finalization is invoked by the JVM. So it is actually counter productive of having these method overridden for attempting to clean up resources; and not to say that, it's an anti-pattern in Java of overriding finalize() for the exact reason it's stated.
> 
> Also, these classes already have the close() method implemented, so callers should call close() to release resource and that's how JDBC [Statement](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#close--) and [ResultSet](https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html#close--) interfaces are designed to operate.
> 
> I tested this fix in a heavy loaded env with 50 threads and each makes 10+ per second db calls. Without this fix, the JVM (Xmx=16GB) ran out of memory and hang in a relatively short period of time (15 mins). By using a Java profiler, I observed over a few GBs of the PreparedStatement and ResultSet objects held by the java.lang.ref.Finalizer and garbage collector could not reclaim these heap memory due to a reference to the Finalizer. Once I applied this fix, the JVM ran just fine without the memory issue and GC successfully reclaimed the heap resources.

thanks for your fix, I will try your fix ASAP.

guangwei-qiao:
> > This PR is to fix #217. Overriding finalize() on these 3 classes actually causes memory leak since the finalize() method adds a reference to the JVM Finalizer and this prevents the objects from garbage collection until the System finalization is invoked by the JVM. So it is actually counter productive of having these method overridden for attempting to clean up resources; and not to say that, it's an anti-pattern in Java of overriding finalize() for the exact reason it's stated.
> > Also, these classes already have the close() method implemented, so callers should call close() to release resource and that's how JDBC [Statement](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#close--) and [ResultSet](https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html#close--) interfaces are designed to operate.
> > I tested this fix in a heavy loaded env with 50 threads and each makes 10+ per second db calls. Without this fix, the JVM (Xmx=16GB) ran out of memory and hang in a relatively short period of time (15 mins). By using a Java profiler, I observed over a few GBs of the PreparedStatement and ResultSet objects held by the java.lang.ref.Finalizer and garbage collector could not reclaim these heap memory due to a reference to the Finalizer. Once I applied this fix, the JVM ran just fine without the memory issue and GC successfully reclaimed the heap resources.
> 
> thanks for your fix, I will try your fix ASAP.

BTW, before you patched the fix on you application, did you not close the resultset and preparedstatment appropriately？

szetojeff:
Yes, absolutely we did close resultset and prepared statements, but the
problem is, implementing finalize() causes the JVM finalizer have the
references to the resultset and pstatement objects, so even closing them
won't get them GC.

guangwei-qiao:
> ven closing them won't get them GC.

Already verified, your fix makes sense。Thanks a lot。

guangwei-qiao:
Preparedstaments also cost physical memory, I encountered an issue that off-heap memory can't be gc after I closed preparedstatements, have you ever run into this problem? @szetojeff @xerial

PR #543, opened 2020-09-21
backup/restore use sqlite3_open_v2

wuhongzhi:
support backup/restore with URI filename see https://sqlite.org/uri.html


wuhongzhi:
supports restore/backup attached database, especial for in memory database

xerial:
looks good. This will extend restore/backup functionalities.

PR #553, opened 2020-11-11
adding timeout to OSInfo::isAlpine

martinmeeser:
I am on Ubuntu 20, openJdk 14.0.2,
I am using sqlite from Apache Netbeans with Gradle 6.7

When debugging, isAlpine() kind of always blocks, not sure why. Sometimes it does not block.

When emitting "cat /etc/os-release | grep ^ID" to console in ubuntu, there is no problem, works as excepted.

When running, also sometimes blocks, sometimes runs.

I am wondering if it is a problem with openJdk.

Shortly before getConnection, I delete  the database file with File.deleteIfExists !?

ANYHOW: Having a timeout here is a small improvement. It adds stability.

### DIFF
--- a/src/main/java/org/sqlite/util/OSInfo.java
+++ b/src/main/java/org/sqlite/util/OSInfo.java
@@ -29,6 +29,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


xerial:
Thanks for the fix. Merged

Issue #561, opened 2020-12-04
getString slow for very large ResultSets


tildedave:
Hello, I'm having an issue with the library for large resultsets (many MB).   I can quickly write data into SQLite but (4 minutes) but reading data out takes on the order of 1-2 hours.  By "slow" it looks to be on the order of 7ms per row - unfortunately I have a few million rows.

From running jstack I've pinned it down to `column_text_utf8` and specifically the allocation of a new byte array here: https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java/org/sqlite/core/NativeDB.c#L90-L109

Example jstack (`jstack`):
```
"thread-id" #25621 prio=5 os_prio=0 tid=0x00007fa2a8028800 nid=0x38b7 runnable [0x00007fa1732b6000]
   java.lang.Thread.State: RUNNABLE
        at org.sqlite.core.NativeDB.column_text_utf8(Native Method)
        - locked <0x0000000462e008d8> (a org.sqlite.core.NativeDB)
        at org.sqlite.core.NativeDB.column_text(NativeDB.java:249)
        - locked <0x0000000462e008d8> (a org.sqlite.core.NativeDB)
        at org.sqlite.jdbc3.JDBC3ResultSet.getString(JDBC3ResultSet.java:437)
```

Mixed stack trace (`jstack -m`):
```
----------------- 14519 -----------------
0x00007fa2ce285dfa      _ZN14JNIHandleBlock15allocate_handleEP7oopDesc + 0x4a
0x00007fa2ce2584cb      jni_NewByteArray + 0x21b
```

I believe the issue here ends up being that using `getString` to pull the data out is slow, since it allocates an Java buffer here: https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java/org/sqlite/core/NativeDB.c#L99

I think this ends up being a bit of a double-memory allocation since if SQLite returns a very large dataset the driver needs to allocate a very large Java string to fit the dataset into.

Something I'm not clear on is that this doesn't happen for every query in our application (in fact, unloading some of the troublesome queries is quite fast when run at different times), so I think there must be something special about the how the driver interacts with internal memory that makes this particularly bad.  (Certain amount of native memory already allocated, insufficient pages given to the application, ...) I'm not really a JNI expert so I'm not sure what I might look at to pin this behavior down a bit better.

## Proposed Solution

[...]

PR #564, opened 2020-12-09
Remove byte array allocation from JNI string methods

tildedave:
Getting string data out of SQLite has involved 2 copies: 1 in JNI-land
(allocating a byte buffer to return to Java) and 1 in Java-land (creating a
String instance).

Rather than do the first copy we can return a direct ByteBuffer instance and
then do the conversion from byte -> String in Java as before.  So this removes
a copy getting string data out of SQLite and into Java.

Since these arrays are immediately copied into Java string instances accessing
SQLite memory in Java will be safe.  I added a test for a string that used
non-ASCII codepoints just to make sure this was still working as expected.


tildedave:
Fixes #561 (well, it's much better than existing behavior, at least).

src/test/java/org/sqlite/ResultSetTest.java
```diff
@@ -154,4 +156,18 @@ public void testCloseStatement()
 
         assertTrue(resultSet.isClosed());
     }
+
+    @Test
+    public void testReturnsNonAsciiCodepoints()
```
tildedave, line 161: My changes here didn't break anything, I just noticed there wasn't a test for this.

xerial:
Thanks for the quick fix! The code basically looks good to me. I'll double-check whether we need to properly release the allocated memory in sqlite.

tildedave:
Nice, I was working off the guidance here https://sqlite.org/c3ref/column_blob.html which seems to indicate it would handle it until 
 a type conversion or step() function was called... but there may be more going on.  Thanks for the pointers again!

xerial:
@tildedave Thanks. Yeah. In this usage, allocated DirectByteBuffer will be immediately converted into Strings, so it should be safe.

wuhongzhi:
be care for,   nativeByteBuffer to string is a slow operation due to the decoder access each byte by JNI

PR #565, opened 2020-12-13
fix aarch64 detection

gotson:
This code path probably never worked. `uname -m` returns a value terminated by a newline, so the comparison never worked. Even if it did, it would have returned `arm64`, which is not a valid folder name for the native library.

Why it may have worked on `aarch64` systems is that after that failed comparison, it would detect the JVM and find out it's `armv7`, which is compatible with `arm64`, so probably went unnoticed.


xerial:
@gotson Nice catch!

Issue #546, opened 2020-10-08
Connection.setTransactionIsolation not spec-compliant


A248:
From what I can tell of the Connection.setTransactionIsolation documentation, the driver is supposed to accept the four transaction isolation constants in Connection.  The other JDBC drivers I tested accept all 4 isolation levels, even those that don't implement them.

HSQLDB:
> * The Standard allows the engine to return a higher isolation level than requested by the application. HyperSQL promotes a READ UNCOMMITTED request to READ COMMITTED and promotes a REPEATABLE READ request to SERIALIZABLE. - http://hsqldb.org/doc/guide/guide.html

H2:
> * If repeatable read isolation level is requested when using a PageStore engine it is replaced with serializable isolation level. - http://www.h2database.com/html/advanced.html#transaction_isolation

However, SQLite-JDBC throws if it does not implement the isolation level:

https://github.com/xerial/sqlite-jdbc/blob/6ca62fd07bf7faf58e923ba8651757621ffc693c/src/main/java/org/sqlite/SQLiteConnection.java#L164

SQLite should instead promote "invalid" transaction isolation levels to one of the values implemented.

Also, it looks like SQLite-JDBC takes a similar approach with unsupported fetch directions (#515 ). As stated there, these issues make it slightly harder to write general JDBC code.

A248:
More info on this:

> Note: A JDBC driver might not support all transaction isolation levels. If a driver does not support the isolation level specified in an invocation of setTransactionIsolation, the driver can substitute a higher, more restrictive transaction isolation level. If a driver cannot substitute a higher transaction level, it throws a SQLException. – https://docs.oracle.com/javase/tutorial/jdbc/basics/transactions.html

Since SQLite-JDBC's highest supported transaction isolation is SERIALIZABLE, it logically follows that the case "the driver cannot substitute a higher level" is never satisfied.

A248:
Solved per PR

PR #573, opened 2021-01-12
Improve Compliance With JDBC Specifications

### LINKED ISSUE #515: Statement.setFetchDirection non-compliant?
```
     * Gives the driver a hint as to the direction in which
     * rows will be processed in <code>ResultSet</code>
     * objects created using this <code>Statement</code> object.  The
     * default value is <code>ResultSet.FETCH_FORWARD</code>.
```

Giving this hint causes an exception.

```
java.sql.SQLException: ResultSet closed
	at org.sqlite.core.CoreResultSet.checkOpen(CoreResultSet.java:76)
	at org.sqlite.jdbc3.JDBC3ResultSet.setFetchDirection(JDBC3ResultSet.java:125)
	at org.sqlite.jdbc3.JDBC3Statement.setFetchDirection(JDBC3Statement.java:326)
```

A248:
Should fix #515 and #546 

* Connection.setTransactionIsolation - promote isolation levels READ_COMMITTED and REPEATABLE_READ, which are not natively supported by SQLite, to SERIALIZABLE. The driver should accept all four values as long as it can substitute a higher isolation level than the one requested.
* Statement.setFetchDirection - while SQLite does not act on the fetch direction hint, it should accept the call so long as the argument is valid.

In both of these cases, spec-compliance makes it easier to write driver-agnostic user code.

ResultSet.setFetchDirection was not changed, because the current impl is actually correct. It is correct because SQLite only allows FORWARD_ONLY ResultSets.

### DIFF
--- a/src/main/java/org/sqlite/SQLiteConnection.java
+++ b/src/main/java/org/sqlite/SQLiteConnection.java
@@ -154,6 +154,9 @@ public abstract class SQLiteConnection
         checkOpen();
 
         switch (level) {
+            case java.sql.Connection.TRANSACTION_READ_COMMITTED:
+            case java.sql.Connection.TRANSACTION_REPEATABLE_READ:
+                // Fall-through: Spec allows upgrading isolation to a higher level
             case java.sql.Connection.TRANSACTION_SERIALIZABLE:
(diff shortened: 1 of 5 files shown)


xerial:
Thanks! Merged

Issue #569, opened 2020-12-25
the merge from https://github.com/xerial/sqlite-jdbc/pull/564 led to performance lost


wuhongzhi:
As in in thread 564 report, nativeByteBuffer to string is a slow operation as the decoder get byte one by one via JNI. In my test in a large table(1m rows), lost 15% compare to previous version.

src for open jdk
bytebuffer to string in sun.nio.cs.UTF_8.Decoder.decodeBufferLoop(ByteBuffer, CharBuffer)

----------------
According to http://www.java-performance.com/
 Performance of various methods of binary serialization in Java: java.nio.ByteBuffer, sun.misc.Unsafe, java.io.DataInputStream, java.io.DataOutputStream, java.io.ByteArrayInputStream, java.io.ByteArrayOutputStream: comparison of binary serialization performance using various classes: 

RULE1:(same for reading)
 It is extremely slow to write single bytes to direct byte buffers. You should avoid using direct byte buffers for writing records with mostly single byte fields.

wuhongzhi:
note: direct bytebuffer has limit by jvm, -XX:MaxDirectMemorySize default to 64Mb in same jvm, but byte[] limit to 2Gb

xerial:
👀

tildedave:
Hey @wuhongzhi can you share your benchmarking code so I can run through it and confirm the performance loss?  Also could you please include your JVM version?

PR #564 didn't add any new single-byte writes to a direct byte buffer - so it doesn't violate RULE 1 here.  Instead, the patch allocates a new buffer around the existing data returned from SQLite via https://docs.oracle.com/javase/8/docs/technotes/guides/jni/jni-14.html#NewDirectByteBuffer.  Then that data is read out of the ByteBuffer and converted to a String object in Java (NativeDB.java).

We've been running with the patch from #564 in production and our worst case read times have improved.  Happy to take a look at this again if you're running into problems though.

wuhongzhi:
Hello @tildedave, I have a test on bytes vs buffer to simulate your patch
-------
import java.nio.ByteBuffer;
import java.nio.charset.StandardCharsets;

public class BytesVsBuffer {
	private static final byte[] data;
	static {
[...]

PR #574, opened 2021-01-12
revert buffer to bytes between jni , see discuss in #569

wuhongzhi:
As buffer used in JNI led to performance decline, this pull just revert buffer to bytes


tildedave:
Hey @wuhongzhi I have a different fix for this that I'll be opening this week.  Just need to do a little more investigation on my side.  (I work on this in combination with my usual work responsibilities.)

wuhongzhi:
OK, I'll close this PR

PR #575, opened 2021-01-12
Don't use ByteBuffer decode() - slower

tildedave:
Profiling this has revealed it's faster to just use the String decoding methods vs the ByteBuffer decoding methods.  This adds a copy back to the process of getting data out of SQLite but it removes the JNI allocation (my original issue).

### DIFF
--- a/src/main/java/org/sqlite/core/NativeDB.java
+++ b/src/main/java/org/sqlite/core/NativeDB.java
@@ -504,7 +504,9 @@ public final class NativeDB extends DB
             return null;
         }
         try {
-            return Charset.forName("UTF-8").decode(buffer).toString();
+            byte[] buff = new byte[buffer.remaining()];
+            buffer.get(buff);
+            return new String(buff, Charset.forName("UTF-8"));
         }
         catch (UnsupportedCharsetException e) {
             throw new RuntimeException("UTF-8 is not supported", e);
--- a/src/main/java/org/sqlite/core/NativeDB.java
+++ b/src/main/java/org/sqlite/core/NativeDB.java
@@ -491,26 +491,16 @@ public final class NativeDB extends DB
         if (str == null) {
             return null;
         }
-        try {
-            return str.getBytes("UTF-8");
-        }
-        catch (UnsupportedEncodingException e) {
-            throw new RuntimeException("UTF-8 is not supported", e);
-        }
+        return str.getBytes(StandardCharsets.UTF_8);
     }
 
     static String utf8ByteBufferToString(ByteBuffer buffer) {
         if (buffer == null) {
             return null;
         }
-        try {
-            byte[] buff = new byte[buffer.remaining()];
-            buffer.get(buff);
-            return new String(buff, Charset.forName("UTF-8"));
-        }
-        catch (UnsupportedCharsetException e) {
-            throw new RuntimeException("UTF-8 is not supported", e);
-        }
+        byte[] buff = new byte[buffer.remaining()];
+        buffer.get(buff);
+        return new String(buff, StandardCharsets.UTF_8);
     }
 
     public native synchronized void register_progress_handler(int vmCalls, ProgressHandler progressHandler) throws SQLException;


tildedave:
@wuhongzhi Please try this out and see if it meets your needs

wuhongzhi:
hello @tildedave, i refined my test case, can you please test in your side
```
import java.nio.ByteBuffer;
import java.nio.charset.StandardCharsets;
import java.sql.Connection;
import java.sql.ResultSet;
import java.sql.Statement;

public class BytesVsBuffer {
	private static final byte[] data;
	private static final ByteBuffer buff;
	static {
		data = "바이트 버퍼는 직접 또는 비 직접입니다. 직접 바이트 버퍼를 감안할 때 Java 가상 시스템은 네이티브 I/O 작업을 직접 수행하기 위해 최선을 다할 것입니다. 즉, 버퍼의 콘텐츠를 기본 운영 체제의 기본 I/O 작업 중 하나를 호출하기 전(또는 이후) 중간 버퍼에 복사하지 않도록 시도합니다. "
					.getBytes(StandardCharsets.UTF_8);
		buff = ByteBuffer.allocateDirect(data.length);
		buff.put(data);
	}
	public static void main(String[] args) throws Throwable {
		int round = 1_000_000;
		long stime = System.nanoTime();
		for (int i = 0; i < round; i++) {
			bytesBenchmark();
		}
		long etime = System.nanoTime();
		System.out.println(String.format("new String from bytes taken %,dns", etime - stime));
		stime = System.nanoTime();
		for (int i = 0; i < round; i++) {
			bufferBenchmark();
		}
		etime = System.nanoTime();
		System.out.println(String.format("new String from buffer taken %,dns", etime - stime));
		
		sqliteBenchmark(round * 5);
	}
	public static String bytesBenchmark() {
		return new String(data, StandardCharsets.UTF_8);
	}
	public static String bufferBenchmark() {
		return StandardCharsets.UTF_8.decode(buff.flip()).toString();
	}
	public static void sqliteBenchmark(int round) throws Throwable {
		try (Connection conn = DriverManager.getConnection("jdbc:sqlite::memory:");
				Statement stat = conn.createStatement()) {
			stat.execute("CREATE TABLE A(CA TEXT)");
			stat.execute("BEGIN");
			long stime = System.nanoTime();
			try (PreparedStatement ps = conn.prepareStatement("INSERT INTO A VALUES (?)")) {
				String str = bytesBenchmark();
				for (int i = 1; i <= round; i++) {
					ps.setString(1, str);
					ps.addBatch();
					if (i % 100_000 == 0) {
						ps.executeBatch();
					}
				}
			}
			long etime = System.nanoTime();
			System.out.println(String.format("Insert into SQLite taken %,dns", etime - stime));
			stime = System.nanoTime();
			try (ResultSet rs = stat.executeQuery("SELECT * FROM A")) {
				while(rs.next()) {
					rs.getString(1);
				}
			}
			etime = System.nanoTime();
			System.out.println(String.format("Select from SQLite taken %,dns", etime - stime));
			stat.execute("END");
		}
	}
}
```
Here is my test result
with byte array
---
new String from bytes taken 944,881,397ns
new String from buffer taken 1,239,318,420ns
Insert into SQLite taken 8,899,088,083ns
Select from SQLite taken 5,946,611,094ns

new String from bytes taken 809,367,322ns
new String from buffer taken 1,139,459,562ns
Insert into SQLite taken 9,253,462,318ns
Select from SQLite taken 5,575,465,904ns

new String from bytes taken 816,239,137ns
new String from buffer taken 1,159,504,275ns
Insert into SQLite taken 9,034,006,951ns
Select from SQLite taken 5,492,643,857ns

with byte buffer
---
new String from bytes taken 817,852,197ns
new String from buffer taken 1,175,292,010ns
Insert into SQLite taken 8,886,987,639ns
Select from SQLite taken 6,604,906,645ns

new String from bytes taken 800,021,708ns
new String from buffer taken 1,197,219,744ns
Insert into SQLite taken 8,838,888,945ns
Select from SQLite taken 6,422,573,051ns

new String from bytes taken 794,072,063ns
new String from buffer taken 1,165,106,852ns
Insert into SQLite taken 8,967,192,173ns
Select from SQLite taken 6,587,834,230ns

wuhongzhi:
My new patch https://github.com/wuhongzhi/sqlite-jdbc/commit/cccc37ee5a1a9fe32c813394e81d6e53bbdfec84
little faster(5%) than byte[] by change UTF-8 to Modified UTF-8(CESU-8) in JNI

src/main/java/org/sqlite/core/NativeDB.java
```diff
@@ -504,7 +504,9 @@ static String utf8ByteBufferToString(ByteBuffer buffer) {
             return null;
         }
         try {
-            return Charset.forName("UTF-8").decode(buffer).toString();
+            byte[] buff = new byte[buffer.remaining()];
+            buffer.get(buff);
+            return new String(buff, Charset.forName("UTF-8"));
```
xerial, line 509: By the way, we can use StandardCharsets.UTF_8 here https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html#UTF_8

tildedave: Oh, OK.  I think I was following some existing code which was supporting language version < 7 - I'll update.

xerial: Yeah. I think we don't need to support Java6 anymore.

tildedave: OK, pushed a fix for this.

tildedave:
Hey @wuhongzhi I'm going to have to spend some more time looking into your C code.  Thanks for the examples; I wonder if there's a way to do the same through one of the standard Java libraries.

I'm disappointed that the decode method on the ByteBuffer class is slower, I still want to understand why this is a bit more.  However, I only have so much time to look at these things and this week has been a different priority at work :)

xerial:
Merged to master. Thanks all for checking the performance

PR #590, opened 2021-03-23
SQLite 3.35.0, build failure

fatalmind:
SQLite 3.35.0 introduced math functions, of which at least signFunc collides with the one provided by extension-functions.c.

This is a minimal patch just removing this function for 3.35.0 onwards so that it at least compiles. I haven't checked yet if other functions became obsolete, but I'm following this up "upstream": https://gitlab.com/liamh/extension-functions/-/issues/4


pom.xml
```diff
@@ -4,7 +4,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.xerial</groupId>
   <artifactId>sqlite-jdbc</artifactId>
-  <version>3.34.0.1-SNAPSHOT</version>
+  <version>3.35.0</version>
```
xerial, line 7:
```suggestion
  <version>3.35.0-SNAPSHOT</version>
```
We don't need to change this version before the release.

xerial:
Thanks for the fix. I'll prepare 3.35.0 release

Issue #88, opened 2016-01-17
JDBC driver unable to parse ISO 8601 dates


lnreeder:
The https://www.sqlite.org/datatype3.html indicates date/time functions in SQLite  can store dates as:  "TEXT as ISO8601 strings ("YYYY-MM-DD HH:MM:SS.SSS"). "

However, when storing an ISO 8601 string like 2016-01-01 00:00:00, then attempting to use sqlite-jdbc to retrieve as a java Timestamp, ResultSet.getTimestamp returns an error:

testSelectIso8601Timestamp(org.sqlite.ResultSetTest)  Time elapsed: 0.003 sec  <<< ERROR!
java.sql.SQLException: Error parsing time stamp
    at org.sqlite.jdbc3.JDBC3ResultSet.getTimestamp(JDBC3ResultSet.java:532)
    at org.sqlite.ResultSetTest.testSelectIso8601Timestamp(ResultSetTest.java:156)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:483)
    at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:44)
    at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:15)
    at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:41)
    at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:20)
    at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:28)
    at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:31)
    at org.junit.runners.BlockJUnit4ClassRunner.runNotIgnored(BlockJUnit4ClassRunner.java:79)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:71)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:49)
    at org.junit.runners.ParentRunner$3.run(ParentRunner.java:193)
[...]

PR #609, opened 2021-06-08
Normalize source string when parsing

davidjpfeiffer:
This pull request fixes the bug described by issue https://github.com/xerial/sqlite-jdbc/issues/88

This bug can be reproduced by using `ResultSet.getDate()` to parse a `DATETIME` created using `CURRENT_TIMESTAMP`

The exception thrown will look something like this:
`java.text.ParseException: Unparseable date: "2021-06-08 02:00:47" does not match (\p{Nd}++)\Q-\E(\p{Nd}++)\Q-\E(\p{Nd}++)\Q \E(\p{Nd}++)\Q:\E(\p{Nd}++)\Q:\E(\p{Nd}++)\Q.\E(\p{Nd}++)`

The regex used to parse dates does not account for the fact that new dates created by SQLite's `CURRENT_TIMESTAMP` do not include millisecond information. The regex used by the date parser tries to find a string matching `yyyy-MM-dd HH:mm:ss.SSS` but the dates created by SQLite do not include the milliseconds `.SSS` so the date strings are not matching and this is causing the parser to return null. The solution implemented in this pull request detects this exact case (string length is exactly 19) and then appends `.000` to the source string. This causes the regex to match on the string allowing the date string to be parsed and the Date object to be returned. The source object is not mutated, and the extra characters do not change the `DATETIME` represented by the date string, so I think this is a safe fix.

Please let me know if you have any questions about this.

### DIFF
--- a/src/main/java/org/sqlite/date/FastDateParser.java
+++ b/src/main/java/org/sqlite/date/FastDateParser.java
@@ -288,15 +288,16 @@ public class FastDateParser implements DateParser, Serializable {
      * @see org.apache.commons.lang3.time.DateParser#parse(java.lang.String)
      */
     public Date parse(final String source) throws ParseException {
-        final Date date= parse(source, new ParsePosition(0));
+        String normalizedSource = source.length() == 19 ? (source + ".000") : source;
+        final Date date= parse(normalizedSource, new ParsePosition(0));
         if(date==null) {
             // Add a note re supported date range
             if (locale.equals(JAPANESE_IMPERIAL)) {
                 throw new ParseException(
                         "(The " +locale + " locale does not support dates before 1868 AD)\n" +
-                                "Unparseable date: \""+source+"\" does not match "+parsePattern.pattern(), 0);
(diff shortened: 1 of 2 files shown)


bendem:
No unit test to catch possible future regression?

davidjpfeiffer:
@bendem I'm a little outside my comfort zone here, but I gave it a shot. I added a unit test to the StatementTest class. Let me know what you think.

bendem:
Looks good to me, but I'm not maintainer :)

xerial:
Thanks for the fix. It looks good to me

xerial:
@davidjpfeiffer Sorry. It's once merged, but the CI is failing because of this change. Could you submit a change again?

davidjpfeiffer:
@xerial sure, should I create a new PR with the exact same changes?

xerial:
Yes. In another PR, GitHub Action CI will run to find any issue in advance

davidjpfeiffer:
@xerial https://github.com/xerial/sqlite-jdbc/pull/620

PR #627, opened 2021-07-21
Support custom collation creation

### LINKED ISSUE #594: How to create collation?
SQLite [exposes](https://www.sqlite.org/c3ref/create_collation.html) a `create_collation` function to add custom collation.

I couldn't find the equivalent in this library.

I noticed that the Python [library](https://docs.python.org/3/library/sqlite3.html) for SQLite exposes that function, and let the user write Python code for the collation behaviour. That would be very interesting to have in Java, for example to implement `UNICODE` collation that is not shipped by default.

gotson:
In the same manner as custom functions, this PR brings support for custom collations. An example in the tests shows how a reverse collation can be implemented.

My knowledge of JNI and C is close to zero, I managed to get everything to work by copy/pasting a lot of code done for custom functions / UDF. I would expect that there are many ways to optimize the code, or fix some issues i have no idea were there in the first place 🙃 

The good thing is that the code is working, and tests pass on my local machine (after recompiling the native library).

Comments and updates are welcome!

Closes #594


gotson:
I think i will need a bit more help here, i was trying to use this in my own project, and i receive some garbled strings in the `xCompare` function sometimes. Usually `str2` is correct, but `str1` is garbled.

That's probably something to do with the text encoding, and how the data is converted to a Java String, unfortunately my JNI / C skills are lame and i have no idea how to fix that.

Hopefully it will be obvious to someone who knows what they are doing 🙃

gotson:
I tried to fix the garbled strings, to no avail. I think the problem is when building the Java strings, too much data is retrieved, or in the wrong encoding, which generates some garbled data after the actual string characters.

I am following the SQLite [documentation](https://www.sqlite.org/c3ref/create_collation.html) for implementation. It says:

> The collating function callback is invoked with a copy of the pArg application data pointer and with two strings in the encoding specified by the eTextRep argument. The two integer parameters to the collating function callback are the length of the two strings, in bytes.

`eTextRep` has been set to `SQLITE_UTF16`:

```c
ret = sqlite3_create_collation(
            gethandle(env, this),
            name_bytes,            // collation name
            SQLITE_UTF16,          // preferred chars
            coll,
            &xCompare
    );
```

This is the `xCompare` function, i think there's something wrong with the way the `jstring` are built:

```c
int xCompare(void* context, int len1, const void* str1, int len2, const void* str2)
{
    static jmethodID mth = 0;
    JNIEnv *env;
    struct CollationData *coll = (struct CollationData*)context;
    (*coll->vm)->AttachCurrentThread(coll->vm, (void **)&env, 0);

    if (!mth) {
        mth = (*env)->GetMethodID(env, cclass, "xCompare", "(Ljava/lang/String;Ljava/lang/String;)I");
    }

    jstring jstr1=(*env)->NewString(env, str1, len1);
    jstring jstr2=(*env)->NewString(env, str2, len2);

    return (*env)->CallIntMethod(env, coll->func, mth, jstr1, jstr2);
}
```

I am using `NewString` which takes Unicode characters, I suppose that's UTF_16.

> String Operations
> NewString
> jstring NewString(JNIEnv *env, const jchar *unicodeChars,
> jsize len);
> 
> Constructs a new java.lang.String object from an array of Unicode characters.

I am not sure about the last argument, which is supposed to be a `jsize len`. The one coming from SQLite is _the length of the two strings, in bytes_.

gotson:
Good news ! I managed to fix the problem of strings being garbled in Java, thanks to someone else's expertise.

Would appreciate some review on the code, especially on potential memory leaks in C code, which is not my area of expertise.

gotson:
i've added support to destroy custom collation

xerial:
I'd like to include this PR into the next release. Trying to find a time for the review. Wait for a while. Thanks

src/main/java/org/sqlite/core/NativeDB.c
```diff
@@ -429,6 +434,26 @@ void xFinal(sqlite3_context *context)
     (*env)->DeleteGlobalRef(env, *func);
 }
 
+int xCompare(void* context, int len1, const void* str1, int len2, const void* str2)
+{
+    static jmethodID mth = 0;
```
xerial, line 439: What if we register two or more Collation functions to the connection? I guess only the first one wins and collation functions created later will never be used.

src/test/java/org/sqlite/CollationTest.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    @After
+    public void close() throws SQLException {
+        stat.close();
+        conn.close();
+    }
+
+    @Test
+    public void reverseCollation() throws SQLException {
```
xerial, line 43: Good exmaple ;)

gotson: Inspired by the Python Sqlite [doc](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.create_collation) ;)

src/main/java/org/sqlite/core/NativeDB.c
gotson: I have added a test to validate that it works, please check the latest commit.

The custom collation works a bit differently than the UDF. [Here](https://www.sqlite.org/c3ref/create_collation.html) you can read:

> The fourth argument, pArg, is an application data pointer that is passed through as the first argument to the collating function callback.

[This](https://github.com/gotson/sqlite-jdbc/blob/884288d946bfc1204b4db1567df91ca81436464d/src/main/java/org/sqlite/core/NativeDB.c#L1360) is what we pass, it's a struct containing the pointer to the custom function

> The collating function callback is invoked with a copy of the pArg application data pointer

In the callback, [we can access](https://github.com/gotson/sqlite-jdbc/blob/884288d946bfc1204b4db1567df91ca81436464d/src/main/java/org/sqlite/core/NativeDB.c#L441) the previously defined `coll` struct. [Here](https://github.com/gotson/sqlite-jdbc/blob/884288d946bfc1204b4db1567df91ca81436464d/src/main/java/org/sqlite/core/NativeDB.c#L454) we call the method on that particular object, so no conflict can arise.

---

I have some doubts on whether the memory is actually freed for those `coll` structs. When I wrote this code, i tried to find how the UDF data pointers are freed, but couldn't find anything. Today i found that it's done [here](https://github.com/gotson/sqlite-jdbc/blob/884288d946bfc1204b4db1567df91ca81436464d/src/main/java/org/sqlite/core/NativeDB.c#L1386). So there's probably something to do to properly free the memory for the custom collation functions too.

gotson:
regarding the memory cleanup, would you be ok if I added the cleanup in the [existing cleanup function](https://github.com/gotson/sqlite-jdbc/blob/884288d946bfc1204b4db1567df91ca81436464d/src/main/java/org/sqlite/core/NativeDB.c#L1386) used for UDF ? Or do you want me to create a new one just for collation?

I will use the same mechanism of linked list.

gotson:
I have added cleanup code, similar to the UDF one, inside the existing function, which is already called on connection close. Let me know if you want to split that function into a dedicated collation cleanup function.

gotson:
Hi @ericdallo, since this PR adds some native methods, do we need to amend the jni json file added in #631?

ericdallo:
@gotson probably yes, but the ideal would be compile sqlite-jdbc with graalvm and test your new feature if it works

gotson:
> @gotson probably yes, but the ideal would be compile sqlite-jdbc with graalvm and test your new feature if it works

I tried but i can't even get the project to run the sample in native mode 🤷🏻‍♂️

ericdallo:
@gotson, you can try using [my sample sqlite graalvm project](https://github.com/ericdallo/sqlite-graalvm-sample), it already has a script to compile with graalvm so you generate a local sqlite-jdbc snapshot, change the `project.clj` version and just run the [compile script](https://github.com/ericdallo/sqlite-graalvm-sample/blob/master/graalvm/native-unix-compile.sh)

PR #628, opened 2021-07-23
bring back armv6 support

### LINKED ISSUE #583: ARMv6: no native library found
There is no `armv6` folder in https://github.com/xerial/sqlite-jdbc/tree/master/src/main/resources/org/sqlite/native/Linux

However, the code [can return](https://github.com/xerial/sqlite-jdbc/blob/a497c72922b043b6295a753fe44239eb91da3541/src/main/java/org/sqlite/util/OSInfo.java#L177) `armv6` as an arch type.

It seems armv6 has been disabled, according to [this makefile comment](https://github.com/xerial/sqlite-jdbc/blob/21055a2b75b70039bd9f688deea2408d7d1c2114/Makefile#L106).

Given armv6 is backward compatible with armv5, could we copy the content of https://github.com/xerial/sqlite-jdbc/tree/master/src/main/resources/org/sqlite/native/Linux/arm to an armv6 folder in the meantime?

gotson:
Following the fix done in https://github.com/dockcross/dockcross/pull/510 we can bring back armv6 support

It would be good if someone with a Raspberry PI with armv6 could test this.

Closes #583

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,8 +111,7 @@ NATIVE_TARGET_DIR:=$(TARGET)/classes/org/sqlite/native/$(OS_NAME)/$(OS_ARCH)
 NATIVE_DLL:=$(NATIVE_DIR)/$(LIBNAME)
 
 # For cross-compilation, install docker. See also https://github.com/dockcross/dockcross
-# Disabled linux-armv6 build because of this issue; https://github.com/dockcross/dockcross/issues/190
-native-all: native win32 win64 mac64 linux32 linux64 linux-arm linux-armv7 linux-arm64 linux-android-arm linux-ppc64 alpine-linux64
+native-all: native win32 win64 mac64 linux32 linux64 linux-arm linux-armv6 linux-armv7 linux-arm64 linux-android-arm linux-ppc64 alpine-linux64
 
 native: $(NATIVE_DLL)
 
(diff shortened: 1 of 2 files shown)


gotson:
It seems this doesn't work as expected. This has been tested on a Raspberry Pi Zero:

```
Aug 04 20:30:15 Zero Komga[11089]: Failed to load native library:sqlite-3.32.3.7-a9581672-5e6b-4b2c-8ade-7a4c4ad586e8-libsqlitejdbc.so. osinfo: Linux/armv6
Aug 04 20:30:15 Zero Komga[11089]: java.lang.UnsatisfiedLinkError: /tmp/sqlite-3.32.3.7-a9581672-5e6b-4b2c-8ade-7a4c4ad586e8-libsqlitejdbc.so: /lib/arm-linux-gnueabihf/libc.so.6: version `GLIBC_2.33' not found (required by /tmp/sqlite-3.
32.3.7-a9581672-5e6b-4b2c-8ade-7a4c4ad586e8-libsqlitejdbc.so)
```

From the tester:
> This is sadly not applicable for use since not even the upcoming Debian Bullseye ships it: https://packages.debian.org/bullseye/libc6
Not sure if there is a way to build it against v2.28 at least, which would mean compatibility with Debian Bullseye.

I was surprised to see that `libsqlitejdbc.so` is dynamically built, but it seems statically building `glibc` is [discouraged](https://stackoverflow.com/questions/57476533/why-is-statically-linking-glibc-discouraged), or will [still result in dynamic dependencies](https://stackoverflow.com/questions/26304531/compiling-with-static-libgcc-static-libstdc-still-results-in-dynamic-depende).

Is there a way to choose the `glibc` version we use via `dockcross` ? It seems it can be done via `crosstools-ng`, but the way we use the `dockcross` images doesn't seem to make this easy.

xerial:
What we basically can do is building native libraries using very old glibc. As dockcross prefers recent OSs containing newer glibc, there is no way to fix this backward incompatibility unless we build a statically linked binary. 

For example, for building x86_64, sqlite-jdbc is using a really old version of Linux to use ancient GLIBC version https://github.com/xerial/sqlite-jdbc/blob/master/docker/Dockerfile.linux_x86_64

xerial:
I'm ok with merging this even though it's not working for some armv6 machines (e.g., old Rapsberry Pi, etc.). After that, we can think about potential fixes.

gotson:
> I'm ok with merging this even though it's not working for some armv6 machines (e.g., old Rapsberry Pi, etc.). After that, we can think about potential fixes.

we are testing the new `dockcross/linux-armv6-lts` image today, so this can probably wait a day or 2.

gotson:
Tests are successful, I'll update this PR to use the lts dockcross image

gotson:
@xerial this should be ready to go, i've updated the dockcross script to use `dockcross/linux-armv6-lts` which uses an older glibc (2.28).

xerial:
@gotson Thanks! Merged

PR #629, opened 2021-07-23
fix Linux PPC64 build failing because of recent dockcross changes

gotson:
`make linux-ppc64` would fail following changes in https://github.com/dockcross/dockcross/commit/4ac8689b5f91f95e5b54d704f8a810df24dc37e5


xerial:
LGTM. Thanks for the fix

PR #630, opened 2021-07-23
update all dockcross scripts via dockcross-xxx update

gotson:
Dockcross scripts installed locally have been updated using `./docker/dockcross-<xxx> update`


xerial:
Thanks for the update. Merged

PR #631, opened 2021-07-24
[graalvm] Add necessary config for graalvm native-image compilation

ericdallo:
Hello,
Following [this recommended way](https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/#embedding-a-configuration-file) from Oracle's Graalvm, this should make it easier for others that want to use sqlite-jdbc in a native-image compiled with GraalVM, this way users don't need to add those configs manually as GraalVM will find it in the classpath if using sqlite-jdbc lib.

[This](https://github.com/ericdallo/sqlite-graalvm-sample) is a sample showing that config works
I was using [this lib](https://github.com/ericdallo/sqlite-jni-graal-fix) in other projects to fix those missing configs, but after compiling it in json correctly now it's possible to bring those configs to sqlite-jdbc to make it available to everyone, if this PR get merged, that lib could be deprecated/archived.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,6 +51,12 @@
       </resource>
       <resource>
         <directory>src/main/resources</directory>
+        <includes>
+          <include>META-INF/native-image/**</include>
+        </includes>
+      </resource>
+      <resource>
+        <directory>src/main/resources</directory>
         <targetPath>META-INF/services</targetPath>
         <includes>
           <include>java.sql.Driver</include>
--- a/src/main/resources/META-INF/native-image/org.xerial/sqlite-jdbc/jni-config.json
+++ b/src/main/resources/META-INF/native-image/org.xerial/sqlite-jdbc/jni-config.json
@@ -1,0 +1,52 @@
+[
+    {
+        "name":"org.sqlite.core.DB",
+        "allDeclaredMethods":true,
+        "allPublicMethods": true,
+        "allDeclaredFields":true,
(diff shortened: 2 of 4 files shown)


ericdallo:
I tested this SNAPSHOT with [clojure-lsp](https://clojure-lsp.github.io/clojure-lsp/) and it worked as well

ericdallo:
Ping @xerial

xerial:
Thanks! It's great to know sqlite-jdbc works with GraalVM in such a manner.

kkriske:
This adds the configuration unconditionally, at the very least there should be a condition added that `org.sqlite.core.NativeDB` is reachable (or a more appropriate entrypoint, this one is just what we use): https://github.com/oracle/graal/blob/master/docs/reference-manual/native-image/Reflection.md#conditional-configuration

Secondly, this also includes all resources on all operating systems, while only the relevant native library should be packaged. It may even be possible to statically link in the library and not include any of them as resources but that's a whole other story.

We've been running sqlite-jdbc in native-image for a while now with our own configuration which is a fair bit more detailed than this, allbeit as a [Feature](https://github.com/oracle/graal/blob/vm-21.3.0/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/Feature.java) implementation. That way be can add detailed reachabillity handlers that only add configuration for what is required.
Config options such as `allDeclaredMethods` should be avoided in my opinion. This registers too much and only causes image bloat as it included both unused methods and reflection metadata for methods/fields that are never accessed anyway.

I'm also curious why the reflection config is needed, as we don't have any.

I can find out if I am allowed to share our configuration setup if you would be interested.

PR #639, opened 2021-08-09
Add FreeBSD x86 and x86_64 native library cross-compilation via Docker

gotson:
This uses [empterdose/freebsd-cross-build:9.3](https://github.com/MrDOS/freebsd-cross-build) to compile for FreeBSD x86 and x86_64. This has been tested via Komga by some of our users and reported to work well.

Note that dockcross doesn't support any FreeBSD build.

I thought about creating a derived Docker image that would have `bash`, `java` and `perl` installed, but given the frequency of rebuilds this is probably overkill.

Supersedes #494


xerial:
Great! Merged

PR #644, opened 2021-08-20
Add windows armv7 and arm64 native libraries via dockcross

### LINKED ISSUE #577: Compile native libs for Windows ARM 32 and 64
Hi - Can you please include binaries for the 32bit and 64bit ARM architectures for Windows so that devices such as the Microsoft Surface Pro X are compatible with the library.

```
Caused by: java.lang.Exception: No native library is found for os.name=Windows and os.arch=aarch64. path=/org/sqlite/native/Windows/aarch64
```

Thanks!

gotson:
The result of the `file` command is not strictly equivalent to the x86 binaries, but that may be due to the compiler being `clang`.

```
src/main/resources/org/sqlite/native/Windows> file */*
aarch64/sqlitejdbc.dll: PE32+ executable (DLL) (GUI), for MS Windows
armv7/sqlitejdbc.dll:   PE32 executable (DLL) (GUI) ARMv7 Thumb, for MS Windows
x86/sqlitejdbc.dll:     PE32 executable (DLL) (console) Intel 80386 (stripped to external PDB), for MS Windows
x86_64/sqlitejdbc.dll:  PE32+ executable (DLL) (console) x86-64 (stripped to external PDB), for MS Windows
```

Closes #577


xerial:
Thanks! RIght. For some reason, Windows .dll will always have some differences. I guess some date-related value is embedded.

PR #650, opened 2021-08-27
Github Action tests for GraalVM native-image

gotson:
This adds CI tests for GraalVM native-image. It uses the official plugin, and builds upon #649 as Junit5 is required to run tests automatically inside the native-image.

Some tests are failing, but since Maven Surefire plugin can only exclude tests by class, I had to split some classes, so most of the tests would still run. I tried adding as much information as I could in `pom.xml` about the test exclusions. Note that the exclusions are only done when running with the `native` profile.

I also had to disable native tests on Java 8 under ubuntu, it fails with some `unknown symbol: cos`, while it works fine on Java 11, and on other platforms.

@ericdallo your feedback would be appreciated, as I had to change a few things around configuration for JNI reflection.


xerial:
#649 is merged. We need to resolve conflicts.

.github/workflows/test.yml
```diff
@@ -40,3 +40,40 @@ jobs:
           restore-keys: ${{ runner.os }}-jdk${{ matrix.java }}-
       - name: Test
         run: mvn test
+
+  test_graalvm:
+    name: test ${{ matrix.os }} ${{ matrix.java }} GraalVM native-image
+    strategy:
+      fail-fast: false
+      matrix:
```
xerial, line 48: As the GitHub Action capacity is not infinite, I'd like to limit the scope of GraalVM only to the lastest JDK (java11 or java16) and Linux.

ericdallo: Agreed, I use graalvm only with jdk11 for clojure-lsp test CI as well

gotson: @xerial what do you mean it's not infinite? There's no credit or anything for OSS projects AFAIK.

gotson:
> #649 is merged. We need to resolve conflicts.

I will take care of this.

.github/workflows/test.yml
ericdallo: @gotson there is no credit, but there are hardware limitation when you spawn a lot of heavy processing actions GitHub only run new jobs when others finish

gotson: I think they run in order, so the classic tests run first, and the graalvm ones run after. I'm not sure we need feedback that fast to be honest?

There are differences between the different platforms though. I agree on java 8, we can drop it and keep only 11 for now.

Ultimately it's up to @xerial, I am happy to change it anyway you prefer!

.github/workflows/test.yml
```diff
@@ -40,3 +40,40 @@ jobs:
           restore-keys: ${{ runner.os }}-jdk${{ matrix.java }}-
       - name: Test
         run: mvn test
+
+  test_graalvm:
+    name: test ${{ matrix.os }} ${{ matrix.java }} GraalVM native-image
+    strategy:
+      fail-fast: false
```
gotson, line 47: I need to remove this, I disabled fail fast during tests but it should be activated.

xerial:
Merged. Thanks for the test code cleanup too.

PR #653, opened 2021-08-31
Code formatting via Spotless with google java format

gotson:
as discussed here: https://github.com/xerial/sqlite-jdbc/discussions/651

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,6 +21,17 @@ on:
       - '.github/workflows/*.yml'
 
 jobs:
+  lint:
+    name: Check code formatting
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v1
+        with:
+          java-version: '11'
+      - name: spotless:check
+        run: mvn spotless:check
+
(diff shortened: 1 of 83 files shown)


gotson:
@xerial by binding Spotless to the `compile` phase, the CI running on Ubuntu JDK 8 fails. What do you want to do about it ?

Options I see:
1. drop JDK 8 from CI
2. bind spotless to the `verify` phase, and run a separate CI job to check formatting

The problem with option 2 is that people could very much commit and push code without running spotless, so we would catch the formatting at CI only, making the feedback loop longer.

gotson:
One more solution would be to compile in CI using jdk 11, but run the ci job with jdk8. That would probably mean splitting the job in 2, and store artifacts temporarily.

xerial:
I prefer
> 2. bind spotless to the verify phase, and run a separate CI job to check formatting

gotson:
> I prefer
> 
> > 1. bind spotless to the verify phase, and run a separate CI job to check formatting

I changed it, please have a look when you have time 🙏🏻

xerial:
LGTM. Thanks for fixing the long standing issue :)

PR #654, opened 2021-08-31
Makefile: remove `-it`/`-ti` flags from docker calls

lummax:
These control whether to run docker in interactive mode and attach the tty,
but may prohibit to run the build in CI environments where no tty is available
(e.g. Jenkins).


xerial:
Thanks!

PR #655, opened 2021-09-01
Update cross prefix following recent dockcross updates

gotson:
- update cross prefix for Linux arm64, following recent dockcross changes. see: https://github.com/dockcross/dockcross/pull/547/files#diff-c54899d03cb7216b8fd1f26f7d3596963d167b8954f83c9c26f33d9d4b84470aR19
- update strip for Linux Android arm, following recent dockcross changes. see: dockcross/dockcross#570 and dockcross/dockcross#568
- simplify the cross prefix, as the binaries are available in the PATH


xerial:
Thanks for the fix!

PR #657, opened 2021-09-02
migrate Github Actions to setup-java@v2 with built-in maven cache

gotson:
I think the existing cache action was not doing anything, but i am not sure. The new v2 action has built-in caching for maven dependencies, this should speed up the builds.


xerial:
Good to know that. Thanks!

PR #660, opened 2021-09-10
Using TCK Tested JDK builds & fixed versions of OpenJDK

carldea:
The AdoptOpenJDK has been discontinued since July 2021. When using Zulu you get all the latest updated (TCK Tested) builds for all versions of OpenJDK. Also added fixed (major) versions to the list of JDKs to build. This is good practice when customers or vendors are at fixed versions in production and using the latest can / has broken. If the fixed version is green (passed) and the latest is red (test failed) that means it isn’t your code that broke, but the latest build was the cause of the issue. 

As AdoptOpenJDK -> Adoptium -> Eclipse Foundation has suggested Temurin builds (which also are nice TCK tested builds) they do not support every version of OpenJDK such as many non-LTS versions and fixed version prior to july/aug 2021.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest, windows-latest, macos-latest ]
-        java: [ 8, 11, 16 ]
+        java: [ 8, 8.0.192, 11, 11.0.3, 16 ]
         exclude:
           - os: windows-latest
             java: 11
@@ -55,7 +55,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-java@v2
         with:
-          distribution: 'adopt'
+          distribution: 'zulu'
           java-version: ${{ matrix.java }}
           java-package: jdk
           cache: 'maven'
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest, windows-latest, macos-latest ]
-        java: [ 8, 11, 16 ]
+        java: [ 8, 8.0.192, 11, 11.0.3, 16 ]
```
xerial, line 43: No need to add latest head versions of JDK here. Testing against the latest stable versions is preferred:
```suggestion
        java: [ 8, 11, 16 ]
```

.github/workflows/test.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-java@v2
         with:
-          distribution: 'adopt'
+          distribution: 'zulu'
```
xerial, line 58:
```suggestion
          distribution: 'temurin'
```

xerial:
Thanks for the information. I learned a lot about recent changes around AdoptOpenJDK. Using the latest head versions of JDK is not my concern (e.g., 11.0.3), so just switching to `temurin` looks more natural to me. 

Let me try `temurin` first.

xerial:
It seems temurin build has no problem. I'll go with it. Thanks

carldea:
You're welcome. Glad to have helped contribute. :-)

PR #665, opened 2021-09-13
Fix file descriptor (handle) leak when opening database fails

### LINKED ISSUE #431: File descriptor leaked when opening an illegal file as sqlite database
Reproducer:
  * open a random .exe file as db (probably works with other non-sqlite files as well) using SQliteConfig.createConnection("jdbc:sqlite:path-to-file.exe")
  * catch the resulting Exception (Sqlite ErrorCode is NotADb) and keep the java process running

Result:
  * with a process inspector or similar you should be able to see, that the java process still has the db-file open
  * on windows it will not be possible to delete the file

I have tracked the problem down to the SQLiteConnection-constructor with code similar to this:

    public SQLiteConnection(String url, String fileName, Properties prop) throws SQLException {
        this.db = open(url, fileName, prop);
        SQLiteConfig config = db.getConfig();
        this.connectionConfig = db.getConfig().newConnectionConfig();
        config.apply(this);
    }

the file is opened in the open() call and the exception is thrown in config.apply(). Also there seems to be no GC finalizer that cleans up when the DB-instance is collected.

Andy-2639:
Fixes #431

### DIFF
--- a/src/main/java/org/sqlite/SQLiteConnection.java
+++ b/src/main/java/org/sqlite/SQLiteConnection.java
@@ -55,11 +55,22 @@ public abstract class SQLiteConnection implements Connection {
      * @throws SQLException
      */
     public SQLiteConnection(String url, String fileName, Properties prop) throws SQLException {
-        this.db = open(url, fileName, prop);
-        SQLiteConfig config = db.getConfig();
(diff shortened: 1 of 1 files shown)


xerial:
Looks good. Thanks

PR #666, opened 2021-09-25
Use more specific types when fetching JDBC3ResultSet.getBigDecimal

Tapac:
For now when fetching BigDecimal via JDBC3ResultSet.getBigDecimal the string value representation is used and in some cases the precision is loosed when underneath type if Float or Decimal.

In the PR the decision on how to build BigDecimal is made based on the real column type (integer and decimal types are checked).


eirikbakke:
I think this patch might have broken the case where the value is null; getBigDecimal now returns zero instead of null in that case.

gotson:
> I think this patch might have broken the case where the value is null; getBigDecimal now returns zero instead of null in that case.

please don't comment on closed issues/PR, especially 1y+. Please open a new issue, mention this PR, and provide some self-contained unit test that would exhibit the incorrect behaviour. Thanks.

PR #667, opened 2021-09-25
[graalvm] Add missing class to config

ericdallo:
Just add a missing class to be initialized during build time for graalvm config works

Tested with [clojure-lsp](https://github.com/clojure-lsp/clojure-lsp/)


ericdallo:
Closing this as this specific config belongs to [next.jdbc](https://github.com/seancorfield/next-jdbc)

PR #675, opened 2021-10-05
detect musl by checking /proc/self/map_files

### LINKED ISSUE #623: Problem with alpine-musl detection
The Alpine Linux detection introduced in #480 attempts to detect alpine from `/etc/os-release` and proceeds to use a dynamic library linked against musl.

There are two issues with this:
1) Alpine can also be used with the glibc libc instead of musl and in this case the detection succeeds but the library loading fails:
```
Failed to load native library:sqlite-unknown-a4e57938-df9c-49cf-9303-b1b135ed31f2-libsqlitejdbc.so. osinfo: Linux-Alpine/x86_64
java.lang.UnsatisfiedLinkError: /opt/x/temp/sqlite-unknown-a4e57938-df9c-49cf-9303-b1b135ed31f2-libsqlitejdbc.so: libc.musl-x86_64.so.1: cannot open shared object file: No such file or directory
Caused by: java.lang.Exception: No native library found for os.name=Linux-Alpine, os.arch=x86_64, paths=[/org/sqlite/native/Linux-Alpine/x86_64:/usr/java/packages/lib/amd64:/usr/lib64:/lib64:/lib:/usr/lib]
        at org.sqlite.SQLiteJDBCLoader.loadSQLiteNativeLibrary(SQLiteJDBCLoader.java:367)
```

Maybe it would make more sense to detect musl libc via `/proc/self/map_files` matching against `ld-musl-` or a similar method

2) The detection attempts to match the key-value pair ID= from `/etc/os-release` by

[Process p = Runtime.getRuntime().exec("cat /etc/os-release | grep ^ID");](https://github.com/xerial/sqlite-jdbc/compare/3.30.1...3.31.1#diff-a038cb61e9d7bec1708a7153e5daf3867d335a128b2df5c28636dc14585912d4R118)

This does not actually work since pipes are a feature of the shell, not `execve` and ends up matching against the contents of the entire file.

```
[pid 32176] execve("/bin/cat", ["cat", "/etc/os-release", "|", "grep", "^ID"], 0x7ffd8568e3c8 /* 18 vars */ <unfinished ...>
```

gotson:
This implements the Musl detection suggested by @Lauri-Nomme in #623, by checking the names of the libs loaded by the JVM in `/proc/self/map_files`. If any of the file name contains `musl`, then we conclude it's running on musl.

I also renamed the `Linux-Alpine` directory to `Linux-Musl`, to make it consistent.

It's been tested on the Alpine image provided by @Lauri-Nomme, and a regular Alpine distro, and works as expected.

Closes #623


src/main/java/org/sqlite/util/OSInfo.java
```diff
@@ -109,15 +114,26 @@ public static boolean isAndroid() {
...
+                                            return "";
+                                        }
+                                    })
+                            .collect(Collectors.toList());
+            if (mapFilesNames.stream().anyMatch(s -> s.toLowerCase().contains("musl"))) {
+                return true;
+            }
+        } catch (IOException ignored) {
         }
+        return false;
```
PromanSEW, line 136:
```suggestion
            return dirStream
                            .map(
                                    path -> {
                                        try {
                                            return path.toRealPath().toString();
                                        } catch (IOException e) {
                                            return "";
                                        }
                                    })
                            .anyMatch(s -> s.toLowerCase().contains("musl")));
            }
        } catch (IOException ignored) {
        }
        return false;
```

src/main/java/org/sqlite/util/OSInfo.java
```diff
@@ -109,15 +114,26 @@ public static boolean isAndroid() {
...
+                                        try {
+                                            return path.toRealPath().toString();
+                                        } catch (IOException e) {
+                                            return "";
+                                        }
+                                    })
+                            .collect(Collectors.toList());
+            if (mapFilesNames.stream().anyMatch(s -> s.toLowerCase().contains("musl"))) {
+                return true;
+            }
```
PromanSEW, line 133:
```suggestion
            return dirStream
                            .map(
                                    path -> {
                                        try {
                                            return path.toRealPath().toString();
                                        } catch (IOException e) {
                                            return "";
                                        }
                                    })
                            .anyMatch(s -> s.toLowerCase().contains("musl")));
```

Issue #509, opened 2020-06-29
Infrequent SIGSEGV crash - pthread_mutex_lock+0x4


tildedave:
Hi, we've been seeing infrequent crashes on xerial/sqlite-jdbc 3.31.1.  This happens really infrequently (about once a week) so pinning it down has been difficult.  I also haven't been able to get a working core dump here yet.  Creating this issue to try to see if anyone else has seen anything like this.  

# Extra info

* Previous version was xerial/sqlite-jdbc 3.28.1-SNAPSHOT - no issues.  I have been making a number of changes to how our application interacts with SQLite over the last few weeks so I don't know that this issue is related (or not) to the upgrade.
* Have upgraded to 3.32.3 (also added `-g` to the SQLite compilation options in Makefile.common) to see if that would change anything and no dice.
* We are running a few local custom modifications to the library but nothing really related to this.

If I get more information or can reproduce this consistently I'll update the issue and submit a PR.

# Toplevel crash info

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f6e7cca9d44, pid=438, tid=0x00007f6d2d66c700
#
# JRE version: Java(TM) SE Runtime Environment (8.0_171-b11) (build 1.8.0_171-b11)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.171-b11 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [libpthread.so.0+0x9d44]  pthread_mutex_lock+0x4
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
```

# Investigation

The crash report itself doesn't look too useful and it happens while a bunch of other work is being done in the system so including it verbatim seems like it would just be noise.  

What I did pin down from it is this:
```
Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
J 4947  org.sqlite.core.NativeDB.step(J)I (0 bytes) @ 0x00007f6e65d5ed18 [0x00007f6e65d5ece0+0x38]
```

[...]

Issue #680, opened 2021-10-19
JVM crash due to sqllite jdbc 3.36.0.3


zaidsultan:
Hello, The JVM has crashed multiple times due to sqlite jdbc latest version 3.36.0.3. Please check the crash log.
[hs_err_pid54421.log](https://github.com/xerial/sqlite-jdbc/files/7375931/hs_err_pid54421.log)

pyckle:
Hey. I see that the seg fault happened in the free native methods. Is it possible that you have a thread trying to close a prepared statement at the same time another is closing a connection? This could result in a double free (and subsequent seg fault)

See: https://github.com/xerial/sqlite-jdbc/issues/509 and https://github.com/xerial/sqlite-jdbc/pull/685

gotson:
@zaidsultan can you try the latest snapshot as it contains the fix from #685

gotson:
Should be closed by #685, released as part of `3.39.2.0`. I will be closing this, please comment if the problem happens again on `3.39.2.0`.

PR #685, opened 2021-11-07
Wrap Statement Pointers to prevent use after free race

pyckle:
If Prepared Statements are misused in multiple threads (as can happen with a buggy prepared statement pool implementation), an application segmentation fault can occur because of a use after free condition.

Two unit tests that previously reliably caused segmentation faults. One uses a prepared statement while closing the connection in another thread. The other unit test uses the prepared statement, closing the prepared statement itself in another thread.

This fixes https://github.com/xerial/sqlite-jdbc/issues/509


src/main/java/org/sqlite/core/CoreStatement.java
```diff
@@ -37,7 +37,7 @@ protected CoreStatement(SQLiteConnection c) {
         rs = new JDBC4ResultSet(this);
     }
 
-    public DB getDatbase() {
+    public DB getDatabase() {
```
TassSinclair, line 40: 😁  Nice.

pyckle:
@gotson  - please consider this for inclusion before releasing 3.39.2 Quite a few people (including myself) have hit this bug, which results in a cryptic jvm crash #719 #707 #680 and #509

It's java only changes, so no need to recompile the binaries.

This PR includes a unit test which reproduces the seg fault when the java code changes are not applied.

gotson:
That's on my list, along with a few others.

I want to get the automatic snapshot published to the Maven snapshot repo too, so people can give a try to the fixes and confirm that it works.

gotson:
Thanks a lot @pyckle for this PR. I am working to get snapshots published, so people can confirm if it fixes the issues #719 #680 #509

PR #691, opened 2021-11-21
update OSInfo.java

ChanthMiao:
* add ppc64el/ppc64le mapping

* recompile OSInfo.class with JDK 1.8

* delete old 'lib/org/sqlite/OSInfo.class'

### DIFF
--- a/src/main/java/org/sqlite/util/OSInfo.java
+++ b/src/main/java/org/sqlite/util/OSInfo.java
@@ -84,6 +84,8 @@ public class OSInfo {
         archMapping.put("powerpc64", PPC64);
         archMapping.put("power_pc64", PPC64);
         archMapping.put("power_rs64", PPC64);
+        archMapping.put("ppc64el", PPC64);
+        archMapping.put("ppc64le", PPC64);
     }
 
     public static void main(String[] args) {


gotson:
The class files should not even be here, I think they are not used and should be deleted completely.

ChanthMiao:
> The class files should not even be here, I think they are not used and should be deleted completely.

https://github.com/xerial/sqlite-jdbc/blob/ada2b147dc858ef557a995e2f4ef25f7c65fc6a7/Makefile.common#L20-L24

It seems that the class files are referenced here.

gotson:
Looks like it's used only when `OS_NAME` and `OS_ARCH` are not passed, so probably when using `make native`.

IMHO we should compile osinfo if needed when those variables are not needed, instead of committing them to source control.

ChanthMiao:
You are right. We should not commit class files to git repo

ChanthMiao:
@gotson I just made a new PR https://github.com/xerial/sqlite-jdbc/pull/697, which deleted precompiled OSInfo.class safely.

gotson:
@ChanthMiao looking at the existing code in master, it should work fine.

The only problem is that that class files must have been committed by error.

I would suggest a single PR that deletes the existing class, and adds a `.gitignore` entry to avoid committing again in the future.

Separating your PR about ppc64le in another PR would also make the PR clearer and easier to merge IMHO.

ChanthMiao:
It was not commited by error. The OSInfo.class was commit by https://github.com/xerial/sqlite-jdbc/commit/f09fb1ea13fb5ab17cdda50d64950f040e023eef

Then it Integrated by the other commit https://github.com/xerial/sqlite-jdbc/commit/6d3bfb76b43218753833f25a9674b278a9e10882

ChanthMiao:
> @ChanthMiao looking at the existing code in master, it should work fine.
> 
> The only problem is that that class files must have been committed by error.
> 
> I would suggest a single PR that deletes the existing class, and adds a `.gitignore` entry to avoid committing again in the future.
> 
> Separating your PR about ppc64le in another PR would also make the PR clearer and easier to merge IMHO.

The new PR, which  deletes the existing class is here https://github.com/xerial/sqlite-jdbc/pull/698.

If it merged, I would make ther other one to add ppc64le mapping

gotson:
@ChanthMiao i will merge #715 soon that will handle OSInfo on the fly compilation.

Can you cleanup this PR to keep only the ppc64el/ppc64le mapping ? This should help to close #450

ChanthMiao:
> 

I may do it this weekend.

ChanthMiao:
@gotson Done!

PR #707, opened 2022-01-06
Multidatabase Memory Leak and Bug Fixes

pyckle:
Correctness fixes:
* Multiple databases/threads with different handlers is now handled correctly
  * Busy Handler
  * Progress Handler
  * Update Handler
  * Commit Handler
* Utf8 UDF and Collation names now properly length checked in Utf8 bytes

Memory Leak Fixes:
* Handlers now correctly free'd on Connection close, and replacement
  * Busy Handler
  * Progress Handler
  * Update Handler
  * Commit Handler
* User Defined Functions (UDF) are free'd on removal
* Collations are now properly free'd on removal

Race Condition fixes:
* Eliminate possible race condition with static method id assignment from
word tearing. Now all methods ids are loaded upon library load

Miscellanous fixes:
* Remove unused nargs argument from destroy function
* Remove redundant null checks before instanceof checks
* Change reference conversion to use memcpy to avoid possible aliasing bug


gotson:
I fear that PR will be difficult to review, there's a lot of code changes, a single commit, and no comments whatsoever.

Do all those fixes have to be bundled together? Or could they be split into separate commits at least, or even better separate PRs?

pyckle:
> I fear that PR will be difficult to review, there's a lot of code changes, a single commit, and no comments whatsoever.
> 
> Do all those fixes have to be bundled together? Or could they be split into separate commits at least, or even better separate PRs?

The changes fall info three categories:
- Remove static C variables because they are not safe if more than one Database is opened. Rather, store this state in pointers in the NativeDB class, just like the DB pointer is stored there.
- Move all static method variables which are set only once into global static variables initialized on library load to prevent word tearing
- Minor tiny bug fixes in the java code, as documented in the PR.

Once you realize that these are the majority of the changes, the commit becomes easier to decipher. With that said, I agree it is a lot of a single commit.

I will do the following:
- Separate the minor bug fixes into a separate commit. Unfortunately, because the first two parts change mostly native code, I don't think it can be pulled apart without significant additional effort.
- I will review my changes and find places to add more comments.

I struggled to add clear comments in the code to justify the static variable changes. A comment next to a stored pointer in the NativeDB class that "this can't be stored in a global C variable because there may be multiple databases" doesn't make sense, unless you knew that it was previously attempted to be stored there. Perhaps the best place to document this is the unit tests which tests this scenario.

gotson:
Thanks, that already gives a lot of context. 

> A comment next to a stored pointer in the NativeDB class that "this can't be stored in a global C variable because there may be multiple databases" doesn't make sense, unless you knew that it was previously attempted to be stored there.

Agreed.

A better way might be to have some kind of general style guide about the native C code. Xerial may have a different opinion.

> Perhaps the best place to document this is the unit tests which tests this scenario.

This is always a good practice, on top of failing if behavior changes, tests are good living documentation that can explain the expected and failing behaviors.

pyckle:
> A better way might be to have some kind of general style guide about the native C code. Xerial may have a different opinion.

I agree, but I don't feel comfortable writing a best practices guide for a project that I didn't write. I'm happy to review one, or give input.

> This is always a good practice, on top of failing if behavior changes, tests are good living documentation that can explain the expected and failing behaviors.

I completely agree. They are also crucial in trying to understand the projects code. Its much easier to step through a unit code, than to read code without context.

I wrote unit tests for the bugs that I fixed when possible (testing word tearing in C with java unit tests isn't feasible for example). I made sure that the unit tests that I added fail (with seg faults, exceptions, or bad assertions) before the bug fixes.

As as side note, in this bug fix PR (https://github.com/xerial/sqlite-jdbc/pull/685), the unit test reliably produces a seg fault from a race condition in less than a second. It took a while to get this distilled and run this quickly, but in my opinion, for the above reasons, it's time very well spent.

pyckle:
Hey. I separated this into two commits and added some comments to the tests. Feedback is welcome (:

src/main/java/org/sqlite/Function.java
```diff
@@ -94,9 +94,9 @@ public static final void create(Connection conn, String name, Function f, int fl
      * @param nArgs The number of arguments that the function takes.
      * @param flags Extra flags to pass, such as {@link #FLAG_DETERMINISTIC}
      */
-    public static final void create(Connection conn, String name, Function f, int nArgs, int flags)
```
gotson, line 97: why remove the `final` here?

src/main/java/org/sqlite/Function.java
```diff
@@ -127,8 +123,8 @@ public static final void create(Connection conn, String name, Function f, int nA
      * @param nArgs The number of args for the function.
      * @throws SQLException
      */
-    public static final void destroy(Connection conn, String name, int nArgs) throws SQLException {
```
gotson, line 130: why remove the `final` here?

src/main/java/org/sqlite/Function.java
```diff
@@ -141,7 +137,7 @@ public static final void destroy(Connection conn, String name, int nArgs) throws
      * @param name The name of the function.
      * @throws SQLException
      */
-    public static final void destroy(Connection conn, String name) throws SQLException {
```
gotson, line 144: why remove the `final` here?

src/main/java/org/sqlite/ProgressHandler.java
```diff
@@ -14,9 +14,9 @@
      * @param progressHandler the progressHandler
      * @throws SQLException
      */
-    public static final void setHandler(
-            Connection conn, int vmCalls, ProgressHandler progressHandler) throws SQLException {
-        if (conn == null || !(conn instanceof SQLiteConnection)) {
+    public static void setHandler(Connection conn, int vmCalls, ProgressHandler progressHandler)
```
gotson, line 17: why remove the `final` here?

src/main/java/org/sqlite/ProgressHandler.java
```diff
@@ -32,7 +32,7 @@ public static final void setHandler(
      * @param conn the SQLite connection
      * @throws SQLException
      */
-    public static final void clearHandler(Connection conn) throws SQLException {
```
gotson, line 35: why remove the `final` here?

gotson:
> Hey. I separated this into two commits and added some comments to the tests. Feedback is welcome (:

It does help a lot in the review ! My C skills are really poor so i cannot review the changes you've done in the JNI part unfortunately.

src/main/java/org/sqlite/Function.java
pyckle: Hey. Sorry for the long time to respond - it's been busy. I removed it because my IDE (IntelliJ) suggested it is useless for a method to be both static and final.

I now see that this isn't entirely true. If you have a class with a static final method, a subclass that implements its own static final method of the same name won't compile. It's a silly use case though, because there is no polymorphism for static methods, so preventing a subclass from overloading create with these exact parameters doesn't seem to serve much purpose.

Nonetheless, I see that there are this syntax does change the code meaning, and as I'm not the author of this codebase it's not my place to make code style decisions and this definitely shouldn't be packaged into a bugfix MR. I will remove this from the PR shortly.

With that said, I am curious to know the rational for adding this keyword here. Is it just good style to denote that this method shouldn't be changed? Or maybe some IDEs offer to override this static function (in error) and final helps them know not to?

pyckle:
Thank you for the review @gotson and sorry for the delay at looking at it. I've been quite busy lately.

gotson:
@pyckle seems this was never run through CI for some reason. I updated your branch and it fails in CI.

pyckle:
Hey @gotson - this requires a rebuild of the jni libs because it fixes bugs in the JNI code. The new tests fail with the old JNI lib because it has bugs.

I will try and find some time to look at this again next week. Thanks for all your work reviewing the PRs. It's really hard work.

gotson:
> this requires a rebuild of the jni libs because it fixes bugs in the JNI code. The new tests fail with the old JNI lib because it has bugs.

You're right, completely forgot about that 🤦🏻

No rush to update, this will need some serious reviewing!

src/main/java/org/sqlite/Function.java
pyckle: I removed this change. This change was done because intellij thinks that declaring static methods as final is useless, as polymorphism isn't possible with static methods.

Looking into this a bit more, it seems that it does actually do something - if you extend the class Function, it will prohibit you from making a method on it called destroy, static or otherwise.

I think this is a stylistic choice, and therefore keeping the style of the project makes sense. And if the style was to change, it certainly shouldn't be pushed in as part of a bugfix.

pyckle:
New Binaries included (including for m1 mac), and everything passes CI. This is ready for review.

gotson:
@pyckle seems like there's some new conflicts

pyckle:
> @pyckle seems like there's some new conflicts

Rebased (:

src/test/java/org/sqlite/ListenerTest.java
```diff
@@ -82,108 +88,105 @@ public void onUpdate(Type type, String database, String table, long rowId) {
...
+        connectionTwo.addCommitListener(commitListener2);
+
+        try (Statement statement = connectionOne.createStatement()) {
+            statement.execute("INSERT INTO sample (description) VALUES ('smert za smert')");
         }
 
+        // update should be synchronous
+        List<UpdateEvent> updates = listener1.getAllUpdates();
+
         if (updates.isEmpty()) throw new AssertionError("Never got update!");
```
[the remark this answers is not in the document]
gotson: don't mind me, that was not your code 🤦🏻

gotson:
@pyckle i just reviewed this, that's a solid contribution ! As i said i'm not very good with C, but with all the explanations you gave, the tests, and the commits broken down into smaller size, that gives me confidence to merge this.

gotson:
@pyckle there was a slight change in the build process in https://github.com/xerial/sqlite-jdbc/commit/c2412cf673975437c7b5cb6a06816a2afb30ec2c, but your rebase was done prior to that.

Do you think you could rebase, and rebuild the binaries ?

Also, i would like to keep your commits as-is, instead of squashing them, as they give a better idea of what was done. We are now trying to use conventional commits, so could you change the titles of the commits like so:
- `fix: minor bug fixes and code cleanup`
- `fix: various correctness, memory leak and race condition fixes`
- `chore: update binaries`

pyckle:
Hey @gotson 
I rebased, rebuilt and reworded commit titles. Please let me if it looks OK/if anything else is needed.

Thank you for the good work maintaining this project.

gotson:
> I rebased, rebuilt and reworded commit titles. Please let me if it looks OK/if anything else is needed.

thanks a lot, speedy work 🚀 

> Thank you for the good work maintaining this project

and thank you for the amazing contributions you've made recently, that really helps !

PR #728, opened 2022-04-14
Upgrade to 3.38.0

pyckle:


gotson:
@pyckle is this current ?

gotson:
@pyckle i see 3.39 was released in the meantime, should we just try to upgrade to that instead ?

pyckle:
Yes, please see https://github.com/xerial/sqlite-jdbc/pull/746 - I included updated binaries. It builds cleanly.

Issue #737, opened 2022-05-30
Please support the Data Service Specification for JDBC™ Technology


laeubi:
OSGi defines a [Data Service Specification for JDBC™ Technology](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.jdbc.html) that provides a convenient and save way to obtain DataSources.
Even though it can be implemented by clients it is advised that JDBC driver vendors include an implementation if possible for the following reasons:

1. The vendor has the best knowledge how to create and configure the participating objects.
2. Changes in the internal implementation, e.g. refactoring of class names can be reflected an managed easier in the original repository
3. No gap between deployment of new version and connector code
4. For non-osgi users, this is completely transparent and the is no need for them to change, but OSGi users will be happy to not reinvent the wheel, the library can just be dropped inside a framework and seaming less integrate, for example with the [JPA Specification](http://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.jpa.html). 

Because of this I'd like to propose adding an implementation for a [DataSourceFactory](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.jdbc.html#org.osgi.service.jdbc.DataSourceFactory) in this repository and provide a PR for this.

FYI @stbischof

laeubi:
@xerial I already created a PR that shows the basic changes for a first version, can you take a look and let me know what you think?

gotson:
Hi, could you provide context on osgi? Who's using it, how, etc? I any other jdbc drivers support this?

laeubi:
> Hi, could you provide context on osgi?

You can find a short introduction here: http://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#d0e1755
Actually, beside some implementation of a standard java-interface and some extra headers in the manifest nothing will change, so an OSGi-Bundle is still a standard Jar, the only difference is that it could be used in plain java and OSGi without any effort.

> Who's using it, how, etc?

Mayne people are using it including Eclipse-IDE and many big companies see for example https://www.osgi.org/membership/members/

> I any other jdbc drivers support this?

The following JDBC drivers already offer support for OSGi Data Service Specification for JDBC™ Technology:

- https://github.com/microsoft/mssql-jdbc
- https://github.com/h2database/h2database
- https://github.com/mariadb-corporation/mariadb-connector-j
- https://github.com/pgjdbc/pgjdbc
-  ... maybe more I'm not aware of yet ...

PR #738, opened 2022-05-30
Support for the Data Service Specification

laeubi:
Add a first basic implementation of the DataSourceFactory


gotson:
@laeubi can you check the failing CI?

laeubi:
@gotson as it complains about  code formatting is ther an Eclipse format profile or something I can use to format the source code?

gotson:
> @gotson as it complains about code formatting is ther an Eclipse format profile or something I can use to format the source code?

You can use the Spotless plugin: `mvn spotless:check` and `mvn spotless:apply`

laeubi:
Done!

gotson:
This has no unit tests, is this feasible to add some ?

laeubi:
I'll try to add test-cases here, but have first to check how sqlite tests are structured.

gotson:
I've checked the issue you raised at osgi/osfgi-test and i think you framed it quite well: how to make it easier for maintainers that know nothing and have no particular interest in OSGI to gauge a PR and have confidence to merge it.

That probably will take a bit of time, but i think it's the right approach.

laeubi:
@gotson it hase taken a bit longer than desired but now I finally was able to add a small test-case, do you like to review and give some feedback if that would suffice?

gotson:
The more i think of it, the more i wonder whether this needs to be part of `sqlite-jdbc`. Could it not be part of a separate project that would offer that functionality only ? And if someone wants OSGi support they could add one more dependency in their `pom.xml` / `build.gradle`.

laeubi:
As always there are many options but in general this has several drawbacks, as already described here:
- https://github.com/xerial/sqlite-jdbc/issues/737

1. The vendor has the best knowledge how to create and configure the participating objects.
2. Changes in the internal implementation, e.g. refactoring of class names can be reflected an managed easier in the original repository, maybe impose extra restrictions / problems (e.g. in this case a project needs to either embeds the `sqlite-jdbc` or republish it to add appropriate OSGi metadata)
3. No gap between deployment of new version and connector code
4. For non-osgi users, this is completely transparent and the is no need for them to change, but OSGi users will be happy to not reinvent the wheel, the library can just be dropped inside a framework and seaming less integrate, for example with the [JPA Specification](http://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.jpa.html).

Beside that an extra project always needs to be found, needs extra release management and so on

laeubi:
It seems I missed formatting the test code, fixed now ...

PR #746, opened 2022-07-26
Upgrade to 3.39.2

pyckle:


gotson:
Saw this in the release notes:

> The JSON functions are now built-ins. It is no longer necessary to use the [-DSQLITE_ENABLE_JSON1](https://www.sqlite.org/compile.html#enable_json1) compile-time option to enable JSON support. JSON is on by default. Disable the JSON interface using the new [-DSQLITE_OMIT_JSON](https://www.sqlite.org/compile.html#omit_json) compile-time option.

We can probably remove this: https://github.com/xerial/sqlite-jdbc/blob/ada2b147dc858ef557a995e2f4ef25f7c65fc6a7/Makefile#L81

gotson:
I see there is an updated binary for Mac aarch64, just wanted to check, did you compile that on your M1 mac @pyckle ?

gotson:
The build is failing on armv7 and aarch64. Just to be sure i compared with [a recent build](https://github.com/xerial/sqlite-jdbc/actions/runs/2744501281) from the current master, and those work, so this must be because of the new changes in this PR.

We need to find out why this happens with this PR.

gotson:
> Failed to load native library:sqlite-3.39.2-759e4a1a-cc92-4003-88d4-08cff1551a91-libsqlitejdbc.so. osinfo: Linux/armv7
java.lang.UnsatisfiedLinkError: /tmp/sqlite-3.39.2-759e4a1a-cc92-4003-88d4-08cff1551a91-libsqlitejdbc.so: /lib/arm-linux-gnueabihf/libc.so.6: version `GLIBC_2.33' not found (required by /tmp/sqlite-3.39.2-759e4a1a-cc92-4003-88d4-08cff1551a91-libsqlitejdbc.so)

This is weird that is needs a version of Glibc that is very recent. Dockcross should have `-lts` versions which use a lower GLibc, i think we need to switch to that. We've already done that for arm-v6, but since then they extended the `-lts` to other cpu architecture.

gotson:
I see `docker/dockcross-armv7` is replaced with an empty file, can we delete it altogether, since we are using `armv7a` ?

pyckle:
The M1 build was built on my M1 Mac. It was tested.

I just moved armv7a to lts, and that should only requires glibc 2.28+ according to the dockcross docs.

I will make the same change for aarch64 and get that pushed shortly.

I will also remove the JSON build def, and rebuilt all other binaries.

gotson:
> The M1 build was built on my M1 Mac. It was tested.
Awesome, I was not sure you had an M1. 
> I just moved armv7a to lts, and that should only requires glibc 2.28+ according to the dockcross docs.
> 
> I will make the same change for aarch64 and get that pushed shortly.
I think we should use all the available lts to avoid issues.

pyckle:
arm64 is now moved to lts, and new binaries are pushed I will do it with the rest of the archs in the next couple of days.

If the lts toolchains don't make the build pass, we'll have to hardcode an older version of the dockcross docker images.

pyckle:
Actually, the lts versions are only for arm and mips. All arm builds were moved to lts, and we don't build for mips.

So I think this should be good, unless you'd like me to cleanup the commit log/squash things.

gotson:
> So I think this should be good, unless you'd like me to cleanup the commit log/squash things

I will squash on merge, don't worry. Will review tomorrow 👍🏻

Issue #656, opened 2021-09-01
[IO performance] setDefaultUseCaches(false) impacts all URL connections


gwenn:
Using [setDefaultUseCaches(false)](https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java/org/sqlite/SQLiteJDBCLoader.java#L276) impacts all URL connections.
Because [defaultUseCaches](https://github.com/openjdk/jdk/blob/739769c8fc4b496f08a92225a12d07414537b6c0/src/java.base/share/classes/java/net/URLConnection.java#L226) is static (see [setDefaultUseCaches](https://github.com/openjdk/jdk/blob/739769c8fc4b496f08a92225a12d07414537b6c0/src/java.base/share/classes/java/net/URLConnection.java#L1064) implementation).
Maybe you should use [setUseCaches](https://github.com/openjdk/jdk/blob/739769c8fc4b496f08a92225a12d07414537b6c0/src/java.base/share/classes/java/net/URLConnection.java#L1000) instead.

For example, after `setDefaultUseCaches(false)`, each time a new class is loaded from a signed JAR, [initializeVerifier](https://github.com/openjdk/jdk/blob/c36755dedf1a0d7ce0aeadd401e0c70ff84185e7/src/java.base/share/classes/java/util/jar/JarFile.java#L738) is recomputed.

Regards.

gotson:
FYI this was done as part of https://github.com/xerial/sqlite-jdbc/pull/578 IIRC

AuroraLS3:
`setDefaultUseCaches(false)` causes class loading to fail & crash in some Minecraft server implementations since many use URLClassLoader for loading plugins/mods

gotson:
Anyone willing to submit a PR for this?

AuroraLS3:
Opened a PR.

gotson:
This is available in the latest snapshot, please give it a try.

PR #755, opened 2022-07-29
Use setUseCaches instead of setDefaultUseCaches

AuroraLS3:
Use of setDefaultUseCaches impacts all URLClassLoaders and HttpURLConnections, which can lead to erratic behavior of non-sqlite components running on a system since the variable is static.

Related issue: #656


gotson:
Thanks!

PR #758, opened 2022-08-04
fix: fix Connection.getMetaData().getColumns returns wrong/fixed values for column data type details (#733)

### LINKED ISSUE #733: Connection.getMetaData().getColumns returns wrong/fixed values for column data type details
Using the `ResultSet` returned by `Connection.getMetaData().getColumns()` returns wrong values for the size of the column.

Especially these values are wrong:
- `COLUMN_SIZE` (fixed `10`)
- `DECIMAL_DIGITS` (fixed `2000000000`)

The `DATA_TYPE` is correct, but the client is unable to detect the size of the column or the precision of floating point values.

_Example:_
I have a table with this column: `ORIENTATION REAL(6,3)`

This is what I get:

- `COLUMN_NAME`=`"ORIENTATION"` (correct)
- `TYPE_NAME`=`"REAL(6,3)"` (correct)
- `DATA_TYPE`=`6` (correct, is `java.sql.Types.FLOAT`)
- `COLUMN_SIZE`=`2000000000` (wrong, should be `9`?, `6+3`)
- `DECIMAL_DIGITS`=`10` (wrong, should be `3`)

The cause for this can be found [here](https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java#L920) and [here](https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java#L922).

I just did not found a simple fix, so I add this issue instead of a pull request 😉

sebasbaumh:
* provide valid COLUMN_SIZE and DECIMAL_DIGITS values for columns queried using
 Connection.getMetaData().getColumns(java.lang.String, java.lang.String, java.lang.String, java.lang.String)
* add/adjust unit tests to test this change

Fixes #733

### DIFF
--- a/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
@@ -917,9 +917,9 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
         StringBuilder sql = new StringBuilder(700);
         sql.append("select null as TABLE_CAT, null as TABLE_SCHEM, tblname as TABLE_NAME, ")
                 .append(
-                        "cn as COLUMN_NAME, ct as DATA_TYPE, tn as TYPE_NAME, 2000000000 as COLUMN_SIZE, ")
+                        "cn as COLUMN_NAME, ct as DATA_TYPE, tn as TYPE_NAME, colSize as COLUMN_SIZE, ")
                 .append(
-                        "2000000000 as BUFFER_LENGTH, 10   as DECIMAL_DIGITS, 10   as NUM_PREC_RADIX, ")
+                        "2000000000 as BUFFER_LENGTH, colDecimalDigits as DECIMAL_DIGITS, 10   as NUM_PREC_RADIX, ")
                 .append("colnullable as NULLABLE, null as REMARKS, colDefault as COLUMN_DEF, ")
                 .append(
                         "0    as SQL_DATA_TYPE, 0    as SQL_DATETIME_SUB, 2000000000 as CHAR_OCTET_LENGTH, ")
@@ -996,6 +996,10 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
                         }
                         colFound = true;
 
+						// default values
+						int iColumnSize = 2000000000;
+						int iDecimalDigits = 10;
+                        
                         /*
                          * improved column types
                          * ref http://www.sqlite.org/datatype3.html - 2.1 Determination Of Column Affinity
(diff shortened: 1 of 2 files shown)


src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -1011,14 +1015,62 @@ public ResultSet getColumns(String c, String s, String tblNamePattern, String co
...
+                                        // no decimals
+                                        iDecimalDigits = 0;
+                                        // columns size equals dimension
+                                        iColumnSize = iInteger;
+                                    }
+                                } catch (NumberFormatException ex) {
+                                    // just ignore invalid dimension formats here
+                                }
+                            }
+                        }
```
gotson, line 1073: What do you think of using regex instead of `indexOf`/`substring` ?

Would need to define this somewhere:
```java
protected static final Pattern COLUMN_LENGTH = Pattern.compile(".*\\((?<integer>[0-9]+)(,(?<decimal>[0-9]+))?\\)");
```

Then we can replace this whole block:

```suggestion
                        Matcher matcher = COLUMN_LENGTH.matcher(colType);
                        if (matcher.matches()) {
                            try {
                                iDecimalDigits = matcher.group("decimal") == null ? 0 : Integer.parseUnsignedInt(matcher.group("decimal"));
                                iColumnSize = Integer.parseUnsignedInt(matcher.group("integer")) + iDecimalDigits;
                            } catch (NumberFormatException ignored) {
                            }
                        }
```

gotson (commented):
Thanks a lot for this PR

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
sebasbaumh: Your welcome :smile:

I would not use a regex as it is much slower than using a plain `indexOf` (which is just a small loop under the hood).
And the parsing calls would still be necessary.

gotson: > I would not use a regex as it is much slower than using a plain `indexOf`

Is it? Even when the pattern is compiled only once?

sebasbaumh: Yes, it is. There is no regex evaluation, which is much more expensive than just looping through a string by characters.
Though you could replace it, if you really like to and want more concise code.
I just wanted to add very performant code with a low memory overhead (small number of objects constructed on the heap). As I thought this might be called pretty often if multiple tables are queried or tables with a lot of columns.

I would use regex for more complex expressions, but here it is just to make sure two brackets are there and distinguishing between one and two parameters. Both variants still have to parse the value(s) to integers, so that part is the same.

sebasbaumh:
Okay, I did a small benchmark. I just does not have any effect 😆

sebasbaumh:
Parsing 3 columns 1000000 times did not make much difference, maybe it just gets optimized anyway.
So just go with the regex if you like.

gotson:
I'm just maintaining the repo, it's not my code 🙃 

Will merge this as is, thanks!

sebasbaumh:
Okay and thank you :smile:

PR #763, opened 2022-08-12
Enable new math functions from 3.35.0

gotson:
This will address #633.

As mentioned in the issue, existing math functions are currently supported by user contributed code (`extension-functions.c ` from https://sqlite.org/contrib).

The https://www.sqlite.org/draft/lang_mathfunc.html introduced in 3.35.0 overlap with some of the previously existing functions.

Since the new functions can be enabled with `SQLITE_ENABLE_MATH_FUNCTIONS`, the old functions will be included only if the new math functions are not enabled. The other functions from that old contrib files would still be included.

⚠️ It is worth noting that there is a breaking change with the `log(X)` function:
- currently it performs `ln(X)`
- with the new math functions it performs `log10(X)`

Here is an explanation from the SQLite website:

> Compatibility note: SQLite works like PostgreSQL in that the log() function computes a base-10 logarithm. Most other SQL database engines compute a natural logarithm for log(). In the two-argument version of log(B,X), the first argument is the base and the second argument is the operand. This is the same as in PostgreSQL and MySQL, but is reversed from SQL Server which uses the second argument as the base and the first argument as the operand.

Given our versioning scheme, I think we should only introduce breaking changes when updating to a major version of SQLite, as users would most likely perform more regression tests on major version updates.

This PR will sit still until we can upgrade to the future 3.40.0 version of SQLite.


thipages:
As a note, I was using an old 3.39 version (with DBeaver) and got some troubles getting ln from log.
Now everything is fine. Thanks !

PR #768, opened 2022-08-17
Support for JDBC 'large' methods returning `long`

### LINKED ISSUE #735: Support for `executeLargeUpdate`
Does `sqlite-jdbc` currently supports `PreparedStatement#executeLargeUpdate`? I'm trying to use Helidon along with SQLite and I'm getting `java.lang.UnsupportedOperationException: executeLargeUpdate not implemented` so I suppose the default implementation of that method is used (which is throwing the exception). I've also browsed the code and it seems this method is not implemented. Are there any plans to do so (or some reasons for which e.g. it's not possible)?

gotson:
There are some JDBC 'large' methods that return a `long` instead of an `int`, which are not supported at the moment. That's the only difference, and it is advised to use the 'large' method if the number of updates would overflow an integer.

This PR replaces:
- `int sqlite3_total_changes(sqlite3*);` with `sqlite3_int64 sqlite3_total_changes64(sqlite3*);`
- `int sqlite3_changes(sqlite3*);` with `sqlite3_int64 sqlite3_changes64(sqlite3*);`

It will always use the large methods, and cast the `long` to `int` in the normal methods.

The SQLite website mentions that when the number of updates is greater that the integer capacity, the return value of methods returning an `int` would be undefined, so the existing behaviour wouldn't be changed much:
- before PR: the return value is undefined by SQLite itself
- after PR: the return value is casted to `int` and would lose precision

It would have been much more work to duplicate all the methods and use the appropriate `int` or `int64` method from SQLite, for probably the same results. Ultimately it is the responsibility of the user's code to call the 'large' methods if they expect the number of updates to be more than the capacity of an `int`.

Would close #735


gotson:
@gwenn would you have some time to review this ? I know you have your own jdbc driver, so your feedback would be valuable.

gwenn:
Not strictly related to this PR but javadoc of [getUpdateCount()](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#getUpdateCount--) / [getLargeUpdateCount()](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#getLargeUpdateCount--) says:
> This method should be called only once per result.

gotson:
> Not strictly related to this PR but javadoc of [getUpdateCount()](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#getUpdateCount--) / [getLargeUpdateCount()](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#getLargeUpdateCount--) says:
> 
> > This method should be called only once per result.

does that mean that the first call should return the update count, and any subsequent calls should return `-1` ?

gwenn:
> > Not strictly related to this PR but javadoc of [getUpdateCount()](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#getUpdateCount--) / [getLargeUpdateCount()](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#getLargeUpdateCount--) says:
> > > This method should be called only once per result.
> 
> does that mean that the first call should return the update count, and any subsequent calls should return `-1` ?

Good question and I guess that any subsequent calls should return `-1` but I am not sure.

gotson:
> > > Not strictly related to this PR but javadoc of [getUpdateCount()](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#getUpdateCount--) / [getLargeUpdateCount()](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#getLargeUpdateCount--) says:
> > > > This method should be called only once per result.
> > 
> > 
> > does that mean that the first call should return the update count, and any subsequent calls should return `-1` ?
> 
> Good question and I guess that any subsequent calls should return `-1` but I am not sure.

I found some information about it on the h2 discussion group and apparently it's not required.

Issue #780, opened 2022-09-06
JRE crashes on Connection#close() with added SQLiteCommitListener


h8leet:
From version 3.39.2.1, JRE crashes on Connection#close() if connection with disabled autoCommit and added SQLiteCommitListener.

**Sample code to reproduce:**
```java
    public static void main(String[] args) throws IOException {
        Path tempPath = Files.createTempDirectory("test-jdbc");
        Path dbPath = Paths.get(tempPath.toString(), "test.db");
        String dbUrl = "jdbc:sqlite:file:" + dbPath;
        try {
            Connection connection = DriverManager.getConnection(dbUrl);
            connection.setAutoCommit(false);
            SQLiteCommitListener commitListener = new SQLiteCommitListener() {
                @Override
                public void onCommit() {
                    System.out.println("COMMIT");
                }
                @Override
                public void onRollback() {
                    //
                }
            };
            ((SQLiteConnection) connection).addCommitListener(commitListener);
            connection.close();
        } catch (SQLException ex) {
            ex.printStackTrace();
        } finally {
            Files.delete(dbPath);
            Files.delete(tempPath);
        }
    }
```

**Logs Linux:**
```
# A fatal error has been detected by the Java Runtime Environment:
# JRE version: OpenJDK Runtime Environment (18.0.2+9) (build 18.0.2+9)
# Java VM: OpenJDK 64-Bit Server VM (18.0.2+9, mixed mode, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
Host: Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz, 8 cores, 31G, Manjaro Linux
---------------  T H R E A D  ---------------

Current thread (0x00007fe2d0015560):  JavaThread "main" [_thread_in_native, id=19616, stack(0x00007fe2d7700000,0x00007fe2d7800000)]

Stack: [0x00007fe2d7700000,0x00007fe2d7800000],  sp=0x00007fe2d77fe770,  free space=1017k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [sqlite-3.39.2.1-086e1707-03fe-46e5-a22d-45c498fac29d-libsqlitejdbc.so+0xda80]

Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
j  org.sqlite.core.NativeDB._close()V+0
j  org.sqlite.core.DB.close()V+76
[...]

PR #781, opened 2022-09-08
fix(native): JRE crash on connection close with CommitListener and auto commit disabled

gotson:
Closes: #780


src/main/java/org/sqlite/core/NativeDB.c
```diff
@@ -1777,7 +1777,7 @@ static void freeCommitHandlerCtx(JNIEnv *env, void *ctx) {
 
 void clear_commit_listener(JNIEnv *env, jobject nativeDB, sqlite3 *db) {
     sqlite3_commit_hook(db, NULL, NULL);
-    sqlite3_update_hook(db, NULL, NULL);
+    sqlite3_rollback_hook(db, NULL, NULL);
```
pyckle, line 1780: I think you found and fixed the problem. Nice work (:
Because we're closing the db, it may have an empty transaction that it rollsback before closing if autocommit is disabled.
Since the rollback hook was still present due to this error, it tries to call the rollback function, which is set to null due to the next line.
Then it crashes.

gotson: unfortunately the tests don't pass, while it works on my machine, and i don't know why

pyckle: The recompiled binaries were added to the incorrect directories in git, so the CI is using the old binaries.

On your local machine (and mine), when you ran make, it put the new binary in the correct directory.

pyckle: Example: Windows/x86/sqlitejdbc.dll should be src/main/resources/org/sqlite/native/Windows/x86/sqlitejdbc.dll
etc.

gotson: oh, let me have a look at that. I thought this worked well, but i may have botched it 🤦🏻

Issue #792, opened 2022-09-25
PreparedStatement::getMoreResults() closes Prepared Statement itself


hadrabap:
**Describe the bug**
I'm trying to integrate `sqlite-jdbc` into OpenLiberty. The problem is that all JDBC native objects are wrapped by OpenLiberty stuff which takes intensive care of resource clean-up. The flow is somehow like this:

1. Application uses injected DataSource, obtains Connection and prepares and executes PreparedStatement with (or without) parameters. The resource management on the application side uses `try-with-resource` pattern:
  ```
  try (var conn = dataSource.getConnection()) {
      try (var pStmt = conn.prepareStatement("SQL")) {
          pStmt.setString(1, …
          …
          try (var rSet = pStmt… {
          }
      }
  }
  ```
2. When the `try-with-resource` reaches `pStmt.close()` the wrapper does the following:
  a. calls `preparedStatement::getMoreResults()` in order to free non-consumed ResultSets
  b. calls `preparedStatement::clearParameters()` in the hope to release underlying resources (e.g. native or in the DB)
  c. finally calls `preparedStatement::close()`

The issue with `sqlite-jdbc` lies in the step (2.b.) The implementation does not support more than one result set so when `PreparedStatement::getMoreResults()` is called, the statement itself is closed. This results in an `SQLException` being thrown by the consequent call to `PreparedStatement::clearParameters()` thus rendering the whole transaction being rolled-back.

In short terms: the implementation of `PreparedStatement::getMoreResults()` should not close the Prepared Statement by itself.

**To Reproduce**
1. Prepare a PreparedStatement
2. execute the statement
3. call `PreparedStatement::getMoreResults()`
4. call `PreparedStatement::clearParameters()`

**Expected behavior**
The step (4) should pass without issues.

**Environment (please complete the following information):**
 - OS: Linux
 - CPU architecture: x86_64
 - sqlite-jdbc version 3.39.3.0

gotson:
I'm currently working on implementing multiple queries in a single statements, hence multiple result sets.

hadrabap:
Excellent! Thank you for your patience. Please take into account the resource clean-up lifecycle.

gotson:
Seems it was already raised in #458

gotson:
Closing in favor of #458

PR #793, opened 2022-09-25
Do not close Statement when checking for more Result Sets (#792)

hadrabap:
Possible fix for #792 (PreparedStatement::getMoreResults() closes Prepared Statement itself)


gotson:
@hadrabap my attempts to implement multiple result set handling have proven more complicated than expected, mostly because of the intertwining of classes.

Thus i would like to merge your change.

I did a few updates and added more tests, if you could have a look before i merge i would appreciate.

PR #796, opened 2022-10-16
Allow correct recognition of Android OS when running Termux environment

### LINKED ISSUE #789: Failed to load native library when running in Termux
This works fine on Windows, according to the README.md, it should also work adaptively on `Termux`, but it doesn't, am I doing something wrong?

**Codes**
```Kotlin
class MySQLiteTool(private val library: Path) {
    val connection: Connection by lazy {
        Class.forName("org.sqlite.JDBC")
        DriverManager.getConnection("jdbc:sqlite:$library")
    }
...
```
```KotlinScript
dependencies {
    implementation("org.xerial:sqlite-jdbc:3.39.3.0")
}
```

**Logs**
```
Failed to load native library:jansi-2.4.0-2db311a958a8e929-libjansi.so. osinfo: Linux/arm64
java.lang.UnsatisfiedLinkError: /data/data/com.termux/files/usr/tmp/jansi-2.4.0-2db311a958a8e929-libjansi.so: dlopen failed: library "libc.so.6" not found: needed by /data/data/com.termux/files/usr/tmp/jansi-2.4.0-2db311a958a8e929-libjansi.so in namespace (default)
...
...
2022-09-21 10:17:22 W/stderr: Failed to load native library:sqlite-3.36.0.3-ffd4f25b-64e0-4683-a009-8a70f7c36428-libsqlitejdbc.so. osinfo: Linux/aarch64
2022-09-21 10:17:22 W/stderr: java.lang.UnsatisfiedLinkError: /data/data/com.termux/files/usr/tmp/sqlite-3.36.0.3-ffd4f25b-64e0-4683-a009-8a70f7c36428-libsqlitejdbc.so: dlopen failed: library "libc.so.6" not found: needed by /data/data/com.termux/files/usr/tmp/sqlite-3.36.0.3-ffd4f25b-64e0-4683-a009-8a70f7c36428-libsqlitejdbc.so in namespace (default)
2022-09-21 10:17:22 E/AzurLaneLibrary: java.sql.SQLException: Error opening connection
java.sql.SQLException: Error opening connection
```

```
## Termux App Info
**APP_NAME**: `Termux`  
**PACKAGE_NAME**: `com.termux`  
**VERSION_NAME**: `0.118.0`  
**VERSION_CODE**: `118`  
**TARGET_SDK**: `28`  
## Device Info
### Hardware
**SUPPORTED_ABIS**: `arm64-v8a, armeabi-v7a, armeabi`  
```

arch-dev:
If `isAndroid()` check fails, `checkIfAndroidTermux()` is called which runs `uname -o` command and check if its output contains `android`. This should fix #789. 

Sample command output in Termux before changes:

Sample command output in Termux after changes:

### DIFF
--- a/src/main/java/org/sqlite/util/OSInfo.java
+++ b/src/main/java/org/sqlite/util/OSInfo.java
@@ -111,7 +111,12 @@ public class OSInfo {
     }
 
     public static boolean isAndroid() {
-        return System.getProperty("java.runtime.name", "").toLowerCase().contains("android");
+        boolean firstCheck =
+                System.getProperty("java.runtime.name", "").toLowerCase().contains("android");
+        if (!firstCheck) {
+            return checkIfAndroidTermux();
+        }
+        return firstCheck;
     }
 
     public static boolean isMusl() {
@@ -255,5 +260,15 @@ public class OSInfo {
 
     static String translateArchNameToFolderName(String archName) {
         return archName.replaceAll("\\W", "");
+    }
+
+    static boolean checkIfAndroidTermux() {
+        String output;
+        try {
+            output = processRunner.runAndWaitFor("uname -o");
+        } catch (IOException | InterruptedException e) {
+            return false;
+        }
+        return output.toString().toLowerCase().contains("android");
     }
 }
--- a/src/main/java/org/sqlite/util/OSInfo.java
+++ b/src/main/java/org/sqlite/util/OSInfo.java
@@ -111,12 +111,19 @@ public class OSInfo {
     }
 
     public static boolean isAndroid() {
-        boolean firstCheck =
-                System.getProperty("java.runtime.name", "").toLowerCase().contains("android");
-        if (!firstCheck) {
-            return checkIfAndroidTermux();
+        return isAndroidRuntime() || isAndroidTermux();
+    }
+
+    public static boolean isAndroidRuntime() {
+        return System.getProperty("java.runtime.name", "").toLowerCase().contains("android");
+    }
+
+    public static boolean isAndroidTermux() {
+        try {
+            return processRunner.runAndWaitFor("uname -o").toLowerCase().contains("android");
+        } catch (Exception ignored) {
+            return false;
         }
-        return firstCheck;
     }
 
     public static boolean isMusl() {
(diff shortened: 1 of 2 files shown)


gotson:
why did you change some of the native libraries ?

gotson:
when running `java` within Termux, is it using a different Java runtime that when running a normal Android app ? What is the result of `System.getProperty("java.runtime.name", "")` within Termux ?

arch-dev:
Here's the output:

src/main/java/org/sqlite/util/OSInfo.java
```diff
@@ -251,4 +268,27 @@ static String translateOSNameToFolderName(String osName) {
...
+            output.append(line + "\n");
+        }
+
+        int exitVal = process.waitFor();
+        if (exitVal == 0) {
+            return output.toString().toLowerCase().contains("android");
+        }
+
+        return false;
+    }
```
gotson, line 293: can probably be replaced by `processRunner.runAndWaitFor("uname -o")`

gotson:
While working on #797 i am actually not sure how Termux works on Android.
- it seems to be a JDK from eclipse, but not specific to Android?
- however the native library needs to have been compiled with the NDK for it to work
- [that bit](https://github.com/xerial/sqlite-jdbc/blob/58858f7b0006cfe3b6f169b20a1b9602d98308d1/src/main/java/org/sqlite/core/NativeDB.java#L39) is apparently not triggered though, and the classic loader is used

Would you be able to check the value of the property `java.vm.vendor` also ?

arch-dev:
Here you are:

gotson:
Thanks, that confirm my understanding.

gotson:
This still requires some work @arch-dev as there are conflicts, and the native libs should not have been changed. Is that something you're still working on?

arch-dev:
Reworked and used available `ProcessRunner` class.

gotson:
@arch-dev i have refactored the code slightly and added some tests, please have a look and let me know if that's fine with you

arch-dev:
Yes good idea separing checks. Tested and it works without problems.

PR #799, opened 2022-10-22
correction of isNullable() in JDBC3ResultSet

koeltv:
changed isNullable() in JDBC3ResultSet to use first column of meta[checkCol(col)] as it returns [nullable, primary, auto_incremented]

added isPrimary() in JDBC3ResultSet to return the value of the second column of meta[checkCol(col)] added corresponding tests in RSMetaDataTest
corrected line 60 in columns test in RSMetaDataTest as a primary key attribute can be null in SQLite

## Example with table "test":
create_test
isNullable_test

## Example with table "test2":
create_test2
isNullable_test2

### DIFF
--- a/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
@@ -827,9 +827,21 @@ public abstract class JDBC3ResultSet extends CoreResultSet {
     /** @see java.sql.ResultSetMetaData#isNullable(int) */
     public int isNullable(int col) throws SQLException {
         checkMeta();
-        return meta[checkCol(col)][1]
+        return meta[checkCol(col)][0]
                 ? ResultSetMetaData.columnNoNulls
                 : ResultSetMetaData.columnNullable;
+    }
+
+    /**
+     * Indicates whether the values in the designated column are primary attributes.
+     *
+     * @param col the first column is 1, the second is 2, ...
+     * @return whether the given column correspond to a primary attribute or not
+     * @throws SQLException if a database access error occurs
+     */
+    public boolean isPrimary(int col) throws SQLException {
+        checkMeta();
+        return meta[checkCol(col)][1];
     }
 
     /** @see java.sql.ResultSetMetaData#isAutoIncrement(int) */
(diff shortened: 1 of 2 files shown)


gotson:
I don't understand what this does, or what was the problem before you tried fixing it, would you care to explain?

koeltv:
The issue was when you tried to know if one column is nullable or not using something like `statement.executeQuery("SELECT * FROM tableName").getMetaData().isNullable(x)` it would not return the nullable state but whether the attribute was primary or not, which correspond to the second column of meta in JDBC3ResultSet.
The actual nullable state is actually contained in the first column of meta (as shown in the 2 examples I put above)

koeltv:
So the main change was to take index 0 instead of 1 here (JDBC3ResultSet) :
`public int isNullable(int col) throws SQLException {
        checkMeta();
        return meta[checkCol(col)][0]
                ? ResultSetMetaData.columnNoNulls
                : ResultSetMetaData.columnNullable;
    }`

The other changes I made were to add the method isPrimary(i) to get the value of the second column, which contained whether the column contained primary keys or not and to update the tests for those methods

gotson:
> The issue was when you tried to know if one column is nullable or not using something like `statement.executeQuery("SELECT * FROM tableName").getMetaData().isNullable(x)` it would not return the nullable state but whether the attribute was primary or not, which correspond to the second column of meta in JDBC3ResultSet. The actual nullable state is actually contained in the first column of meta (as shown in the 2 examples I put above)

Thanks, i do agree with this.



> The other changes I made were to add the method isPrimary(i) to get the value of the second column, which contained whether the column contained primary keys or not and to update the tests for those methods

This is not part of JSBC so i think it doesn't make sense to add this. There is already `DatabaseMetaData#getPrimaryKeys` for that purpose.

---

I think the tests should be changed a bit to reflect what you showed as screenshots above too.

koeltv:
> This is not part of JSBC so i think it doesn't make sense to add this. There is already DatabaseMetaData#getPrimaryKeys for that purpose.

Oh ok I didn't see that method, I removed the `JDBC3ResultSet#isPrimary(i)` that I added previously and the corresponding tests

> I think the tests should be changed a bit to reflect what you showed as screenshots above too.

I expanded the `nullable` test in `RSMetaDataTest` to run a test similar to my previous examples

gotson:
Thanks, looks good.

Can you run `mvn spotless:apply` to fix the code formatting?

koeltv:
Sorry I didn't see that, also it seems like there are some issues when you try to run the command with recents SDK.
I failed with Oracle OpenJDK v.17.0.1 but it worked with Eclipse Temurin v.11.0.17.

gotson:
> Sorry I didn't see that, also it seems like there are some issues when you try to run the command with recents SDK. I failed with Oracle OpenJDK v.17.0.1 but it worked with Eclipse Temurin v.11.0.17.

probably because Maven 🤷🏻‍♂️

Issue #790, opened 2022-09-21
Add module-info.java


onkobu:
**Is your feature request related to a problem? Please describe.**
When packing with jlink sqlite-jdbc breaks the build. It is a so called auto module and cannot be packaged with jlink automatically.

**Describe the solution you'd like**
Add a module-info.java.

**Describe alternatives you've considered**
Issue #478

**Additional context**
While it is feasible as a developer to hack some script or action to re-pack sqlite-jdbc with appropriate module info it is impossible for the lesser experienced. It requires pre-conditions and initial steps that are error prone and not reproducible.

Tool `jdeps` generates this:

```
module org.xerial.sqlitejdbc {
    requires transitive java.logging;
    requires transitive java.sql;

    exports org.sqlite;
    exports org.sqlite.core;
    exports org.sqlite.date;
    exports org.sqlite.javax;
    exports org.sqlite.jdbc3;
    exports org.sqlite.jdbc4;
    exports org.sqlite.util;

    provides java.sql.Driver with
        org.sqlite.JDBC;

}
```

I use this module-info.java in https://sourceforge.net/projects/jworktime/ for a Swing-based project as well as in https://codeberg.org/onkobu/mottow for an Undertow-based REST backend demo. While the first has a section in the README.md the latter ships a shell script to turn a bunch of auto modules into explicit ones.

gotson:
For reference neither [H2](https://github.com/h2database/h2database/issues/1415) nor [pgjdbc](https://github.com/pgjdbc/pgjdbc/issues/1979) have this yet

gotson:
@onkobu seems we could define `Automatic-Module-Name` in the manifest?

gotson:
Oh, actually it's already defined https://github.com/xerial/sqlite-jdbc/blob/7c3f7decb14a30bf55459b3da6a0b1c3171dc161/pom.xml#L128-L128

onkobu:
Checked at least with 3.39.2.1 and jlink from an OpenJDK 17, still treated as automatic module and had to be amended.

> For reference neither https://github.com/h2database/h2database/issues/1415 nor https://github.com/pgjdbc/pgjdbc/issues/1979 have this yet

Steam engines still in use here in Germany never were never updated with Bluetooth. But as soon as I'm packing H2 in favour of SQLite I'll be annoying there, too.

[...]

Issue #767, opened 2022-08-16
Simplify JDBC classes and interfaces


gotson:
While looking at #735 i noticed that we have classes in `org.sqlite.core`, `org.sqlite.jdbc3` and `org.sqlite.jdbc4`.

For example for `Statement`:
- `abstract class CoreStatement`
- `abstract class JDBC3Statement extends CoreStatement`
- `class JDBC4Statement extends JDBC3Statement implements Statement`

And for `PreparedStatement`:
- `abstract class CorePreparedStatement extends JDBC4Statement`
- `abstract class JDBC3PreparedStatement extends CorePreparedStatement`
- `class JDBC4PreparedStatement extends JDBC3PreparedStatement implements PreparedStatement, ParameterMetaData`

That seems like a mess to me:
- `JDBC3Statement` implements some methods from `Statement` while it doesn't implement the interface
- `CorePreparedStatement` in `core` package depends on `jdbc4` package

I'm not sure what's the intended use of the `core` package, as it would always have dependencies on `java.sql` anyway.

There are also some classes within `org.sqlite` like `SQLiteConnection` or `SQLiteDataSource`.

I suppose there's some history behind that:
- JDBC 4 was introduced in version `3.8.2` along with Java 6
- moved to Java 8 in version `3.32.3.3`

While i am no expert in JDBC, it seems JDBC 4.2 shipped with Java 8.

Given we only support Java 8, we could probably remove the JDBC3 package, and have a single JDBC4 package instead.

I'm not sure what to do with the `core` package or the classes that are in `org.sqlite` to be honest.

If anyone wants to chime in, I would be interested to hear your thoughts.

gotson:
@pyckle I would love to get your opinion on this, as you recently did some changes in those packages/classes.

pyckle:
Hey. Things are very busy for me for the next few weeks, but I'll take a look at this soon.

Most likely the best way to understand why the code was written like this is the git history, which will take a while to sift through.

gotson:
> Most likely the best way to understand why the code was written like this is the git history, which will take a while to sift through.

that's 10+ years of history 🙃 

I plan to have a look at other JDBC drivers, namely H2 and PostgreSQL, for inspiration.

gotson:
I noticed that the classes are almost never encapsulated. It's quite often that another class will modify a class, via `public` or `protected` fields.

For example:
- when preparing a statement, the DB class is modifying the Statement object
[...]

PR #804, opened 2022-11-05
Multi-release JAR with module-info.java

hadrabap:
Hello @xerial, @gotson and others!

I tried to implement `module-info.java` (#790). I decided to go the multi-release way.

What I had to do is:
1. Upgrade `maven-compiler-plugin` to support multirelease and the new `<release />` configuration option
2. Upgrade `maven-bundle-plugin` to understand multirelease JARs

Next, I don't know if you are using Maven Toolchains, so I left the version selection commented out. Tell me if I should remove them.

Finally, I also registered other JDBC entrypoints via service loader like `DataSource` to enable automatic configuration in Jakarta EE Containers (Glassfish, OpenLiberty) or others. Once again, it is not necessary. If you don't like it, I'll remove it.

Please tell me your opinion about this.

With best regards,

PETR


jasonli0616:
> Now I see a few builds are failing. It looks like the JDK8 build cancels the rest of the test builds.
> 
> There are three ways fixing it:
> 
>     1. Upgrade the JDK8 build to e.g. JDK11
> 
>     2. Leave the JDK8 build intact and make the multi-release build optional (probably with Maven Profile)
> 
>     3. Use Maven Toolchains so JDK8 will build everything and JDK9+ will do the multi-release stuff.
> 
> 
> What is your opinion?

@hadrabap I believe number 3 is the best option.
Have JDK8 remain to build, and use JDK9+ for the new changes in this PR. Seems like the most straightforward way to do it (although probably easier said than done?).
As I mentioned in #790 though, Java is not my main language and I may be mistaken for some stuff :)

src/main/java9/module-info.java
```diff
@@ -0,0 +1,19 @@
...
+    requires transitive java.sql;
+    requires transitive java.sql.rowset;
+
+    exports org.sqlite;
+    exports org.sqlite.core;
+    exports org.sqlite.date;
+    exports org.sqlite.javax;
+    exports org.sqlite.jdbc3;
+    exports org.sqlite.jdbc4;
+    exports org.sqlite.util;
```
gotson, line 13: I'm not sure about this. Can't we export a single module ? We should also probably use a name that is closer to the maven artifact name if that's possible.

gotson:
> Next, I don't know if you are using Maven Toolchains, so I left the version selection commented out. Tell me if I should remove them.

Honestly i don't know. I personally hate Maven, and would much prefer to use Gradle, with which i have experience with multi-release jars and toolchains. I've never used Maven toolchains.

> Finally, I also registered other JDBC entrypoints via service loader like `DataSource` to enable automatic configuration in Jakarta EE Containers (Glassfish, OpenLiberty) or others. Once again, it is not necessary. If you don't like it, I'll remove it.

Probably best to have this in another PR, and explain the rationale and provide context also.

> * Upgrade the JDK8 build to e.g. JDK11

We already required JDK11 to run Spotless. I think we could make JDK11 mandatory for building, and use the `-release` flag to build with target of 8.

src/main/java9/module-info.java
```diff
@@ -0,0 +1,19 @@
+module org.xerial.sqlitejdbc {
+
+    requires transitive java.logging;
```
gotson, line 3: there is no use of `java.logging` in the project though

src/main/java9/module-info.java
```diff
@@ -0,0 +1,19 @@
...
+    requires transitive java.sql;
+    requires transitive java.sql.rowset;
```
gotson, line 5: same here, there's no mention of this anywhere in the code, why do we need that?

gotson:
I was checking about Maven Toolchains, the problem i see is that the path of each JDK needs to be specified manually. That's a problem because everyone has a different path on their machine, and the plugin cannot find the JDK by using a discovery mechanism. That's a big drawback compared to the Gradle Java toolchains.

Github Actions setup-java [now supports](https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#Modifying-Maven-Toolchains) Maven Toolchains, but that doesn't solve the problem for individual developers.

gotson:
> > I was checking about Maven Toolchains, the problem i see is that the path of each JDK needs to be specified manually. That's a problem because everyone has a different path on their machine, and the plugin cannot find the JDK by using a discovery mechanism. That's a big drawback compared to the Gradle Java toolchains.
> > Github Actions setup-java [now supports](https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#Modifying-Maven-Toolchains) Maven Toolchains, but that doesn't solve the problem for individual developers.
> 
> There are no such problems. The Toolchain plug-in is used just by the developer of the project. Consumers can see the references to it in the POM, but Maven does not take care. When depending on another project Maven reads `<dependencies/>` only, nothing else.

I don't understand, if you installed your JDK in a different folder than my JDK, the `toolchains.xml` still has to reference the absolute path. Then it can't work.

src/main/java9/module-info.java
gotson: I am not familiar with Java modules, but since #767 will most likely change the packages, will there be an impact on the modules for the consumers?

for the name i'm afraid `org.sqlite` is too generic. I know the package is currently set to that, but that's also a problem, as there are more Sqlite drivers out there now. You already set the module name to `org.xerial.sqlitejdbc` which i think is fine. The `exports` depend on the existing package, in that case we can't change them until we rename the packages.

gotson:
> The `toolchain.xml` is used locally by the developer and is not shared

That's what i was missing ! But if someone never set or used toolchains, then it won't work. I think it's not very intuitive, no?

For instance i don't have a `~/.m2/toolchains.xml` on my machine, so checking out the project and trying to run it would fail.

gotson:
> > For instance i don't have a `~/.m2/toolchains.xml` on my machine, so checking out the project and trying to run it would fail.
> 
> Yes, it fails if the configuration file is missing or does not provide JDK requested by the project.

That's what I'm afraid of, we have so many contributors for this project, I don't want to add more impediments by using toolchains

gotson:
> You said that JDK11 is already mandatory due to some testing stuff or alike, don't you?

It's mentioned in the contributing instructions, jdk 11 is required to run the spotless maven plugin.

You could rebase your PR, I have updated all the plugins today!

gotson:
Since I added Maven enforcer today to enforce minimum maven version maybe we can use that to enforce minimum jdk version? https://maven.apache.org/enforcer/enforcer-rules/requireJavaVersion.html

This will also need updating in that case https://github.com/xerial/sqlite-jdbc/blob/master/CONTRIBUTING.md#prerequisites 

And we need to amend the CI job to drop jdk 8 and use jdk 11 instead for all the jobs that were on 8

gotson:
> All done. The only problem is with one test which seems to be crashing from something who-knows-whats-going-on. The tests seem to be fine, but the JVM did something bad…

which one would that be? i see CI is green now

gotson:
> It started at `2022-11-07T19:13:21.9227372Z` and the job in question is
> 
> ```
> 2022-11-07T19:13:34.9620516Z ##[group]Run uraimo/run-on-arch-action@v2 
> 2022-11-07T19:13:34.9620859Z with:                                     
> 2022-11-07T19:13:34.9621065Z   arch: armv7                             
> 2022-11-07T19:13:34.9621308Z   distro: ubuntu_latest
> ```
> 
> It is the previous run as of writing this.
> 
> The log has the following message:
> 
> ```
> 2022-11-07T19:21:10.4765795Z [ERROR] Error occurred in starting fork, check output in log                                                     
> 2022-11-07T19:21:10.4766310Z [ERROR] Process Exit Code: 1              
> 2022-11-07T19:21:10.4781159Z [ERROR] Crashed tests:                    
> 2022-11-07T19:21:10.4795018Z [ERROR] org.sqlite.architecture.TestCodingRulesTest                                                              
> 2022-11-07T19:21:10.4855714Z [ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?                         
> 2022-11-07T19:21:10.4896038Z [ERROR] Command was /bin/sh -c cd /work && /usr/lib/jvm/java-11-openjdk-armhf/bin/java -jar /work/target/surefire/surefirebooter3000460293369316902.jar /work/target/surefire 2022-11-07T19-17-44_960-jvmRun1 surefire3163602568984339717tmp surefire_015744727068334001731tmp
> ```
> 
> I have the complete log downloaded and can provide it to you if necessary.
> 
> Interesting thing is that the second run was successful. Another explanation could be issues on GitHub site. It got extremely slow and I faced lots of out-of-available-servers errors.

If a rerun ends up passing, it should be fine :)

gotson:
> > If a rerun ends up passing, it should be fine :)
> 
> LOL 😆 Sounds deterministic to me 😁

should only care when it's effectively flaky

gotson:
> So we have [one last open question](https://github.com/xerial/sqlite-jdbc/pull/804/files#r1015153784).
> 
> What shall I do next? Any ideas, improvements?

The module name ? I am fine with `org.xerial.sqlitejdbc`

gotson:
@hadrabap is there some rework you need to do, or is this ready for merging?

gotson:
> @gotson From my side it's complete and ready for merge.

just tested this locally for the first time, but the generated jar doesn't contain a java9 folder and doesn't contain the module-info. What am i doing wrong?

PR #818, opened 2022-12-01
Fixes and Improvements to the Backup/Restore SQLite Feature

pyckle:
BugFixes:
backup/restore sql commands run in jdbc fail silently if the backup or restore fails. The bugfix is that we check the return code status in JNI and throw an exception on failure. There is a unit test written to validate this
backup/restore progress observers were never updated. They are now updated if passed in via JNI. This is now unit tested.

Feature:
Various parameters to backup/restore can now be provided via jni rather than being hard coded within C.

JNI libs were rebuilt. I am happy to remove this commit if you'd prefer to build them.


gotson:
Thanks, will have a look when i have some time. I was wondering why we do have that feature in the driver at all to be honest 🤔

pyckle:
> Thanks, will have a look when i have some time. I was wondering why we do have that feature in the driver at all to be honest thinking

Sure, no hurry.

I tend to agree that having regex in the driver that preprocesses a query and runs it through a completely different flow isn't great. It's also not at all portable - backup/restore aren't available in other jdbc drivers. At a minimum, if the backup/restore fails, the driver should throw a SQLException. I had a failing restore in my codebase and it took me a while to understand this is what was happening.

With that said, it makes sense to expose these APIs as it's a useful feature to be able to backup/restore SQLite databases. As it is SQLite specific, I think exposing it as a separate API makes more sense.

Also, while I parameterized some constants, I think that more customization may be necessary for many backup use cases. See https://www.sqlite.org/backup.html

Specifically, see example 2. The way that our backup/restore code is written, concurrent reads to the db would be handled fine, but concurrent writes would fail until the backup finishes. This may not always be desired.

src/main/java/org/sqlite/core/DB.java
```diff
@@ -762,6 +762,28 @@ public abstract int create_function(String name, Function f, int nArgs, int flag
     public abstract int backup(String dbName, String destFileName, ProgressObserver observer)
```
gotson, line 762: do we need to keep that method if we have the new one with more parameters?

pyckle: I left this here in case anybody integrated with it directly, or if they want to do a backup/restore without figuring out what parameters to add. I'm happy to inline/remove if you think it's cleaner.

gotson: thanks for the clarification, you are right that it would change the public contract of the API. Let's leave it as is.

gotson:
> JNI libs were rebuilt. I am happy to remove this commit if you'd prefer to build them.

It seems i am having issues with my workflow to rebuild native libs in PRs. Could you drop that commit ? I will rebuild the native libs after merging this PR to `master`. Thanks.

gotson:
> > JNI libs were rebuilt. I am happy to remove this commit if you'd prefer to build them.
> 
> ~It seems i am having issues with my workflow to rebuild native libs in PRs. Could you drop that commit ? I will rebuild the native libs after merging this PR to `master`. Thanks.~

Don't mind me. I managed to fix the Build Native workflow, all good now. Thanks for the PR!

PR #823, opened 2022-12-27
Refine GraalVM native-image configuration and fix native test execution

kkriske:
* Refine the native-image configuration and make inclusion of the metadata conditional
* Update the jni code to load and cache all java class/field/method references in JNI_OnLoad. This makes it easier to validate native-image configuration as successfully loading the library means all required JNI configuration is present.
* fix native-image test execution and include as much tests as currently possible
  * Tests using AssertJ `assume*` calls are excluded because AssertJ relies on dynamic class loading which I couldn't get to play nice.

The commits currently are a bit of a mess, I can clean it up a bit before review if preferred.

Future work (probably separate PR):
I want to move the JNI configuration to a `Feature` implementation, as I once upon a time suggested here: https://github.com/xerial/sqlite-jdbc/pull/631#issuecomment-997838042. Such a feature is imo easier to maintain because it's just typechecked java code and not a json file. Additionally it would allow dynamic configuration of which resources get included in the native-image, so only the relevant library for the target OS is included instead of all of them reducing the overall size of the resulting executable. It would also enable unpacking the required library during the build, placing it next to the target exeuctable so it doesn't need to be unpacked to a temp dir at all. This would benefit the startup time of the image.

/cc @gotson Taking the liberty to tag you. The git history and [this branch of yours](https://github.com/gotson/sqlite-jdbc/tree/fix-native-test) tell me you've worked on this part of the code before.


gotson:
> /cc @gotson Taking the liberty to tag you. The git history and [this branch of yours](https://github.com/gotson/sqlite-jdbc/tree/fix-native-test) tell me you've worked on this part of the code before.

I'm also the maintainer of the repo :) I will need some time to review this, might ask some questions when i do.

src/main/java/org/sqlite/core/NativeDB.c
```diff
@@ -474,11 +490,23 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
...
 
-    jclass exclass = (*env)->FindClass(env, "java/lang/Throwable");
+    bhandleclass = (*env)->FindClass(env, "org/sqlite/BusyHandler");
+    if(!bhandleclass) return JNI_ERR;
+    bhandleclass = (*env)->NewWeakGlobalRef(env, bhandleclass);
+    bhandle_mth_callback = (*env)->GetMethodID(env, bhandleclass, "callback", "(I)I");
+
+    exclass = (*env)->FindClass(env, "java/lang/Throwable");
+    if(!exclass) return JNI_ERR;
+    exclass = (*env)->NewWeakGlobalRef(env, exclass);
```
kkriske, line 502: Lack of global reference to this class caused the Segfault mentioned in the surefire exclusions. The class reference got garbage collected which invalidates the jmethodID which caused usage of freed memory.

src/main/java/org/sqlite/core/NativeDB.c
```diff
@@ -474,11 +490,23 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
...
     exp_msg = (*env)->GetMethodID(
             env, exclass, "toString", "()Ljava/lang/String;");
 
+    bool_array_class = (*env)->FindClass(env, "[Z");
+    if(!bool_array_class) return JNI_ERR;
+    bool_array_class = (*env)->NewWeakGlobalRef(env, bool_array_class);
```
kkriske, line 508: The jni-config entry for `boolean[]` was missing, which caused the `java.sql.SQLException: Out of memory` mentioned in the surefire exclusions.

src/main/java/org/sqlite/core/NativeDB.c
```diff
@@ -616,10 +650,6 @@ void change_busy_handler(JNIEnv *env, jobject nativeDB, jobject busyHandler)
         (*env)->GetJavaVM(env, &busyHandlerContext->vm);
 
         busyHandlerContext->obj = (*env)->NewGlobalRef(env, busyHandler);
-        busyHandlerContext->methodId = (*env)->GetMethodID(  env,
-                                                   (*env)->GetObjectClass(env, busyHandlerContext->obj),
-                                                   "callback",
-                                                   "(I)I");
```
kkriske, line 622: This did a lookup to `BusyHandler#callback` for every implementation of `BusyHandler` that was passed to this function. It is sufficient to store the reference to the interface method, which will allow calling it on every subclass as well.

The lookup on the subclasses also ment that a jni-config entry would have been needed for every used implementation class which pushes the configuration requirement to the end-user. By storing the interface method id there is no more configuration needed by the user.

src/main/java/org/sqlite/core/NativeDB.c
```diff
@@ -1666,10 +1694,6 @@ static void change_progress_handler(JNIEnv *env, jobject nativeDB, jobject progr
         (*env)->GetJavaVM(env, &progressHandlerContext->vm);
 
         progressHandlerContext->phandler = (*env)->NewGlobalRef(env, progressHandler);
-        progressHandlerContext->mth = (*env)->GetMethodID(  env,
-                                                   (*env)->GetObjectClass(env, progressHandlerContext->phandler),
-                                                   "progress",
-                                                   "()I");
```
kkriske, line 1672: Same thing as `BusyHandler#callback`.

src/test/resources/META-INF/native-image/org.xerial/sqlite-jdbc/reflect-config.json
```diff
@@ -0,0 +1,15 @@
+[
```
kkriske, line 1: This config is only added to the test-suite, so it is not provided to the end-user.
It is only required when using rowsets, which is not code maintained by sqlite-jdbc so it's not up to us to provide the user with this config. It would also cause a direct dependency to java internals which may not be compatible over all jdks/versions.

Same thing for the `resource-config.json` next to this file.

kkriske (commented):
I don't know how up-to-date you are with how GraalVM native-image works. It works with a closed world assumption, because to convert everything to native-code you need to know everything that will be executed. Everything that's not used will be removed but some usages are pretty much impossible to detect using bytecode analysis, for example reflection and JNI calls. So that's why the jni-config.json file is required, to tell native-image which JNI calls will be required so the needed metadata for calling it is available, and the method is not entirely removed either because it may not be detected as being used.

As a library however, we can't assume that just because we're on the classpath means we're used. So the config metadata should always be conditional. If we don't do that, we will always force these jni calls and the reachable code through them to be included in the native-image. So by adding this condition:
```json
"condition": {
    "typeReachable": "org.sqlite.core.NativeDB"
}
```
The JNI calls will only be included if the bytecode analysis detects that the `org.sqlite.core.NativeDB` class is actually used.
So if sqlite-jdbc is on the classpath, but never used/called, we also don't include any code.

For the resource config, there is no such thing as conditional inclusion yet, which is why I want to do a follow-up to create a `Feature` implementation, because it is possible to do it through the java native-image API. It's then also possible to only include the resources for the relevant target OS instead of everything.

---

> Update the jni code to load and cache all java class/field/method references in JNI_OnLoad. This makes it easier to validate native-image configuration as successfully loading the library means all required JNI configuration is present.

I moved all `FindClass`, `GetFieldID` and `GetMethodID` calls to the `JNI_OnLoad` function. This makes maintenance easier in my opinion as it groups all native->java lookups together. Every call of one of these functions results in an entry in the jni-config.json file, so grouping them also makes it easier to keep the native lookups and the required native-image configuration in sync.
This is partially what I ment with `This makes it easier to validate native-image configuration`. A second part of this is that if you're running in native-image and the library loads successfully (meaning the JNI_OnLoad does not result in an exception) then all native->java lookups already happened and you're certain no additional calls will happen at runtime for which you may have overlooked the configuration (this was the case for `FindClass(env, "[Z")` for example).

What this boils down to, is that if the native test execution in the GitHub Actions succeeds, we can be certain that the configuration is complete.

---

> What are the changes in the native code and why are they required?

For the remaining native code changes, I have added some comments in the code. Let me know if I glossed over some other stuff that needs explaining.

gotson:
Thanks @kkriske for the details, that is very useful.

gotson:
Hi @pyckle , if you have a bit of spare time, would you mind reviewing the changes done in the native C code ? 🙏🏻

pom.xml
```diff
@@ -345,22 +346,18 @@
...
-                                <buildArg>--no-fallback</buildArg>
-                                <buildArg>--verbose</buildArg>
-                                <buildArg>--enable-url-protocols=jar</buildArg>
-                                <buildArg>--report-unsupported-elements-at-runtime</buildArg>
+                                <!--
+                                ArchUnit tests don't run in native-image tests.
+                                Remove the ArchUnit JUnit Engine from the ServiceLoader.
+                                -->
+                                <buildArg>
+                                    -H:ServiceLoaderFeatureExcludeServiceProviders=com.tngtech.archunit.junit.internal.ArchUnitTestEngine
```
gotson, line 359: Do we need this, since we already have `<exclude>**/architecture/*.java</exclude>` below in the Surefire configuration?

pom.xml
```diff
@@ -345,22 +346,18 @@
...
-                                <goals>
-                                    <goal>build</goal>
-                                </goals>
-                                <phase>package</phase>
-                            </execution>
                         </executions>
                         <configuration>
-                            <imageName>sqlite-jdbc-osinfo</imageName>
-                            <mainClass>org.sqlite.util.OSInfo</mainClass>
+                            <fallback>false</fallback>
```
gotson, line 351: what does this do exactly?

pom.xml
```diff
@@ -370,33 +367,17 @@
                         <artifactId>maven-surefire-plugin</artifactId>
                         <version>${surefire.version}</version>
                         <configuration>
-                            <!--  If enabled, will run the GraalVM tracing agent while running tests  -->
-                            <!--              <argLine>-agentlib:native-image-agent=config-output-dir=META-INF/native-image</argLine>-->
```
gotson, line 374: don't we need that anymore?

pom.xml
kkriske: A fallback image is basically a wrapper around the jars that executes the jars. It still requires a JDK to run and it isn't a truly native executable. It gets created when something in the configuration is insufficient and the native-image generation would otherwise crash.
Its sole purpose is to still create something that works for any java program you throw at it, which is as I understand is purely a legal requirement to call native-image a valid java implementation. For the same reason it is enabled by default but when you're after a truly native executable it's something you never want. So sadly it's something that always has to be disabled manually. A bit more info here, but as evident by all the legal disclaimers it's not easy to get a straight answer on the matter: https://github.com/oracle/graal/issues/2648. I am also not a lawyer so use that info with this in mind.

pom.xml
kkriske: You can run the agent by adding some configuration to the `native-maven-plugin`: https://graalvm.github.io/native-build-tools/latest/maven-plugin.html#agent-support-enabling
However just running the agent creates quite a bit config that's not required at all, so it's not that great. Plus the only config that you need are the JNI calls in `JNI_OnLoad` and some resources. That should be straightforward enough without the agent.

pom.xml
kkriske: Yes, native-image auto-detects service loaders, so the `ArchUnitTestEngine` is included. At some point during native-image generation it gets instantiated which causes runtime-buildtime initialization issues. Excluding the service entirely solves the same issue you solved with this config: https://github.com/gotson/sqlite-jdbc/blob/fix-native-test/pom.xml#L318

gotson: I'm fine with your change, we don't need to run ArchUnit tests on native anyway.

pyckle:
I'm happy to review, but I won't be able to until next week.

gotson:
> I'm happy to review, but I won't be able to until next week.

no rush, everyone's working for free here, and your help and expertise is much appreciated !

kkriske:
FYI, I am not a C or JNI expert, I'm just working with whatever knowledge I've picked up along the way.

pettyalex:
Awesome, I wish I would have checked this before debugging the failure to run in a native image myself.

I did notice one difference between your jni-config.json and mine: my tracing agent also found

```
  {
    "name":"java.lang.Boolean",
    "methods":[{"name":"getBoolean","parameterTypes":["java.lang.String"] }]
  },
```
in addition to Throwable. Is that necessary with the JNI access to boolean arrays, or is my traced usage probably coming from somewhere in Spring Boot?

kkriske:
Looks like I'll have to do a small update to accomodate the changes of https://github.com/xerial/sqlite-jdbc/pull/818

src/main/java/org/sqlite/core/NativeDB.c
pyckle: It seems bizarre to me that java.lang.Throwable can be unloaded when we instantiate an Exception where Throwable is an interface.

Also, since this is a call to `toString()`, I think that we can use java.lang.Object, rather than Throwable.

Also, as we're holding a weak global reference, I'm confused at what would be stopping the VM from unloading the class fully, as a weak reference shouldn't prevent this.

With this said, I am not super familiar with these APIs.

src/main/java/org/sqlite/core/NativeDB.c
pyckle: Do we need to maintain a global weak reference to this, can we just make sure that the returned class is not null which implies it is accessible from JNI?

src/main/java/org/sqlite/core/NativeDB.c
pyckle: Nice fix (:+1:

pyckle:
@kkriske - very nice work (:

I must apologize as it appears that much of the breakage is rooted in my changes that fixed race conditions, memory leaks, and other code cleanup.

Just one last question - regarding

```
"condition": {
    "typeReachable": "org.sqlite.core.NativeDB"
}
```

If SQLite is initialized for JDBC via reflection - `Class.forName("org.sqlite.JDBC");` then obtained via `DriverManager.getConnection("jdbc:sqlite:");` does this config realize that NativeDB is reachable?

If not, perhaps we want to remove this condition? I suspect that this may be a common way of loading the sqlite driver.

src/main/java/org/sqlite/core/NativeDB.c
kkriske: Do you mean not storing the class ref and doing the `FindClass` call again later when needed? While that would work, I think it would be a bit pointless. The lookup itself is "slow" so it makes sense to cache it if the lookup happens anyway in the `JNI_OnLoad`. It would also break with my suggestion to keep JNI lookups confined in the `JNI_OnLoad` function.

kkriske:
> If SQLite is initialized for JDBC via reflection - `Class.forName("org.sqlite.JDBC");` then obtained via `DriverManager.getConnection("jdbc:sqlite:");` does this config realize that NativeDB is reachable?

This instinctively requires additional reflection configuration to allow native-image to reflectively resolve the `Class.forName` call. However native-image is smart enough to recognise this call with constant string argument, and inlines it with the `org.sqlite.JDBC` class directly.

If you would pass it as an argument so it can't resolve it automatically, e.g.:
```java
public static void main(String[] args) {
    Class.forName(args[0]);
}
```
Then you'd need to manually add the reflective call to make it work, but that's up to the end-user and not our responsibility as it's about how they use the library and not the internal workings.

----
All that being clarified, it basically boils down to that `org.sqlite.JDBC` is seen as reachable in native-image for both reflective or nomal usages and the static block will still run `DriverManager.registerDriver(new JDBC());` so the driver is actually available.
Native-image also knows it is an available subclass of `java.sql.Driver`. So all interface methods that are called through `DriverManager.getConnection("jdbc:sqlite:");` will also be made available on the `org.sqlite.JDBC` class because whenever a `java.sql.Driver` is used, it _could_ be an instance of `org.sqlite.JDBC`.
So in the end `NativeDB` is available through following stacktrace:
```java
Exception in thread "main" java.lang.ExceptionInInitializerError
	at org.sqlite.SQLiteConnection.open(SQLiteConnection.java:278)
	at org.sqlite.SQLiteConnection.<init>(SQLiteConnection.java:65)
	at org.sqlite.JDBC.createConnection(JDBC.java:104)
	at org.sqlite.JDBC.connect(JDBC.java:77) // available because used methods on `java.sql.Driver` mark this as used and include it in native-image
	at java.sql/java.sql.DriverManager.getConnection(DriverManager.java:677)
	at java.sql/java.sql.DriverManager.getConnection(DriverManager.java:251)
	at Scratch.main(scratch.java:8)
Caused by: java.lang.RuntimeException: Hello from NativeDB#<clinit>
	at org.sqlite.core.NativeDB.<clinit>(NativeDB.java:43)
	... 9 more
```
Achieved by simply throwing an exception in `NativeDB#<clinit>`, and running this scratch file:
```java
class Scratch {
    public static void main(String[] args) throws ClassNotFoundException, SQLException {
        Class.forName("org.sqlite.JDBC");
        Connection conn = DriverManager.getConnection("jdbc:sqlite:test.db");
    }
}
```

kkriske:
I tried adding GraalVM jdk 17 to the test suite, and on linux it fails with this exception:
```
Failed to load native library:sqlite-3.40.1-9ed430df-e5b7-48f1-96e2-b5be478a4841-libsqlitejdbc.so. osinfo: Linux/x86_64
java.lang.UnsatisfiedLinkError: Can't load library: /tmp/sqlite-3.40.1-9ed430df-e5b7-48f1-96e2-b5be478a4841-libsqlitejdbc.so
```
https://github.com/kkriske/sqlite-jdbc/actions/runs/3974368989/jobs/6813598379
I honestly don't know why it fails only for linux.. It runs fine on Windows (tested locally).

src/main/java/org/sqlite/core/NativeDB.c
pyckle: Looking into this more, I think what we ran into is this: https://stackoverflow.com/questions/11907832/jni-local-vs-global-reference-is-not-a-valid-jni-reference

It appears we're just lucky that the regular JVM isn't moving these classes around.

It seems that the proper fix is that these should be all global references (these classes should not be unloaded), or FindClass should be called in the places where these classes are used. Calling FindClass more often is likely cleaner, but allows for more runtime exceptions due to graal native image misconfiguration and may be slower.

pyckle:
@kkriske - Thank you for the through explanation of GraalVM native image. This was informative and makes a lot of sense. I didn't realize it was intelligent enough to analyze static strings to Class.forName().

I also reproduced the linux GraalVM 17 failure locally. The root cause appears to be:
```
Failed to load native library:sqlite-3.40.1-9ed430df-e5b7-48f1-96e2-b5be478a4841-libsqlitejdbc.so. osinfo: Linux/x86_64
java.lang.UnsatisfiedLinkError: Can't load library: /tmp/sqlite-3.40.1-9ed430df-e5b7-48f1-96e2-b5be478a4841-libsqlitejdbc.so
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.NativeLibrarySupport.loadLibraryAbsolute(NativeLibrarySupport.java:92)
	at java.base@17.0.5/java.lang.ClassLoader.loadLibrary(ClassLoader.java:57)
```

The library file is indeed extracted to /tmp, and permissions appear correct (at least on my debian laptop).

Looking a bit at the graal source code:
https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibrarySupport.java#L92

If I am reading this code correctly, it is possible for this exception to be thrown because the native jni OnLoad returns JNI_ERR:
https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibrarySupport.java#L165

The next step is to see if we reach the JNI OnLoad function. I will try to debug this a bit in the next few days.

pyckle:
Update: after debugging this with gdb, this seems to be an issue with symbols from the math library not being resolvable:

```
Thread 1 "native-tests" hit Breakpoint 1, __dlopen (file=0x7ffff46218b0 "/tmp/sqlite-3.40.1-36aa1488-5815-428b-ae43-0ede258685eb-libsqlitejdbc.so", mode=1) at dlopen.c:75
75	dlopen.c: No such file or directory.
(gdb) break PosixNativeLibraryFeature.java:188
Breakpoint 2 at 0x5555556e3c34: file com/oracle/svm/core/posix/PosixNativeLibraryFeature.java, line 188.
(gdb) continue
Continuing.

Thread 1 "native-tests" hit Breakpoint 2, com.oracle.svm.core.posix.PosixNativeLibrarySupport$PosixNativeLibrary::doLoad (this=0x7ffff46217f8) at com/oracle/svm/core/posix/PosixNativeLibraryFeature.java:188
188	            return dlhandle.isNonNull();
(gdb) print dlerror()
$1 = 0x55555835aea0 "/tmp/sqlite-3.40.1-36aa1488-5815-428b-ae43-0ede258685eb-libsqlitejdbc.so: undefined symbol: trunc"
```

pyckle:
I did a bit more digging. I think adding -lm to the linker options is the correct fix, and should be safe. Please note, that I don't have significant experience with C/C++ linking, and having a more skilled set of eyes review would be appreciated.

The reason why `libsqlitejdbc.so` does not load in memory is because neither the `native-tests` binary, nor `libsqlitejdbc.so` link against libm.so

```
$ ldd sqlite-jdbc/src/main/resources/org/sqlite/native/Linux/x86_64/libsqlitejdbc.so
	linux-vdso.so.1 (0x00007fff7526f000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff1be36d000)
	/lib64/ld-linux-x86-64.so.2 (0x00007ff1be869000)
$ ldd sqlite-jdbc/target/native-tests
	linux-vdso.so.1 (0x00007ffe94f9a000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fbb6217f000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fbb6215d000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fbb62157000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fbb61f82000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fbb655b1000)
```

However, libjvm.so does:
```
$ ldd graalvm-ce-java17-22.3.0/lib/server/libjvm.so 
	linux-vdso.so.1 (0x00007ffeaa1e3000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fe8d3bd4000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fe8d3bb2000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fe8d3ba8000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fe8d3a64000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fe8d388f000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fe8d4e65000)
```

So when libsqlite.so is loaded in the JVM, it can link against these math symbols.

As such, it makes sense to add -lm to the linker that builds the so, so this inclusion is explicit. After adding, it looks like:
```
$ ldd sqlite-jdbc/libsqlitejdbc.so 
	linux-vdso.so.1 (0x00007fffe636a000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fc23fcf9000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fc23fb24000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fc23ff61000)
```

Also, note that `libsqlitejdbc.so` also uses pthreads functions, but since both native-tests and libjvm.so use pthreads, they're always available.

pyckle:
@kkriske - can you add https://github.com/pyckle/sqlite-jdbc/commit/cad5235c4351c1b05ccf05dacea9e8ca4bb4114a into your branch and then we can see if graal 17 native image works?

kkriske:
@pyckle It looks like macos has the same issue: https://github.com/kkriske/sqlite-jdbc/actions/runs/4009402889/jobs/6884713811
The linux CI didn't run because of the macos failure, so I can't say yet if that did fix the issue.

pyckle:
The MacOS failure is a different issue. Take a closer look at the logs for MacOS:
```
 Failed to load native library:sqlite-3.40.1-81f450e8-b5e3-4dda-93e5-46e374b3a34b-libsqlitejdbc.so. osinfo: Linux/x86_64
java.lang.UnsatisfiedLinkError: Can't load library: /tmp/sqlite-3.40.1-81f450e8-b5e3-4dda-93e5-46e374b3a34b-libsqlitejdbc.so
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.NativeLibrarySupport.loadLibraryAbsolute(NativeLibrarySupport.java:92)
```

Notice `osinfo: Linux/x86_64` and `libsqlitejdbc.so`. It thinks it is running on Linux for some reason, not Mac. The mac jni lib is `libsqlitejdbc.jnilib`

Also, the job setup identifies the os as:
```
Operating System
  Ubuntu
  22.04.1
  LTS
```

In any case, I tested graal native-image on my mac (apple silicon) with both the experimental aarch64 native-image, and amd64 native-image (through rosetta), and both passed all of the unit tests.

There are two options here: fix the macos tests, or disable them. I don't see why fixing this is a prerequisite to merging these changes, this decision is up to @gotson .

Regarding why the graal 11 native image passed without the `-lm` option to gcc. It seems there was a change in graalvm native-image between 11 and 17 which eliminates the dynamic link of a native-image binary to libm.so. Compiling the test binary with graal-ce-11 yields the following dynamic links:
```
$ ldd target/native-tests
	linux-vdso.so.1 (0x00007ffec0bde000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f52b5ecb000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f52b5eae000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f52b5e8c000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f52b5e86000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f52b5cb1000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f52b5b6d000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f52b8d23000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f52b5b51000)
```
Because the binary dynamically links against `libm.so`, `libsqlitejdbc.so` is able to link to the math functions it needs when it is dynamically loaded (`dlopen()`). However, the binary compiled with graal-vm 17, links dynamically links against much less:
```
$ ldd sqlite-jdbc/target/native-tests
	linux-vdso.so.1 (0x00007ffe94f9a000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fbb6217f000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fbb6215d000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fbb62157000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fbb61f82000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fbb655b1000)
```

Less dynamically linked libraries makes the graal-vm 17 executable significantly more portable as it is no longer dependent upon a system having libstdc++, etc. This comes with the consequence that an improperly linked `libsqlitejdbc.so` no longer works as it can no longer piggy back on the native-image's  `libm.so` dependency.

Disclaimer: I'm primarily a Java developer by day, so I'm not 100% sure I understand/explained this correctly.

pyckle:
@kkriske - one more thing I forgot to mention - you need to rebuild the linux JNI binaries in order for my commit to take effect, as adding `-lm` happens at the time when the shared object is linked. My commit changed the Makefile, but I didn't rebuild.

kkriske:
Right, I should do the rebuild still indeed.
I guess between java 11 and 17 there are some changes in the JDK which allow native-image to detect that the native-image we are building does not require awt. I have yet to check the difference in image size of the test executable but fairly sure the one on java 17 should be smaller because it does not have awt statically linked in. It's the only place where dynamic dependencies on `libstdc++` and `libm` are added: https://github.com/oracle/graal/blob/vm-22.3.0/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JNIRegistrationAwt.java#L126-L127

For the macos builds, very odd because once again java 11 does work: https://github.com/kkriske/sqlite-jdbc/actions/runs/3851295662/jobs/6562317916
And here the metadata of the runner is also Ubuntu so I guess there's something going on with the actions runner itself. I'll disable the macos checks for now.

kkriske:
So about the "github actions" build, turns out I messed up the config. Who would've thought that I was at fault and not GitHub 🙄

gotson:
> @kkriske - can you add [pyckle@cad5235](https://github.com/pyckle/sqlite-jdbc/commit/cad5235c4351c1b05ccf05dacea9e8ca4bb4114a) into your branch and then we can see if graal 17 native image works?

Hi @pyckle , i'm not sure that's right for the second change, which is for Android.

gotson:
`Linux-aarch64` is not defined in the `known_targets` in `Makefile.common`, so it would default to using the `Default` values. On top of that we should probably add the `-lm` to other Linux architectures like `Linux-x86`, `Linux-arm`, `Linux-armv6` and `Linux-armv7`.

pyckle:
@gotson  - nice catch. I think you're correct on all of your points.

Adding `-lm` to all of the linux targets would be proper. And it's certainly necessary on the Linux aarch64 target as graal native-image will likely be used on aarch64. The other targets are less important, as graalvm native-image can't make 32 bits arm/x86 images (nor is it likely they will).

@kkriske - can you make these changes? Would you prefer if I make these changes?

As a side note, it seems that it would be more proper to explicitly link with `pthread` too, as pthreads is also used:
```
$ objdump -T  ./src/main/resources/org/sqlite/native/Linux/x86_64/libsqlitejdbc.so  | grep pthread
0000000000000000      D  *UND*	0000000000000000              pthread_mutex_trylock
0000000000000000      D  *UND*	0000000000000000              pthread_mutexattr_settype
0000000000000000      DF *UND*	0000000000000015  GLIBC_2.2.5 pthread_mutex_init
0000000000000000      D  *UND*	0000000000000000              pthread_mutexattr_destroy
0000000000000000      D  *UND*	0000000000000000              pthread_create
0000000000000000      D  *UND*	0000000000000000              pthread_mutexattr_init
0000000000000000      DF *UND*	0000000000000015  GLIBC_2.2.5 pthread_mutex_lock
0000000000000000      DF *UND*	0000000000000015  GLIBC_2.2.5 pthread_mutex_destroy
0000000000000000      DF *UND*	0000000000000015  GLIBC_2.2.5 pthread_mutex_unlock
0000000000000000      D  *UND*	0000000000000000              pthread_join
```

kkriske:
@pyckle I have little experience with Makefiles so I would definitely be more confortable if you're willing to do the changes. Besides it would simplify things over you knowing what I should add and me fumbling around util I get it right.

kkriske:
In the meantime, I was taking a look at adding a [Feature](https://github.com/oracle/graal/blob/vm-22.3.0/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/Feature.java) in the code instead of the config files. However I'm having some dificulties making it work with the multirelease jar, and adding it would require a dependency on the `org.graalvm.sdk` module in the moduleinfo but that would probably force compilation to happen on a GraalVM JDK. I did reach out to some folks at GraalVM to find out what the best practices are for libraries to add support through Feature implementations.

gotson:
> As a side note, it seems that it would be more proper to explicitly link with `pthread` too, as pthreads is also used:

we can add that too. If you want to prepare a separate PR with those changes i will review it.

pyckle:
@kkriske @gotson take a look at https://github.com/pyckle/sqlite-jdbc/commit/0681710cf865c7a61234779189e75ad80d87607c
I added `-pthread -lm` by default as it should be in all linux builds.

@kkriske -Regarding the `Feature` implementation: I think there are two options.
1. Add the library as a dependency with maven scope `provided`. That way this dependency will not be implicitly added to anybody depending upon the sqlite-jdbc library, as provided dependencies are not transitive. The Feature class is present though, but won't cause issues unless somebody explicitly tries to access it.
2. Make the sqlite project a multi module maven project, and put these classes in a separate module.

I think @gotson  should decide which option is preferable, or suggest another option.

With that said, I think these changes should be merged before finishing the better implementation. People are using the graalvm native images with sqlite-jdbc, and this restores what previously worked.

kkriske:
@pyckle 

> Add the library as a dependency with maven scope `provided`.

How I understand it, a provided scope means that the JDK will provide the dependency. So that should work as long as you are compiling on a GraalVM JDK. What I was thinking was a `optional` modifier, which would also not make it transitive, and would allow compiling on any JDK.

You're right though that a possible Feature implementation is definitely a followup after this is done, so maybe we should move that discussion to a separate issue.

Just FYI, this is where I reached out to clarify library development best practices: https://github.com/oracle/graal/discussions/5811
And this is a different issue I ran into with multimodule jars & native tests: https://github.com/graalvm/native-build-tools/issues/112. However this may be possible to work around using a separate testing module.

pyckle:
Having read through the maven documentation, as well as the linked graalvm thread, I think that `provided` makes more sense than `optional`, but in practice it's a semantic difference.

> How I understand it, a provided scope means that the JDK will provide the dependency. So that should work as long as you are compiling on a GraalVM JDK.

Quoting from the maven documentation https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html:
> **provided**
> This is much like compile, but indicates you expect the JDK or a container to provide the dependency at runtime. For example, when building a web application for the Java Enterprise Edition, you would set the dependency on the Servlet API and related Java EE APIs to scope provided because the web container provides those classes. A dependency with this scope is added to the classpath used for compilation and test, but not the runtime classpath. It is not transitive.

Therefore, a regular JDK would compile the classes which use graalvm sdk just fine. They would even load during the maven testing phase. 

> What I was thinking was a `optional` modifier, which would also not make it transitive, and would allow compiling on any JDK.

This implies that there's a case where this library should be included on the runtime classpath. But this is not the case. Either an application is used with graalvm which provides this library, or using a regular JDK and these classes should never be loaded. In neither case is there a reason to include `graal-sdk` on the classpath.

gotson:
> @kkriske @gotson take a look at [pyckle@0681710](https://github.com/pyckle/sqlite-jdbc/commit/0681710cf865c7a61234779189e75ad80d87607c)
> I added `-pthread -lm` by default as it should be in all linux builds.

i left a comment on that commit, the default could apply to any platform that is not a known target, and it could be a non-linux OS. I don't know if that could have an impact.

pyckle:
> i left a comment on that commit, the default could apply to any platform that is not a known target, and it could be a non-linux OS. I don't know if that could have an impact.

Thanks for the feedback (:

I believe that it is only used for Linux based OS. See my response.

gotson:
> and rebuild the jni libs.

this can be done via Github Actions, see https://github.com/xerial/sqlite-jdbc/blob/master/CONTRIBUTING.md#build-from-ci

> Also, please feel free to squash my changes into a single commit without any need to include my name/credit - they're minor changes. And once again, very nice work with these improvements (:

the PR will be squashed anyhow :)

@pyckle are we all good with the code changes in the C JNI code ? I think so but just wanted to make sure!

pyckle:
> @pyckle are we all good with the code changes in the C JNI code ? I think so but just wanted to make sure!

Looks good to me. There's a few issues in the existing JNI code that this PR exposed to me, but those should be addressed in a different PR.

kkriske:
> @kkriske - Please cherry-pick https://github.com/pyckle/sqlite-jdbc/commit/0681710cf865c7a61234779189e75ad80d87607c and https://github.com/pyckle/sqlite-jdbc/commit/031073138971402da71126cffdb63ce4308a0a16
and rebuild the jni libs.

@pyckle I'll try to find some time for that this evening.
Was there anything that still needed to be done with `NewWeakGlobalRef` vs `NewGlobalRef`?

pyckle:
> Was there anything that still needed to be done with `NewWeakGlobalRef` vs `NewGlobalRef`?

I think this change should be in a separate PR. It's been this way for a long time, and those classes are referenced strongly elsewhere so I think this is more of a theoretical bug.

kkriske:
> I think this change should be in a separate PR. It's been this way for a long time, and those classes are referenced strongly elsewhere so I think this is more of a theoretical bug.

I don't think that's accurate.
It's not about the class being garbage collected, it's about the native reference to the class being released which could result in the garbage collection of the class reference. But it's not one and the same.
So in a memory constrained situation, the weak reference could be invalidated in order to free up that little bit of memory, while still keeping the other strong references. Either way, they should be the last thing to be released before going OOM if I understand it correctly, so usually it won't be enough to get out of the memory constrained situation anyway and the risk of it being an issue is extremely small.

Either way, I have cherry-picked the makefile changes and rebuilt the native libraries.

pyckle:
Looking into this a bit more, I actually think WeakReferences are correct. In fact, there's a comment that documents this in `NativeDB.c`:

```
// These classes are weak references to that if the classloader is no longer referenced (garbage)
// It can be garbage collected. The weak references are freed on unload.
```

If these were strongly referenced, we'd have a circular reference that would prevent this classloader from ever be freed resulting in a memory leak. Loaded classes are always strongly referenced until a classloader is closed.

I think that this can be merged. Nice work.

gotson:
@kkriske thanks a lot for the work on this PR ! It's merged and available in the snapshot repo.

@pyckle thanks for your help on that one!

kkriske:
I created a first draft for the feature implementation, you can have a look already and see if it's something you want to persue.

PR #824, opened 2023-01-03
Support for LoongArch64

Panxuefeng-loongson:
LoongArch is a risc architecture, it is developed by Loongson. This PR is used to support loongarch native build.


gotson:
Hi, please provide a PR description.

gotson:
I don't understand what this PR brings to the table:
- you reference a new target `linux-loongarch64` but it is not defined
- the compilation options in `Makefile.common` are the same as the default Linux ones

If this architecture is not buildable using docker, then there's no point adding any configuration. You can just call `make native` on the machine with that processor to build the library, then use a specific native library as explained [here](https://github.com/xerial/sqlite-jdbc/blob/master/USAGE.md#how-to-use-a-specific-native-library).

Panxuefeng-loongson:
I am not very familiar with this repo. According to you, this pr is really unnecessary and I will close this pr．Thanks

PR #826, opened 2023-01-05
fix: graalVM native image with runtime class initialization

pettyalex:
sqlite-jdbc 3.40.0.0 does not correctly run in a GraalVM native image unless full build-time class initialization is done with `--initialize-at-build-time`. Full build time initialization is really hard, maybe impossible, for big reflection-heavy frameworks like Spring Boot, where I ran into this.

I did the recommended approach of using the GraalVM Tracing agent to run my .jar, and then seeing what metadata it generated. It reported JNI usage of Throwable.toString, which this adds. This got my Spring Boot 3.0.1 application working with sqlite-jdbc!

I feel that the most correct way to solve GraalVM native image hints would be to have automated tests that exercise all code-paths of sqlite-jdbc with the tracing agent, and then use that captured metadata. There may be other code paths that this does not cover, meaning that other usages of this may fail at runtime in a GraalVM native image.

Here's another user who ran into this same issue:
https://github.com/oracle/graal/issues/5558

This change should fix GraalVM native image run when doing runtime class initialization. Loading the native library needs JNI access to `java.lang.Throwable.toString`

### DIFF
--- a/src/main/resources/META-INF/native-image/org.xerial/sqlite-jdbc/jni-config.json
+++ b/src/main/resources/META-INF/native-image/org.xerial/sqlite-jdbc/jni-config.json
@@ -1,5 +1,9 @@
 [
     {
+        "name":"java.lang.Throwable",
+        "methods":[{"name":"toString","parameterTypes":[] }]
+    },
+    {
         "name":"org.sqlite.core.DB",
         "allDeclaredMethods":true,
         "allPublicMethods": true,


pettyalex:
Welp, I should have checked active PRs. That's what I get for searching around Spring first to try and understand this :)

It looks like this includes this fix and much more https://github.com/xerial/sqlite-jdbc/pull/823

gotson:
> Welp, I should have checked active PRs. That's what I get for searching around Spring first to try and understand this :)
> 
> It looks like this includes this fix and much more #823

We should be able to merge #823 pretty soon, do we still need your PR ?

pettyalex:
@gotson #823 looks like a much more complete improvement. If it's getting merged soon, than this one can be closed. 

If #823 is going to take a while, then this could be merged and allow sqlite-jdbc to be used with default settings in GraalVM native projects right now. Your call.

gotson:
I prefer to wait for the #823

PR #827, opened 2023-01-10
links: use https for sqlite.org

imba-tjd:
I think the more `https` the better.

1. Replaced `http://www.sqlite.org` with `https://www.sqlite.org`
2. Replaced `http://sqlite.org` with `https://www.sqlite.org`
3. Replaced `https://sqlite.org` with `https://www.sqlite.org`
4. Changed another http to https in Readme


gotson:
When typing `https://www.sqlite.org` in a browser, one gets redirected to `https://sqlite.org`. I think we should not add the `www`.

gotson (approved):
LGTM

PR #829, opened 2023-01-16
feat: Add an ability to read metadata for attached databases

Destrolaric:
Hi SQLite team!
Here are my proposed changes to the reading of metadata. To allow to read of metadata from attached databases 
https://www.sqlite.org/lang_attach.html
~~This functionality will be optional and disabled by default because it adds more layers `[schema].[table]` for metadata reading.~~ 
The reason why this functionality will be useful:
It isn't easy to work with attached databases using JDBC API because all metadata reading assumes that we are in the main schema. This is fixable by adding schema information to the existing queries.

### DIFF
--- a/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
@@ -920,7 +920,7 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
             if (i == 0) {
                 sql.append("SELECT * FROM (");
             }
-            appendGetSchemaColumns(sql ,c, schemasNames.get(i), tblNamePattern, colNamePattern);
+            appendGetSchemaColumns(sql, c, schemasNames.get(i), tblNamePattern, colNamePattern);
             if (i != schemasNames.size() - 1) {
                 sql.append(" UNION ALL ");
             } else {
@@ -930,39 +930,38 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
         if (schemasNames.size() == 0) {
             sql.append("select ");
             sql.append("\tnull as TABLE_CAT,\n")
-                .append("\tnull as TABLE_SCHEM,\n")
-                .append("\tnull as TABLE_NAME,\n")
-                .append("\tnull as COLUMN_NAME,\n")
-                .append("\tnull as DATA_TYPE,\n")
-                .append("\tnull as TYPE_NAME,\n")
-                .append("\tnull as COLUMN_SIZE,\n")
-                .append("\tnull as BUFFER_LENGTH,\n")
-                .append("\tnull as DECIMAL_DIGITS,\n")
-                .append("\tnull as NUM_PREC_RADIX,\n")
-                .append("\tnull as NULLABLE,\n")
-                .append("\tnull as REMARKS,\n")
-                .append("\tnull as COLUMN_DEF,\n")
-                .append("\tnull as SQL_DATA_TYPE,\n")
-                .append("\tnull as SQL_DATETIME_SUB,\n")
-                .append("\tnull as CHAR_OCTET_LENGTH,\n")
-                .append("\tnull as ORDINAL_POSITION,\n")
-                .append("\tnull as IS_NULLABLE,\n")
-                .append("\tnull as SCOPE_CATLOG,\n")
-                .append("\tnull as SCOPE_SCHEMA,\n")
-                .append("\tnull as SCOPE_TABLE,\n")
-                .append("\tnull as SOURCE_DATA_TYPE,\n")
-                .append("\tnull as IS_AUTOINCREMENT,\n")
-                .append("\tnull as IS_GENERATEDCOLUMN\n")
-                .append("\t limit 0");
+                    .append("\tnull as TABLE_SCHEM,\n")
+                    .append("\tnull as TABLE_NAME,\n")
+                    .append("\tnull as COLUMN_NAME,\n")
+                    .append("\tnull as DATA_TYPE,\n")
+                    .append("\tnull as TYPE_NAME,\n")
+                    .append("\tnull as COLUMN_SIZE,\n")
+                    .append("\tnull as BUFFER_LENGTH,\n")
+                    .append("\tnull as DECIMAL_DIGITS,\n")
+                    .append("\tnull as NUM_PREC_RADIX,\n")
+                    .append("\tnull as NULLABLE,\n")
+                    .append("\tnull as REMARKS,\n")
+                    .append("\tnull as COLUMN_DEF,\n")
+                    .append("\tnull as SQL_DATA_TYPE,\n")
+                    .append("\tnull as SQL_DATETIME_SUB,\n")
+                    .append("\tnull as CHAR_OCTET_LENGTH,\n")
+                    .append("\tnull as ORDINAL_POSITION,\n")
+                    .append("\tnull as IS_NULLABLE,\n")
+                    .append("\tnull as SCOPE_CATLOG,\n")
+                    .append("\tnull as SCOPE_SCHEMA,\n")
+                    .append("\tnull as SCOPE_TABLE,\n")
+                    .append("\tnull as SOURCE_DATA_TYPE,\n")
+                    .append("\tnull as IS_AUTOINCREMENT,\n")
+                    .append("\tnull as IS_GENERATEDCOLUMN\n")
+                    .append("\t limit 0");
         }
         Statement stat = conn.createStatement();
         return ((CoreStatement) stat).executeQuery(sql.toString(), true);
     }
 
     private StringBuilder appendGetSchemaColumns(
-        StringBuilder sql, String c, String s, String tblNamePattern, String colNamePattern
-    )
-    throws SQLException {
+            StringBuilder sql, String c, String s, String tblNamePattern, String colNamePattern)
+            throws SQLException {
 
         sql.append("select null as TABLE_CAT, ")
                 .append(quote(s == null ? "main" : s))
@@ -1001,7 +1000,8 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
                     rsColAutoinc =
                             statColAutoinc.executeQuery(
                                     "SELECT LIKE('%autoincrement%', LOWER(sql)) FROM "
-                                            + prependSchemaPrefix(s, "sqlite_schema WHERE LOWER(name) = LOWER('")
+                                            + prependSchemaPrefix(
+                                                    s, "sqlite_schema WHERE LOWER(name) = LOWER('")
                                             + escape(tableName)
                                             + "') AND TYPE IN ('table', 'view')");
                     rsColAutoinc.next();
@@ -1026,7 +1026,8 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
                 // For each table, get the column info and build into overall SQL
                 String pragmaStatement =
                         "PRAGMA "
-                                + prependSchemaPrefix(s, "table_xinfo('" + escape(tableName) + "')");
+                                + prependSchemaPrefix(
+                                        s, "table_xinfo('" + escape(tableName) + "')");
                 try (Statement colstat = conn.createStatement();
                         ResultSet rscol = colstat.executeQuery(pragmaStatement)) {
 
@@ -1181,8 +1182,8 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
         } else {
             sql.append(
                     "select null as ordpos, null as colnullable, null as ct, null as colsize, null as "
-                        + "colDecimalDigits, null as tblname, null as cn, null as tn, null as colDefault, null as "
-                        + "colautoincrement, null as colgenerated limit 0)");
+                            + "colDecimalDigits, null as tblname, null as cn, null as tn, null as colDefault, null as "
+                            + "colautoincrement, null as colgenerated limit 0)");
         }
         return sql;
     }
@@ -1229,7 +1230,7 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
             getSchemas =
                     conn.prepareStatement(
                             "select name as TABLE_SCHEM, null as TABLE_CATALOG from pragma_database_list order by "
-                                + "TABLE_SCHEM;");
+                                    + "TABLE_SCHEM;");
         }
 
         return getSchemas.executeQuery();
@@ -1244,9 +1245,14 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
             schemaPattern = "main";
         }
         Statement stat = conn.createStatement();
-        String sql = "select name as TABLE_SCHEM, null as TABLE_CATALOG from pragma_database_list\n"
-            + "where TABLE_SCHEM like '" + schemaPattern + "' escape '" + getSearchStringEscape() + "' order by "
-            + "TABLE_SCHEM;";
+        String sql =
+                "select name as TABLE_SCHEM, null as TABLE_CATALOG from pragma_database_list\n"
+                        + "where TABLE_SCHEM like '"
+                        + schemaPattern
+                        + "' escape '"
+                        + getSearchStringEscape()
+                        + "' order by "
+                        + "TABLE_SCHEM;";
         return stat.executeQuery(sql);
     }
 
@@ -1279,10 +1285,11 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
         }
         if (schemaNames.size() == 0) {
             sql.append("select null as TABLE_CAT, ")
-                .append("null")
-                .append(" as TABLE_SCHEM, '")
-                .append("null")
-                .append("' as TABLE_NAME, null as COLUMN_NAME, null as KEY_SEQ, null as PK_NAME limit 0;");
+                    .append("null")
+                    .append(" as TABLE_SCHEM, '")
+                    .append("null")
+                    .append(
+                            "' as TABLE_NAME, null as COLUMN_NAME, null as KEY_SEQ, null as PK_NAME limit 0;");
         }
 
         return ((CoreStatement) stat).executeQuery(sql.toString(), true);
@@ -1296,7 +1303,8 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
         return schemaNames;
     }
 
-    private void createSchemaPrimaryKeysQuery(String s, String table, StringBuilder sql) throws SQLException {
+    private void createSchemaPrimaryKeysQuery(String s, String table, StringBuilder sql)
+            throws SQLException {
         PrimaryKeyFinder pkFinder = new PrimaryKeyFinder(table, s);
         String[] columns = pkFinder.getColumns();
         sql.append("select null as TABLE_CAT, ")
@@ -1365,10 +1373,11 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
                 // retrieve table list
                 ArrayList<String> tableList;
                 try (ResultSet rs =
-                    stat.executeQuery(
-                        "select name from "
-                            + prependSchemaPrefix(
-                            schema, "sqlite_schema where type = " + "'table'"))) {
+                        stat.executeQuery(
+                                "select name from "
+                                        + prependSchemaPrefix(
+                                                schema,
+                                                "sqlite_schema where type = " + "'table'"))) {
                     tableList = new ArrayList<>();
 
                     while (rs.next()) {
@@ -1410,29 +1419,34 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
                                 }
                             }
                             String pkName =
-                                (usePkName && pkFinder.getName() != null) ? pkFinder.getName() : "";
+                                    (usePkName && pkFinder.getName() != null)
+                                            ? pkFinder.getName()
+                                            : "";
 
                             exportedKeysQuery
-                                .append(count > 0 ? " union all select " : "select ")
-                                .append(keySeq)
-                                .append(" as ks, '")
-                                .append(escape(tbl))
-                                .append("' as fkt, '")
-                                .append(escape(FKColName))
-                                .append("' as fcn, '")
-                                .append(escape(PKColName))
-                                .append("' as pcn, '")
-                                .append(escape(pkName))
-                                .append("' as pkn, ")
-                                .append(RULE_MAP.get(foreignKey.getOnUpdate()))
-                                .append(" as ur, ")
-                                .append(RULE_MAP.get(foreignKey.getOnDelete()))
-                                .append(" as dr, ");
+                                    .append(count > 0 ? " union all select " : "select ")
+                                    .append(keySeq)
+                                    .append(" as ks, '")
+                                    .append(escape(tbl))
+                                    .append("' as fkt, '")
+                                    .append(escape(FKColName))
+                                    .append("' as fcn, '")
+                                    .append(escape(PKColName))
+                                    .append("' as pcn, '")
+                                    .append(escape(pkName))
+                                    .append("' as pkn, ")
+                                    .append(RULE_MAP.get(foreignKey.getOnUpdate()))
+                                    .append(" as ur, ")
+                                    .append(RULE_MAP.get(foreignKey.getOnDelete()))
+                                    .append(" as dr, ");
 
                             String fkName = foreignKey.getFkName();
 
                             if (fkName != null) {
-                                exportedKeysQuery.append("'").append(escape(fkName)).append("' as fkn");
+                                exportedKeysQuery
+                                        .append("'")
+                                        .append(escape(fkName))
+                                        .append("' as fkn");
                             } else {
                                 exportedKeysQuery.append("'' as fkn");
                             }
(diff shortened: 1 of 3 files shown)


gotson:
Hi, i don't think we should add a pragma for that. The JDBC methods accept a schema, and if it is `null` then it should not be used for filtering.

Destrolaric:
Hi, pragma is used only for getSchemas to mark if we read them at all. Other parts just check for null.

gotson:
> Hi, pragma is used only for getSchemas to mark if we read them at all. Other parts just check for null.

i don't think the switch is necessary, we can always retrieve them all

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -681,7 +681,7 @@ public boolean supportsSavepoints() {
 
     /** @see java.sql.DatabaseMetaData#supportsSchemasInDataManipulation() */
     public boolean supportsSchemasInDataManipulation() {
-        return false;
+        return true;
```
gotson, line 684: is this related to the change of this PR, or something we missed ? Database metadata is not DML, thus your changes should not change this value IMHO.

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -915,7 +915,7 @@ public ResultSet getColumns(String c, String s, String tblNamePattern, String co
         checkOpen();
 
         StringBuilder sql = new StringBuilder(700);
-        sql.append("select null as TABLE_CAT, null as TABLE_SCHEM, tblname as TABLE_NAME, ")
+        sql.append("select null as TABLE_CAT, ").append(quote(s)).append(" as TABLE_SCHEM, tblname as TABLE_NAME, ")
```
gotson, line 918: if `s` is null, then will it quote the `NULL` string thus returning `"NULL"` instead of the SQL `null` ?

gotson (commented):
i did a very quick first review

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
Destrolaric:
```
    protected static String quote(String tableName) {
        if (tableName == null) {
            return "null";
        } else {
            return String.format("'%s'", tableName);
        }
    }
```
It will be null for SQL requests. And also, it can be seen by unit tests that it will result in a `null` value.

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
Destrolaric: agree

gotson:
Code formatting is failing, can you run `mvn spotless:apply` ?

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -1173,9 +1172,8 @@ public ResultSet getCrossReference(
     /** @see java.sql.DatabaseMetaData#getSchemas() */
```
gotson, line 1172: My understanding is that schemas are currently stored instead of being retrieved everytime, but as databases can be attached dynamically, we should probably get the schemas all the time, and add a test to cover that too.

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1398,10 +1400,10 @@ public void columnOrderOfgetImportedKeys() throws SQLException {
         stat.executeUpdate(
                 "create table address (pid integer, name, foreign key(pid) references person(id))");
 
-        ResultSet importedKeys = meta.getImportedKeys("default", "global", "address");
+        ResultSet importedKeys = meta.getImportedKeys("default", null, "address");
         assertThat(importedKeys.next()).isTrue();
         assertThat(importedKeys.getString("PKTABLE_CAT")).isEqualTo("default");
-        assertThat(importedKeys.getString("PKTABLE_SCHEM")).isEqualTo("global");
+        assertThat(importedKeys.getString("PKTABLE_SCHEM")).isEqualTo(null);
```
gotson, line 1406: for this shouldn't we return `main` instead of `null`?

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1423,12 +1425,12 @@ public void columnOrderOfgetExportedKeys() throws SQLException {
         stat.executeUpdate(
                 "create table address (pid integer, name, foreign key(pid) references person(id))");
 
-        ResultSet exportedKeys = meta.getExportedKeys("default", "global", "person");
+        ResultSet exportedKeys = meta.getExportedKeys("default", null, "person");
         assertThat(exportedKeys.next()).isTrue();
         assertThat(exportedKeys.getString("PKTABLE_CAT")).isEqualTo("default");
-        assertThat(exportedKeys.getString("PKTABLE_SCHEM")).isEqualTo("global");
+        assertThat(exportedKeys.getString("PKTABLE_SCHEM")).isEqualTo(null);
```
gotson, line 1431: same here, it should be the actual schema of the table, not the parameter that was passed

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1423,12 +1425,12 @@ public void columnOrderOfgetExportedKeys() throws SQLException {
...
         assertThat(exportedKeys.getString("FKTABLE_CAT")).isEqualTo("default");
-        assertThat(exportedKeys.getString("FKTABLE_SCHEM")).isEqualTo("global");
+        assertThat(exportedKeys.getString("FKTABLE_SCHEM")).isEqualTo(null);
```
gotson, line 1433: same here, it should be the actual schema of the table, not the parameter that was passed

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1568,4 +1570,313 @@ public void version() throws Exception {
...
+            assertThat(rsMeta.getColumnCount()).isEqualTo(2);
+            assertThat(rsMeta.getColumnName(1)).isEqualTo("TABLE_SCHEM");
+            assertThat(rs.getString("TABLE_SCHEM")).isEqualTo("main");
+            assertThat(rs.next()).isTrue();
+            assertThat(rs.getString("TABLE_SCHEM")).isEqualTo("db2");
+            assertThat(rsMeta.getColumnName(2)).isEqualTo("TABLE_CATALOG");
+        }
+
+        @Test
+        public void getTablesForAttachedDatabase() throws SQLException {
```
gotson, line 1602: we should probably have a test where schema is passed as `null` in `getTables`, and verify that returned tables do have the correct `schema` set.

Destrolaric:
Hi, I Added 2 test cases for the default DB and for attaching an additional one after `getschemas()` call. Also fixed formatting issues and now `null` schema replaced with `main`

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -915,7 +915,9 @@ public ResultSet getColumns(String c, String s, String tblNamePattern, String co
         checkOpen();
 
         StringBuilder sql = new StringBuilder(700);
-        sql.append("select null as TABLE_CAT, ").append(quote(s)).append(" as TABLE_SCHEM, tblname as TABLE_NAME, ")
+        sql.append("select null as TABLE_CAT, ")
+                .append(quote(s == null ? "main" : s))
+                .append(" as TABLE_SCHEM, tblname as TABLE_NAME, ")
```
gotson, line 920: I'm not sure about that.

The function should return the actual schema of the table, no? But it seems we just feed back the parameter that was passed.

If we have 2 tables with the same name in 2 different schemas, and we call this function without specifying the schema `s`, then that function should return columns from both tables, but we would need to have `TABLE_SCHEM` to differentiate them.

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
gotson: still not sure about this one, i see you resolved it, but as `getSchemas` is only initialized if it's null, it would not be updated after a database is attached or detached.

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -1198,7 +1206,9 @@ public ResultSet getPrimaryKeys(String c, String s, String table) throws SQLExce
 
         Statement stat = conn.createStatement();
         StringBuilder sql = new StringBuilder(512);
-        sql.append("select null as TABLE_CAT, ").append(quote(s)).append(" as TABLE_SCHEM, '")
+        sql.append("select null as TABLE_CAT, ")
+                .append(quote(s == null ? "main" : s))
```
gotson, line 1210: similar as above, we should return the actual schema of the retrieved PK's table, not the parameter that was passed to the function

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -1249,7 +1259,7 @@ public ResultSet getExportedKeys(String catalog, String schema, String table)
 
         catalog = (catalog != null) ? quote(catalog) : null;
 
-        String quotedSchema = (schema != null) ? quote(schema) : null;
+        String quotedSchema = (schema != null) ? quote(schema) : quote("main");
```
gotson, line 1262: similar as above, we should return the actual schema of the retrieved PK's table, not the parameter that was passed to the function

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -1406,12 +1416,12 @@ public ResultSet getImportedKeys(String catalog, String schema, String table)
         sql.append("select ")
                 .append(quote(catalog))
                 .append(" as PKTABLE_CAT, ")
-                .append(quote(schema))
+                .append(quote(schema == null ? "main" : schema))
```
gotson, line 1419: similar as above, we should return the actual schema of the retrieved PK's table, not the parameter that was passed to the function

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -1406,12 +1416,12 @@ public ResultSet getImportedKeys(String catalog, String schema, String table)
...
                 .append(" as PKTABLE_SCHEM, ")
                 .append("ptn as PKTABLE_NAME, pcn as PKCOLUMN_NAME, ")
                 .append(quote(catalog))
                 .append(" as FKTABLE_CAT, ")
-                .append(quote(schema))
+                .append(quote(schema == null ? "main" : schema))
```
gotson, line 1424: similar as above, we should return the actual schema of the retrieved PK's table, not the parameter that was passed to the function

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -1526,7 +1536,7 @@ public ResultSet getIndexInfo(String c, String s, String table, boolean u, boole
         // define the column header
         // this is from the JDBC spec, it is part of the driver protocol
         sql.append("select null as TABLE_CAT,")
-                .append(quote(s))
+                .append(quote(s == null ? "main" : s))
```
gotson, line 1539: similar as above, we should return the actual schema of the retrieved PK's table, not the parameter that was passed to the function

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -1689,7 +1705,10 @@ public synchronized ResultSet getTables(
         StringBuilder sql = new StringBuilder();
         sql.append("SELECT").append("\n");
         sql.append("  NULL AS TABLE_CAT,").append("\n");
-        sql.append("  NULL AS TABLE_SCHEM,").append("\n");
+        sql.append("  ")
+                .append(quote(s == null ? "main" : s))
```
gotson, line 1709: similar as above, we should return the actual schema of the retrieved PK's table, not the parameter that was passed to the function

Destrolaric:
@gotson Ok, this one looks doable and will probably, be ready sometime next week.

Destrolaric:
@gotson Hi, sorry for waiting; unfortunately, I had no time to finish the pr before. I have a question. JDBC has some functions like getExportedKeys that require using full schema name and returns everything if null. Do we want to read the data from all existing schemas?

Destrolaric:
```
        sql.append("    SELECT").append("\n");
        sql.append("      NAME,").append("\n");
        sql.append("      'GLOBAL TEMPORARY' AS TYPE").append("\n");
        sql.append("    FROM").append("\n");
        sql.append("      sqlite_temp_master").append("\n"); 
```
Also question what to do with GLOBAL TEMPORARY tables. They kind of have their scheme now.

Destrolaric:
@gotson, Hi, again, can we discuss acceptance criteria regarding this pr? Currently, it needs to add some unit tests, and we can probably discuss if anything else is required. Also it looks like this one leads to the significant change of default behaviour, so it probably may require more accurate handling.

gotson:
> @gotson, Hi, again, can we discuss acceptance criteria regarding this pr? Currently, it needs to add some unit tests, and we can probably discuss if anything else is required. Also it looks like this one leads to the significant change of default behaviour, so it probably may require more accurate handling.

Hi, unit tests would be needed yes. What kind of significant changes of default behavior would you foresee though ?

Destrolaric:
@gotson Well, there are some. Most importantly, if the value of the schema pattern is null, it will show all DBs instead of only 'main', that may look like a significant API change.

gotson:
> @gotson Well, there are some. Most importantly, if the value of the schema pattern is null, it will show all DBs instead of only 'main', that may look like a significant API change.

It's fine, it is more correct

Destrolaric:
@gotson I resolved the conflict with the recent two commits

gotson:
I still see conflicts though. Are you sure you pushed your latest changes?

Destrolaric:
image
Weird, for me branch have no conflicts

gotson:
Weird indeed

SmartSelect_20230419_220608_Chrome

gotson:
My bad, it's only the rebase (which was selected by default) that can't be done.

Squash and merge is ok (and it's what we want).

SmartSelect_20230419_220755_Chrome

gotson:
code formatting is failing, can you fix it with `mvn spotless:apply` ?

Destrolaric:
@gotson Fixed

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -94,7 +94,10 @@ public void getTables() throws SQLException {
         rs = meta.getTables(null, null, null, new String[] {"system table"});
         assertThat(rs.next()).isTrue();
         assertThat(rs.getString("TABLE_NAME")).isEqualTo("sqlite_schema");
-        assertThat(rs.next()).isFalse();
+        assertThat(rs.getString("TABLE_SCHEM")).isEqualTo("main");
+        //        assertThat(rs.next()).isTrue();
+        //        assertThat(rs.getString("TABLE_NAME")).isEqualTo("sqlite_schema");
+        //        assertThat(rs.getString("TABLE_SCHEM")).isEqualTo("temp");
```
gotson, line 100: What's with those commented lines?

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1669,4 +1667,593 @@ public void getPrimaryKeys(String table) throws SQLException {
...
+        }
+
+        @AfterEach
+        public void exit() throws SQLException {
+            stat.executeUpdate("Detach database db2;");
+            testDB.deleteOnExit();
+        }
+    }
+
+    private void assertSystemSchema(ResultSet rs) throws SQLException {
```
gotson, line 2246: shouldn't that also check for `TABLE_SCHEM` and compare it to a parameter passed in the function?

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -417,18 +420,13 @@ public void getColumns() throws SQLException {
...
-        assertThat(rs.getString("COLUMN_NAME")).isEqualTo("tbl_name");
-        assertThat(rs.next()).isTrue();
-        assertThat(rs.getString("COLUMN_NAME")).isEqualTo("rootpage");
-        assertThat(rs.next()).isTrue();
-        assertThat(rs.getString("COLUMN_NAME")).isEqualTo("sql");
+
+        // SYSTEM TABLE "sqlite_schema" for main
+        assertSystemSchema(rs);
+
+        // SYSTEM TABLE "sqlite_schema" for temp
```
gotson, line 427: where is that `temp` database attached? i can't find any mention to it

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1001,7 +999,7 @@ public void columnOrderOfgetTables() throws SQLException {
         ResultSet rsTables =
                 meta.getTables(
                         null,
-                        null,
+                        "",
```
gotson, line 1002: I'm not sure about that, i don't think passing `""` should match any schema, that's what passing `null` is for.

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1158,7 +1156,7 @@ public void columnOrderOfgetProcedurColumns() throws SQLException {
     @Test
     public void columnOrderOfgetSchemas() throws SQLException {
         ResultSet rs = meta.getSchemas();
-        assertThat(rs.next()).isFalse();
+        assertThat(rs.next()).isTrue();
```
gotson, line 1159: shouldn't we have more assertions to verify the content of the result set?

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1669,4 +1667,593 @@ public void getPrimaryKeys(String table) throws SQLException {
...
+        assertThat(rsmeta.getColumnName(11)).isEqualTo("NULLABLE");
+        assertThat(rsmeta.getColumnName(12)).isEqualTo("REMARKS");
+        assertThat(rsmeta.getColumnName(13)).isEqualTo("COLUMN_DEF");
+        assertThat(rsmeta.getColumnName(14)).isEqualTo("SQL_DATA_TYPE");
+        assertThat(rsmeta.getColumnName(15)).isEqualTo("SQL_DATETIME_SUB");
+        assertThat(rsmeta.getColumnName(16)).isEqualTo("CHAR_OCTET_LENGTH");
+        assertThat(rsmeta.getColumnName(17)).isEqualTo("ORDINAL_POSITION");
+        assertThat(rsmeta.getColumnName(18)).isEqualTo("IS_NULLABLE");
+        // should be SCOPE_CATALOG, but misspelt in the standard
+        assertThat(rsmeta.getColumnName(19)).isEqualTo("SCOPE_CATLOG");
```
gotson, line 1725: this was fixed in 4429515442f999457e0517216c1e5e0c5b14cb09 and thus should be changed

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1669,4 +1667,593 @@ public void getPrimaryKeys(String table) throws SQLException {
...
+
+            ResultSet rs = meta.getIndexInfo(null, "db2", "testindex", false, false);
+            ResultSetMetaData rsmd = rs.getMetaData();
+
+            assertThat(rs).isNotNull();
+            assertThat(rsmd).isNotNull();
+        }
+
+        @AfterEach
+        public void exit() throws SQLException {
```
gotson, line 2240: it's better for readability to move this function next to the `init()` function

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1669,4 +1667,593 @@ public void getPrimaryKeys(String table) throws SQLException {
...
+        public void readSchemaWithAttachedDatabases() throws SQLException {
+            ResultSet rs = meta.getSchemas();
+            assertThat(rs.next()).isTrue();
+            ResultSetMetaData rsMeta = rs.getMetaData();
+            assertThat(rsMeta.getColumnCount()).isEqualTo(2);
+            assertThat(rsMeta.getColumnName(1)).isEqualTo("TABLE_SCHEM");
+            assertThat(rsMeta.getColumnName(2)).isEqualTo("TABLE_CATALOG");
+            assertThat(rs.getString("TABLE_SCHEM")).isEqualTo("db2");
+            assertThat(rs.next()).isTrue();
+            assertThat(rs.getString("TABLE_SCHEM")).isEqualTo("main");
```
gotson, line 1806: is there some ordering of the schemas mandated by JDBC ? or we do always return them in alphabetical order?

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1669,4 +1667,593 @@ public void getPrimaryKeys(String table) throws SQLException {
...
+            rs = meta.getTables(null, null, "bob", null);
+            assertThat(rs.next()).isFalse();
+            rs.close();
+            rs = meta.getTables(null, null, "test", null);
+            assertThat(rs.next()).isTrue();
+            assertThat(rs.next()).isFalse();
+            rs.close();
+            rs = meta.getTables(null, null, "test%", null);
+            assertThat(rs.next()).isTrue();
+            assertThat(rs.next()).isTrue();
```
gotson, line 1838: some comments on what is being tested would help

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1669,4 +1667,593 @@ public void getPrimaryKeys(String table) throws SQLException {
...
+            assertThat(rs.getString("TABLE_NAME")).isEqualTo("testView");
+            assertThat(rs.next()).isFalse();
+            rs.close();
+
+            rs = meta.getTables(null, "main", null, new String[] {"system table"});
+            assertThat(rs.next()).isTrue();
+            assertThat(rs.getString("TABLE_NAME")).isEqualTo("sqlite_schema");
+            assertThat(rs.next()).isFalse();
+            rs.close();
```
gotson, line 1857: shouldn't we have the same suite of tests for `db2` for consistency?

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1669,4 +1667,593 @@ public void getPrimaryKeys(String table) throws SQLException {
...
+            try {
+                schemas = meta.getSchemas();
+                boolean schemaFound = false;
+                while (schemas.next()) {
+                    schemaFound = "db3".equals(schemas.getString("TABLE_SCHEM"));
+                    if (schemaFound) {
+                        break;
+                    }
+                }
+                assertThat(schemaFound).isTrue();
```
gotson, line 1874: maybe put all schemas in a list and verify that it contains all the values we expect, ie `db2`,`db3` and `main`.
If we expect some ordering, we should also test for that here.

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1669,4 +1667,593 @@ public void getPrimaryKeys(String table) throws SQLException {
...
+
+            } finally {
+                stat.executeUpdate("detach database db3;");
+                thirdDB.deleteOnExit();
+            }
+        }
+
+        @Test
+        public void testGetSchemasForAttachedDatabases() throws SQLException, IOException {
```
gotson, line 1883: the above function mentions it's about dynamic attachment, but this one doesn't

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -1669,4 +1667,593 @@ public void getPrimaryKeys(String table) throws SQLException {
...
+                while (schemas.next()) {
+                    schemaFound = "db3".equals(schemas.getString("TABLE_SCHEM"));
+                    if (schemaFound) {
+                        break;
+                    }
+                }
+                assertThat(schemaFound).isTrue();
+
+            } finally {
+                stat.executeUpdate("detach database db3;");
```
gotson, line 1877: should we also test after the detach that the schemas are properly returned?

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -1175,12 +1229,33 @@ public ResultSet getSchemas() throws SQLException {
...
     }
 
+    /** @see java.sql.DatabaseMetaData#getSchemas(String, String) */
+    public ResultSet getSchemas(String catalog, String schemaPattern) throws SQLException {
+        if (schemaPattern == null) {
+            return getSchemas();
+        }
+        if ("".equals(schemaPattern)) {
+            schemaPattern = "main";
+        }
```
gotson, line 1246: I don't think we should have that behaviour. The documentation says:

> schemaPattern – a schema name; must match the schema name as it is stored in the database; null means schema name should not be used to narrow down the search.

There is no mention of a special handling when passing `"""`

gotson (changes requested):
For this review i mostly focused on the test class. I think we should have more coverage on dynamic attachment, ie for all functions:
- test the behaviour
- attach 1 db
- test the behaviour
- detach the db
- test the behaviour

Is that something you could add ?

src/test/java/org/sqlite/DBMetaDataTest.java
Destrolaric:
```
"select name as TABLE_SCHEM, null as TABLE_CATALOG from pragma_database_list order by "
                                    + "TABLE_SCHEM;");
```
Ordering in SQL request

src/test/java/org/sqlite/DBMetaDataTest.java
Destrolaric: temp is the sqlite temporary files, I think are returned as system external schema if any exist.

src/test/java/org/sqlite/DBMetaDataTest.java
gotson: Just noticed that:

> The results are ordered by TABLE_CATALOG and TABLE_SCHEM

In that case the tests should test for the ordering.

PR #835, opened 2023-02-01
native-image - use Feature implementation instead of config files

kkriske:
This PR replaces the `resource-config.json` and `jni-config.json` files by a `org.graalvm.nativeimage.hosted.Feature` implementation.
* The registered JNI configuration is unchanged, but is now configured using the native-image java API. This is easier to maintain than the json files.
* The resource config is different, as previously all library resources were included. Now only the library relevant for the target runtime of the native-image compilation is included as a resource. To do this, the existing logic in `SQLiteJDBCLoader` was reused. Additionally the metadata resources containing version info and properties are read at build-time in static code blocks which avoid the need to include those resources and avoid the need to do the IO operations at runtime.

Finally, I added a new `org.sqlite.lib.exportPath` which takes in a directory that is used by the `SqliteJdbcFeature` to export the JNI library to at build-time. If this property is set, the library will no longer be included as a resource and it becomes the responsibility of the end user to provide the exported library next to the native-image executable.
The benefit of doing this is that the library no longer needs to be unpacked at runtime and again some IO operations can be eliminated.


pyckle:
Nice work (: This should reduce native image size by only including the relevant JNI lib.

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
```diff
@@ -0,0 +1,149 @@
...
+                "META-INF/maven/org.xerial/sqlite-jdbc/pom.properties");
+        RuntimeResourceAccess.addResource(
+                SQLiteJDBCLoader.class.getModule(),
+                "META-INF/maven/org.xerial/sqlite-jdbc/VERSION");
+        RuntimeResourceAccess.addResource(
+                JDBC3DatabaseMetaData.class.getModule(), "sqlite-jdbc.properties");
+
+        // TODO need a smarter way to get this resource location
+        String libraryResource;
+        if (Platform.includedIn(Platform.WINDOWS_AMD64.class)) {
```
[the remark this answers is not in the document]
kkriske: That was indeed the idea, hence the TODO. I didn't have time to look at it too closely yet but it will probably need some refactoring still because there is currently no clear cut `String getLibraryResourcePathForCurrentOS()` method.

For android, I'm a bit confused how this works, specifically why this exists, and why android is treated specially:
https://github.com/xerial/sqlite-jdbc/blob/78633768a40c13d244799c2527f5cf20eb8ba3af/src/main/java/org/sqlite/core/NativeDB.java#L42-L52

kkriske:
> This should reduce native image size by only including the relevant JNI lib.

That's the intention indeed. I would like to go one step further even, and allow to choose between having the resource unpacked at build-time or at runtime, probably in a property that can be set at build time. Because having it stored internally is bad for startup time, and it is entirely possible to unpack it at build-time and store it next to the executable. Then no library would need to be stored in the executable itself, and it could be loaded directly from next to the executable.

pom.xml
```diff
@@ -381,6 +386,17 @@
...
+        This dependency makes compilation on non-GraalVM versions possible.
+        The dependency should however never actually be required by end-users because if they require
+        GraalVM features, they will be provided by the JDK they are using.
+        -->
+        <dependency>
+            <groupId>org.graalvm.sdk</groupId>
+            <artifactId>graal-sdk</artifactId>
+            <version>${graalvm.version}</version>
+            <scope>provided</scope>
+        </dependency>
```
kkriske, line 399: Sidenote on this, provided scope works perfectly fine as discussed earlier. Seems like my understanding of the provided scope was wrong.

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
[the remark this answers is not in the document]
pyckle: > That was indeed the idea, hence the TODO

Whoops, I missed that.

> For android, I'm a bit confused how this works, specifically why this exists, and why android is treated specially:

Seems like it was done over 10 years ago for a reason that seems no longer relevant:
https://github.com/xerial/sqlite-jdbc/commit/097530bd109143e1da48a901928aefda04fd5d10

pom.xml
pyckle: Thanks for looking into this (: I added a comment in the other thread in case others reference it, and perhaps there's something we're missing here.

pyckle:
> you could argue that startup time isn't really impacted that much

This sort of bloat adds up as software use more and more libraries. I think that adding this as an option is useful and wise.

src/main/java/org/sqlite/SQLiteJDBCLoader.java
```diff
@@ -326,20 +327,16 @@ private static void loadSQLiteNativeLibrary() throws Exception {
...
+        boolean hasNativeLib =
+                LibraryLoaderUtil.hasNativeLib(sqliteNativeLibraryPath, sqliteNativeLibraryName);
+
+        if (!hasNativeLib && OSInfo.getOSName().equals("Mac")) {
+            // Fix for openjdk7 for Mac
+            String altName = "libsqlitejdbc.jnilib";
+            if (LibraryLoaderUtil.hasNativeLib(sqliteNativeLibraryPath, altName)) {
+                sqliteNativeLibraryName = altName;
+                hasNativeLib = true;
             }
```
kkriske, line 340: Docs indicate this is only required for JDK7, minimal version is JDK8 so can this be removed?

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        // libraryResource always has a leading '/'
+        String libraryResource = libraryPath + "/" + libraryName;
+        String exportLocation = System.getProperty("org.sqlite.lib.exportPath", "");
+        if (exportLocation.isEmpty()) {
+            // Do not export the library and include it in native-image instead
+            RuntimeResourceAccess.addResource(
+                    SQLiteJDBCLoader.class.getModule(), libraryResource.substring(1));
+        } else {
```
kkriske, line 56: Currently only this case is tested.
I documented the other case as "recommended" use-case, but most people will probably still just use the as-is setup without bothering with additional configuration.
Ideally, both flows are tested, but that adds overhead to the CI. What would be your recommendation?
1. only test the default flow, where the libarary is included in the native-image and unpacked at run-time
2. only test the configured flow, where the library is unpacked at build-time and is present next to the native-image at run-time
3. test both cases in the CI

README.md
```diff
@@ -114,6 +114,41 @@ In the other OSs not listed above, the pure-java SQLite is used. (Applies to ver
 
 If you want to use the native library for your OS, [build the source from scratch](./CONTRIBUTING.md).
 
+## GraalVM native-image support
+
+Sqlite JDBC supports [GraalVM native-image](https://www.graalvm.org/native-image/) out of the box starting from version 3.40.1.0.
```
kkriske, line 119: I'm not entirely sure if this is the right version to quote.

gotson:
@kkriske i see you moved the PR from draft to ready, but it doesn't have any description at all.

Would you be able to provide some general description of what this changes aim to address, and how?

gotson:
@kkriske looks like there's a conflict to resolve

kkriske:
> @kkriske is it possible to somehow statically link the exported library into resulting executable?

No. It's a dynamic library. Static linking would require building static libraries.
It would be possible to statically link this hypothetical static library but that would rely on native-image internal api's which I would not recommend using as a library. That would force the end user to add certain `--add-exports` arguments to the native-image command in order to allow the build to work and obviously there's no stability guarantee of that api.

kkriske:
@gotson FYI this should be ready for consideration/review.

gotson:
> @gotson FYI this should be ready for consideration/review.

i will have a look once i have some time. Can you fix the conflicts in the meantime?

kkriske:
There didn't seem to be any conflicts, but I have updated the feature branch.

gotson:
> There didn't seem to be any conflicts, but I have updated the feature branch.

Yes my bad, the github screen was showing conflicts for a rebase, but not for a squash merge. I don't know why. Thanks.

.github/workflows/ci.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           components: 'native-image'
           github-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Test
-        run: mvn --batch-mode --no-transfer-progress -P native test
+        run: mvn --batch-mode --no-transfer-progress -P native verify
```
gotson, line 83: why the change ?

README.md
```diff
@@ -114,6 +114,41 @@ In the other OSs not listed above, the pure-java SQLite is used. (Applies to ver
...
+There has been rudimentary support for some versions before that, but this was not actively tested by the CI.
+
+By default, the `sqlitejdbc` library for the compilation target will be included in the native image, accompanied the required JNI configuration.
```
gotson, line 122:
```suggestion
By default, the `sqlitejdbc` library for the compilation target will be included in the native image, accompanied by the required JNI configuration.
```

README.md
```diff
@@ -114,6 +114,41 @@ In the other OSs not listed above, the pure-java SQLite is used. (Applies to ver
...
+At runtime, this library will be extracted to the temp folder and be loaded from there.
```
gotson, line 123:
```suggestion
At runtime, this library will be extracted to the temp folder and loaded from there.
```

README.md
```diff
@@ -114,6 +114,41 @@ In the other OSs not listed above, the pure-java SQLite is used. (Applies to ver
...
+For the best startup performance however, it is recommended to set the `org.sqlite.lib.exportPath` property at build-time.
```
gotson, line 124:
```suggestion
For faster startup however, it is recommended to set the `org.sqlite.lib.exportPath` property at build-time.
```

README.md
```diff
@@ -114,6 +114,41 @@ In the other OSs not listed above, the pure-java SQLite is used. (Applies to ver
...
+This will export the `sqlitejdbc` library at build-time to the specified directory, after which the library will no longer be included as a resource.
```
gotson, line 125:
```suggestion
This will export the `sqlitejdbc` library at build-time to the specified directory, and the library will not be included as a resource.
```

pom.xml
```diff
@@ -344,7 +345,7 @@
                                 <goals>
                                     <goal>test</goal>
                                 </goals>
-                                <phase>test</phase>
+                                <phase>integration-test</phase>
```
gotson, line 348: what does this do compared to `test` ?

src/main/java/org/sqlite/SQLiteJDBCLoader.java
gotson: yes, well spotted, we can remove that

src/main/java9/module-info.java
```diff
@@ -2,6 +2,7 @@
 
     requires transitive java.sql;
     requires transitive java.sql.rowset;
+    requires static org.graalvm.sdk;
```
gotson, line 5: wouldn't that have any effects when running with a normal JDK?

src/main/java/org/sqlite/core/NativeDB.java
```diff
@@ -36,20 +36,8 @@ public final class NativeDB extends DB {
...
-
-    static {
-        if ("The Android Project".equals(System.getProperty("java.vm.vendor"))) {
-            System.loadLibrary("sqlitejdbc");
-            isLoaded = true;
-            loadSucceeded = true;
-        } else {
-            // continue with non Android execution path
-            isLoaded = false;
-            loadSucceeded = false;
```
gotson, line 50: i don't see where that bit ended up being moved to?

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
gotson: we should test both cases in the CI

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
```diff
@@ -0,0 +1,157 @@
...
+        // NativeDB and DB JNI calls
+        RuntimeJNIAccess.register(NativeDB.class);
+        RuntimeJNIAccess.register(
+                fields(
+                        NativeDB.class,
+                        "pointer",
+                        "busyHandler",
+                        "commitListener",
+                        "updateListener",
+                        "progressHandler"));
```
gotson, line 78: do those references have to be strings ? i am not familiar with GraalVM Feature.

.github/workflows/ci.yml
kkriske: to include the `integration-test` maven lifecycle, which runs later in the process than `test`.

pom.xml
kkriske: It runs in a different time in the maven lifecycle.
`test` runs after the code is compiled, and uses `target/classes` on the classpath to resolve the code.
`integration-test` runs after the jar is packaged, and uses the jar itself on the classpath instead of the classfiles.

The jar is a multi-release jar, and the module-info is contained in `META-INF/versions/9` which is automatically used if a jar is loaded. This doesn't work in a non-packaged layout like `target/classes`. So in order to pick up the module-info, the tests need to run on the jars. The module-info is required because the native-image builder always runs on the module system.

src/main/java9/module-info.java
kkriske: It shouldn't have.
It expresses a compile-time dependency that is optional at runtime
> There is also a `requires static` directive to indicate that a module is required at compile time, but is optional at runtime.

https://www.oracle.com/be/corporate/features/understanding-java-9-modules.html

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
kkriske: The references passed to `RuntimeJNIAccess#register` need to be either `Class` references, or reflection elements like `Field`, `Method` and `Constructor`. So to look them up they unfortunately need to be string references indeed.

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
kkriske: Turns out this is still somewhat an issue.
The second case, where the library is unpacked and placed next to the native-image does not work out of the box _yet_ on unix systems. It will work in the next version because of this change: https://github.com/oracle/graal/pull/5932
However it does already work if you add `-Djava.library.path=<path to target folder>` to the runtime arguments.
On windows it does already work in the current version as it already takes the dir of the executable into account.

So to reproduce, when running `mvn verify -Pnative` on a unix system, there will be a `native-tests` executable left in the target directory.
So `./native-tests` will run the tests, and replicate the failures the CI is currently seeing.
`./native-tests -Djava.library.path=.` will run all tests successfully.

What is the best approach here. Document this as a current limitation, to be resolved in GraalVM 23.0? Or wait for the release and set 23.0 as the minimal supported version?

gotson: we should have 2 different CI job to test each scenario, for each of the 3 OSes. If i understand correctly, for now on Linux (and Mac?) it would fail. We can comment that CI job for now, indicating that it should be solved with the new version.

README.md
gotson: It should be fine.

.github/workflows/ci.yml
gotson: the issue i see is that `verify` will also run spotless, which is already ran in a separate CI job. Can we run `test integration-test` ? Do we even need to run `test` for the native image ? Can't we just run `integration-test` only ?

src/main/java/org/sqlite/core/NativeDB.java
gotson: it is very much relevant, without this bit the library won't run anymore on Android.

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
gotson: can't we use reflection ?

gotson:
it seems CI is failing for GraalVM

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
kkriske: It is using reflection. It's just using helper `fields` and `method` methods to hide error handling and to make things a little bit easier to read. I took some inspiration from how internal GraalVM Feature implementations are organised: https://github.com/oracle/graal/blob/ff5373e4db4ab37ab7c6b3e8fc8a2dc346cadce1/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JNIRegistrationUtil.java#L90-L105

I can just inline the calls if you prefer that.

.github/workflows/ci.yml
kkriske: spotless doesn't seem to be registered to run in any default maven phase, so unless you call it directly (such as in the CI) it doesn't run. So `verify` should be fine, as it should be functionally the same as targetting `integration-test` only.

`test` will need to be executed, adding `-DskipTests` will also skip the native tests.

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
gotson: the problem i have with that is that there's strings, i would much prefer that the method references be there if we can, so it would be refactoring-proof.

kkriske: I don't understand what you're hinting at. They need to be `Method` and `Field` objects, and to my knowledge there is no other way to get those than a string-lookup.
Any refactoring that requires a change here, will also require a change in the native JNI part of the code. Those also use string lookups and would fail in much the same way if they were not updated.
The tests running on native-image, and straight up bailing out of the native-image compilation by throwing exceptions if one of the reflective lookups fail should be sufficient to catch any potential issue that can occur by forgetting to update these string references.

.github/workflows/ci.yml
```diff
@@ -84,7 +86,7 @@ jobs:
           components: 'native-image'
           github-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Test
-        run: mvn --batch-mode --no-transfer-progress -P {{ matrix.profiles }} verify
+        run: mvn --batch-mode --no-transfer-progress -P ${{ matrix.profiles }} verify
```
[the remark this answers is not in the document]
kkriske: As already pointed out here (https://github.com/xerial/sqlite-jdbc/pull/835#discussion_r1175638584), spotles doesn't run on `verify`, but yes `integration-test` is sufficient and I've changed it.

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
gotson: I have no experience with Graalvm, that is why my questions may be naive. If there's no other way that's fine.

.github/workflows/ci.yml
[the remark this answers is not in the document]
gotson: You're right. I am not familiar with Maven, Spotless wiki [mentions it](https://github.com/diffplug/spotless/tree/main/plugin-maven#binding-to-maven-phase). I thought it was automatically ran, but there's still a code block to add for execution.

I might add that at some point, so as to reduce the number of PR not following code style, in which case it's better not to use `verify` in the CI. Thanks!

gotson (approved):
A very solid PR, thanks a lot for the hard work!

pyckle:
> A very solid PR, thanks a lot for the hard work!

Seconded. Very nice work done here!

PR #839, opened 2023-02-12
remove temp files post test execution

kkriske:
Leverage the built-in JUnit5 `@TempFile` functionality to automatically remove any temp files created in those temp directories.
This avoids leaving files in `java.io.tmpdir` or in the `target` folder. JUnit5 will also fails the test if any of the created files cannot be deleted.
This was mainly caused by tests where database connections were not properly disposed of, but there is one change in `SQLiteConnection` because the `JarURLConnection` keeps an internal cache of the jar files in which the resourcelookup is happening. This caused `ConnectionTest#openJARResource` to fail as it could not delete the jar file.

### DIFF
--- a/src/main/java/org/sqlite/SQLiteConnection.java
+++ b/src/main/java/org/sqlite/SQLiteConnection.java
@@ -6,6 +6,7 @@ import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.net.URLConnection;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
 import java.sql.Connection;
@@ -330,7 +331,10 @@ public abstract class SQLiteConnection implements Connection {
             //            }
         }
 
-        try (InputStream reader = resourceAddr.openStream()) {
+        URLConnection conn = resourceAddr.openConnection();
+        // Disable caches to avoid keeping unnecessary file references after the single-use copy
+        conn.setUseCaches(false);
+        try (InputStream reader = conn.getInputStream()) {
             Files.copy(reader, dbFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
             return dbFile;
         }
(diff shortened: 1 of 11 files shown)


gotson:
> JUnit5 will also fails the test if any of the created files cannot be deleted.

not sure i get it, is that the existing behavior to fail or the new behavior (within this PR) ?

gotson:
I see you use `@TempDir`, which i attempted to use in https://github.com/xerial/sqlite-jdbc/commit/e14c3b83dcd033244f770840e41af4ab92d60cd7 but i encountered issues on Windows because of https://github.com/junit-team/junit5/issues/2811

kkriske:
> not sure i get it, is that the existing behavior to fail or the new behavior (within this PR) ?

Failing the test if the created temp files cannot be deleted is new (wanted) behavior. Previously, calls to `File#delete` or `File#deleteOnExit` would silently let the file be if it cannot be removed, but leftover files indicate either an issue in the test or an issue in the code (e.g. the jar file handle in `JarURLConnection`).

What you're missing in the linked changes is that you also need to close the connection object, to release the file handle before the file can be removed.

I have worked with `@TempDir` quite a bit, on windows as main OS as well, and have yet to encounter any issues with it that didn't end up being user-error. So allow me to be frank, that JUnit issue you linked is a whole lot of user-error and it's confusing to other developers that stumble upon that issue as it wrongly justifies what they're seeing as a bug in JUnit instead of a bug in their own code that they should fix.
/rant

gotson:
> Failing the test if the created temp files cannot be deleted is new (wanted) behavior

why would we want that behavior?

kkriske:
> why would we want that behavior?

Because it indicates a resource leak

> It worked on Linux and Mac, not on Windows, so i blame the framework.

I blame the OS filesystem implementation and/or the JVM Windows specific IO implementation. Windows has a long history of having a worse filesystem than UNIX based systems. Either way, resources should be closed and if the file system is capable of handling dangling file handlers that's great, but Windows in this case cannot.

gotson:
Thank you !

PR #844, opened 2023-02-28
feat(SQLiteDataSource): add setBusyTimeout

mruddy:
add SQLiteDataSource.setBusyTimeout(int milliseconds) so that we can do:
```
    final SQLiteDataSource ds = new SQLiteDataSource();
    ds.setBusyTimeout(1234);
```

instead of doing like below, or setting up a Properties object and providing it to the SQLiteDataSource constructor:

```
    final SQLiteDataSource ds = new SQLiteDataSource();
    ds.getConfig().setBusyTimeout(2000);
```

Note that the new test will fail because it depends on the change made by https://github.com/xerial/sqlite-jdbc/pull/843.

Without that fix, the line `assertThat(ds.getConfig().getBusyTimeout()).isEqualTo(1234);` in the new test will fail.

### DIFF
--- a/src/main/java/org/sqlite/SQLiteDataSource.java
+++ b/src/main/java/org/sqlite/SQLiteDataSource.java
@@ -140,6 +140,17 @@ public class SQLiteDataSource implements DataSource {
     }
 
     /**
+     * Sets the amount of time that the connection's busy handler will wait when a table is locked.
+     *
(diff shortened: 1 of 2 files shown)


mruddy:
just added a second commit that adds (somewhat crazy looking) source code formatting updates generated `by running mvn spotless:apply`

gotson:
thanks for your contribution!

PR #846, opened 2023-03-05
chore(SQLiteConnection): enforce currentTransactionMode not being null

mruddy:
having null as a possible state for `SQLiteConnection.currentTransactionMode` served no purpose.
these changes include mandating that it is not null and using that fact to simplify accordingly.

### DIFF
--- a/src/main/java/org/sqlite/SQLiteConnection.java
+++ b/src/main/java/org/sqlite/SQLiteConnection.java
@@ -64,9 +64,9 @@ public abstract class SQLiteConnection implements Connection {
         try {
             this.db = newDB = open(url, fileName, prop);
             SQLiteConfig config = this.db.getConfig();
-            this.connectionConfig = this.db.getConfig().newConnectionConfig();
+            this.connectionConfig = config.newConnectionConfig();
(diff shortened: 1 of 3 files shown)


src/main/java/org/sqlite/SQLiteConnection.java
```diff
@@ -86,6 +86,9 @@ public TransactionMode getCurrentTransactionMode() {
     }
 
     public void setCurrentTransactionMode(final TransactionMode currentTransactionMode) {
+        if (null == currentTransactionMode) {
+          throw new NullPointerException("currentTransactionMode must not be null");
```
gotson, line 90: you should not throw NPE, maybe `IllegalArgumentException` instead

mruddy:
also, this patch does away with an inconsistency in the current implementation where the `currentTransactionMode` is set to the configured non-auto-commit transaction mode in the constructor. it doesn't make a runtime difference, but to be consistent with the existing `setAutoCommit`, it should have been set to null. this is because the java.sql.Connection.setAutoCommit docs state: "By default, new connections are in auto-commit mode."

PR #847, opened 2023-03-06
feat: add more comprehensive native sqlite3 read only, add minimum compile time SQLITE_VERSION_NUMBER check

mruddy:
this pr adds a more comprehensive capability to determine if the database was opened read-only.

there are a few ways that the current read only logic misses cases where the database was actually opened in read only mode.
one example is when the file permissions do not allow writing, but do allow reading. the file will be opened in read only mode even though the config requested non-read-only mode. there are some other cases dealing with uri filenames too.

this new more comprehensive checking is done by `sqlite3_db_readonly` which has been in sqlite since 2012-03-20 (3.7.11) according to https://www.sqlite.org/changes.html (Version 3.7.16 and later according to sqlite3.c version 3.41.0). while trying to determine if this change would require a newer version of sqlite than had been previously, i found that the current minimum sqlite version needed before making this change is 2021-11-27 (3.37.0) (due to the use of `sqlite3_changes64`) (Version 3.36.1 and later according to sqlite3.c version 3.41.0). so, i made an update in NativeDB.c to document these findings.

in order to keep this pr manageable, it does not use this new read only checking capability. these changes will be proposed in a future pr.


src/main/java/org/sqlite/core/DB.java
```diff
@@ -80,6 +80,16 @@ public SQLiteConfig getConfig() {
...
 
+    /**
+     * Determines whether the database named databaseName of this connection is read-only.
+     *
+     * @param databaseName name of the database to inquire about (may be null)
+     * @throws SQLException if the connection is already closed
+     * @see <a href="https://www.sqlite.org/c3ref/db_readonly.html">https://www.sqlite.org/c3ref/db_readonly.html</a>
+     *      <a href="https://www.sqlite.org/c3ref/open.html">https://www.sqlite.org/c3ref/open.html</a>
+     */
+    public abstract boolean isReadOnly(final String databaseName) throws SQLException;
```
gotson, line 91: what 's the point of passing the database name ? is that to handle attached databases?

returning a boolean doesn't handle the case where the provided db name is not available on the connection:

> The sqlite3_db_readonly(D,N) interface returns 1 if the database N of connection D is read-only, 0 if it is read/write, or -1 if N is not the name of a database on connection D.

src/test/java/org/sqlite/ConnectionTest.java
```diff
@@ -50,13 +50,16 @@ public void executeUpdateOnClosedDB() throws SQLException {
...
-    public void readOnly() throws SQLException {
-
+    public void readOnly() throws SQLException, IOException {
         // set read only mode
         SQLiteConfig config = new SQLiteConfig();
         config.setReadOnly(true);
 
         try (Connection conn = DriverManager.getConnection("jdbc:sqlite:", config.toProperties())) {
+            assertThat(((SQLiteConnection) conn).getDatabase().isReadOnly("main")).isTrue();
+            assertThat(((SQLiteConnection) conn).getDatabase().isReadOnly(null)).isTrue();
```
gotson, line 60: what does that mean exactly ? what would be the `null` database, and why is it readonly ?

src/test/java/org/sqlite/ConnectionTest.java
```diff
@@ -50,13 +50,16 @@ public void executeUpdateOnClosedDB() throws SQLException {
...
+            assertThat(((SQLiteConnection) conn).getDatabase().isReadOnly("temp")).isFalse();
+            assertThat(((SQLiteConnection) conn).getDatabase().isReadOnly("false")).isFalse();
```
gotson, line 62: what is this `"false"` database ?

src/main/java/org/sqlite/core/DB.java
mruddy: yes, to allow for checking against attached databases. there's a bit of a disconnect between the traditional concept of a connection (that only connects to a single database) and sqlite's implementation of a connection as a collection of traditional connections.

the native method does allow for differentiating all the conditions. `isReadOnly` abstracts away the case of the not found database into "well, it's not read only". this main user of this method will be the implementer of Connection.isReadOnly, and their only great option is to pass "main" (which always exists in sqlite), so pushing a more than boolean result back up to that caller doesn't provide much value that i can see. i could return an enum out of this isReadOnly, but i don't see that actually being useful, especially when we could add another method to specifically test whether a database is found on the connection just by calling is_read_only and checking for -1.

gotson:
> in order to keep this pr manageable, it does not use this new read only checking capability. these changes will be proposed in a future pr.

that confused me, i would prefer to have all the changes at once so it's easier to understand and keep track of.

src/main/java/org/sqlite/core/DB.java
gotson: I feel like the DB implementation should mirror what the sqlite function returns, then we can abstract what we want to return in user land. We may want to throw an exception if a named database doesn't exist on a connection for instance.

As for `java.sql.Connection#isReadOnly`, which is not yet in this PR (and should be IMO), it probably should consider all the attached databases in order to say whether the connection is read only or not, don't you think?

src/test/java/org/sqlite/ConnectionTest.java
gotson: does that mean calling `sqlite3_db_readonly` with `null` will return `1` ?

src/test/java/org/sqlite/ConnectionTest.java
gotson: > but "false" made sense too since the result should be "false".

i feel this is confusing, a more explicit string would be best

src/main/java/org/sqlite/core/DB.java
mruddy: that seems like a slippery slope into using exceptions for flow control. referring to conventional wisdom: "Item 69: Use exceptions only for exceptional conditions" in [effective java 3rd ed](https://github.com/GunterMueller/Books-3/raw/master/Effective%20Java%20(3rd%20Edition).pdf)

`A class with a "state-dependent" method that can be invoked only under certain unpredictable conditions should generally have a separate "state-testing" method indicating whether it is appropriate to invoke the state-dependent method.`

with that in mind, how about we add an abstract method to DB.java and then implement it in NativeDB.java like this:
```
    @Override
    public boolean hasDatabase(final String databaseName) throws SQLException {
        return this.is_read_only(databaseName) != -1;
    }
```
with this, a caller can check and determine if the database is available on the connection.

it's just a matter of style. if you'd rather, i can make DB.isReadOnly return an enum with the 3 states of readonly, not readonly, and not found. maybe you prefer this as it avoids one native call. i could go either way as long as we don't venture into using exceptions for flow control.

regarding `java.sql.Connection#isReadOnly`, i don't think we should consider all databases available on the connection by default. some databases can be read/write while others are read-only. the temp database, and memory databases, i think, are always read-write, so we could not do an AND between all databases.
we could return that we are read-only if any of the databases are read-only.
maybe we make a SQLiteConfig option to let people choose how to determine if the connection is considered read-only. that's probably the best way to avoid tripping some people up. we'd default to basing the decision on only the "main" database, like it does now.

for reference, we can get a list of databases currently on the connection with:
`pragma database_list;` or `select * from pragma_database_list;`

src/test/java/org/sqlite/ConnectionTest.java
mruddy: yep, it can, if the main database is read-only. i made sure to hit this case in the tests.

src/test/java/org/sqlite/ConnectionTest.java
mruddy: ok, i'll update this name when i make any other changes that we decide on in sibling comments.

mruddy:
> that confused me, i would prefer to have all the changes at once so it's easier to understand and keep track of.

the patch gets large quickly because it starts messing with some things that were added by https://github.com/xerial/sqlite-jdbc/pull/782/files. i'd really like to get this patch in before i put up the other changes.

src/main/java/org/sqlite/core/DB.java
gotson: I am not saying we should use exceptions for control flow, just that exposing more information in the DB function seems like a good idea.

> i don't think we should consider all databases available on the connection by default. some databases can be read/write while others are read-only

Exactly. But JDBC doesn't care about how many DBs are on the connection, it needs to know whether it is read only or not.

> so we could not do an AND between all databases.
> we could return that we are read-only if any of the databases are read-only

the end result is the same. If any DB on the connection is read only, it should return `true`

> maybe we make a SQLiteConfig option to let people choose how to determine if the connection is considered read-only

I am not in favor of that. JDBC always needs some interpretation anyway, depending on the underlying database.



> we'd default to basing the decision on only the "main" database, like it does now.

at the moment we only check the configuration AFAIK, and it would apply to attached DB too IIUC.

src/test/java/org/sqlite/ConnectionTest.java
gotson: i would probably have added a check on the parameter being non-null though. Doesn't make any sense to try with `null`, as it has no semantic meaning.

mruddy: i can add that. i had a null check initially until i was testing and realized that null meant the "first" database to sqlite3, which is also always the "main" database as far as i can tell.

gotson: well it depends what is the meaning in the sqlite function, if it is a synonym for "main", or if it is something else

src/main/java/org/sqlite/core/DB.java
mruddy: > I am not in favor of that. JDBC always needs some interpretation anyway, depending on the underlying database.

what i'm advancing is letting the user choose the interpretation by telling the driver via a config option sent into the java.sql.Connection. so, there'd be a config option that java.sql.Connection.isReadOnly uses to determine if it should return true based on a) only the read-only status of the "main" database, or b) the read-only status of any database. i'm excluding the possibility of the result being determined by the status of all databases because when you add a temp database, or a memory database, i'm pretty sure that you'd be locking into never being able to return read-only.

i'm trying to support people that connect initially to a read-write main database and then attach to another read-only database. i can think of valid uses of this sequence of events.

src/test/java/org/sqlite/ConnectionTest.java
mruddy: if we pass in null, sqlite returns the read-only status of the database at index 0 -- the first database, which as far as i can tell always means the "main" database. in effect, i think it is a synonym for "main". that's why i was like, well it doesn't hurt anything to pass in null, and it's less code. it's sloppy/lazy on the caller's part to pass in null, and i prefer being explicit with the database name, so i can add the null check back. can i throw a NullPointerException, or IllegalArgumentException? :-)

src/main/java/org/sqlite/core/DB.java
gotson: i don't think we need that kind of knob tbh

src/test/java/org/sqlite/ConnectionTest.java
gotson: i am ok with passing null if the underlying sqlite function accepts it and has a meaning for it. We need to put that in the function's comment though, so we we know why we didn't guard about null in the first place

mruddy: ah, ok, i made that comment on `is_read_only`, i can copy the description to the other method's comment too

mruddy:
that PR was about transactions and read only mode. for example, the funkiness in `JDBC3Connection.tryEnforceTransactionMode` is what pushed me to start working on this whole series of updates.

src/main/java/org/sqlite/core/DB.java
mruddy: i was thinking of a scenario where someone had a read+write main database. then they attach to a read-only database in the same connection that contains info that they use to make updates to the main database. it's possible that people have existing code doing that, and it wouldn't be crazy. to open the attached database in read only mode, i believe all that has to be done is to make the file read-only with file permissions (or maybe even just mount the filesystem containing the second database as read-only).

so, the existing code would consider the connection read-write, but if we start considering the connection read-only if any one database is read-only, then it would then become read-only, which would be a breaking change. therefore, i think we'd default to the old way of considering the connection read-only depending on only the status of the main database. then we could give people the knob that allows them to tell the SQLiteConnection.isReadOnly implementation whether to consider the connection read-only if any of the additional databases are read-only.

mruddy: just pushed up a commit that returns the database status as an enum to match the native call.

src/test/java/org/sqlite/ConnectionTest.java
mruddy: updated with a new commit

src/test/java/org/sqlite/ConnectionTest.java
mruddy: handled by new commit

mruddy:
not sure if this pr has any traction, so to not waste a lot of time, i put together a rough draft commit dcc31b3a2c351019809109f719bcd909eae05a56 to demonstrate using the new functionality to redo read only and transaction mode handling.

gotson:
> not sure if this pr has any traction

to be frank i don't understand anything you are doing in this PR. It started as exposing `sqlite3_db_readonly`, and now you are ripping off the whole read only transaction feature, which as i mentioned in my previous comment is not related to the database being read only.

Unless you manage to clarify what you are trying to achieve, i will reject this PR.

mruddy:
ok, so #432 and #782 tried to solve the problem of not being able to call `setReadOnly` to change the read only status of a connection object.
one of the insights that those changes brought was the fact that the `query_only` pragma could be used.
i'm building on this insight by making the read only determination more comprehensive via use of the new native function.
without my changes, there are overlooked cases where created connections wind up being read only even though the user did not specify to open the connection as read-only (for example, when the file has only read file system permissions, or is on a filesystem that is mounted read only).

#432 and #782 also introduced the insight that *transactions* that upgrade from read-only to write mode are problematic for ORM tools because of `SQLITE_BUSY` errors. the insight is that if you're going to do only reads, then deferred mode is preferable because you maximize concurrency. if you're going to do writes, then immediate mode is preferable because you either wait (and fail on busy timeout) to start a transaction and are then guaranteed that you won't get `SQLITE_BUSY` errors, or fail right up front. i agree with using immediate mode transactions when there will be writes, but it also has concurrency effects to keep in mind. this is where read-only and transaction modes become intertwined. they really are very closely related in sqlite. but also, there's another transaction mode, exclusive, so we should still allow users to have control and the ability to use it.

i'm not trying to do away with those insights, i'm trying to implement them in a more transaction friendly and api compliant manner.
the implementation of `tryEnforceTransactionMode` is clunky and has some sharp edges. the idea of changing transaction modes after a transaction has started is funky to me. the jdbc api is structured to handle that before starting a transaction.

i'm proposing that we can make the read-only logic more comprehensive, catch more problems early, and also integrate the transaction mode handling in a way that fits better with the jdbc api spec.

i was trying to do this via a series of commits that would incrementally transform, and be understandable in their own right, so as to not overwhelm with a large PR that has a ton of changes.

the commits that i have pushed up so far are not quite complete yet because i wanted to get feedback on if the user configurable options should be implemented (look in the changes for TODO comments).

i think that the configuration options are useful for different users. some users will use the jdbc+driver api's directly and some will use an orm tool. the configuration options are likely more useful for the orm tool users.

gotson:
the 2 PRs you mention have nothing to do with the database or connection being read only, they only address the need for the ORM to avoid SQLITE_BUSY.

Sqlite works differently than other databases, and doesn't always fit in the JDBC APIs.

mruddy:
umm, yes they both very much do. one example from each can be found in method `checkTransactionMode` for #432 and `tryEnforceTransactionMode` for #782. or just search the diff for `query_only`. or you could search the diff for `readOnly`.
it seems like we might not be getting anywhere here.

mruddy:
closing for now. if anybody needs these changes to workaround bugs, you can grab them in the diff. i may also begin to maintain a public fork, but am not doing so right now.

PR #856, opened 2023-03-14
feat: remove length/ dimension of column data types in the `TYPE_NAME` field of `getColumns()`

sualeh:
If columns are defined with `intvalue integer(5)`, `realvalue real(8,3)`, or `charvalue varchar(21)` for example, return `INTEGER`, `REAL` or `VARCHAR` respectively. This is consistent with how other JDBC drivers return the `TYPE_NAME` field in [`DatabaseMetaData.getColumns()`](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-).

Tests are in https://github.com/xerial/sqlite-jdbc/blob/master/src/test/java/org/sqlite/DBMetaDataTest.java#L271

### DIFF
--- a/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
@@ -842,7 +842,6 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
     protected static final Pattern TYPE_INTEGER = Pattern.compile(".*(INT|BOOL).*");
     protected static final Pattern TYPE_VARCHAR = Pattern.compile(".*(CHAR|CLOB|TEXT|BLOB).*");
     protected static final Pattern TYPE_FLOAT = Pattern.compile(".*(REAL|FLOA|DOUB|DEC|NUM).*");
-    protected static final Pattern TYPE_WITH_SIZE = Pattern.compile("^(.*)\\(\\s*\\d+\\s*(\\s*,\\s*\\d+\\s*)?\\)$");
 
 
     /**
@@ -1072,12 +1071,8 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
                                     // just ignore invalid dimension formats here
                                 }
                             }
-                        }
-
-                        // remove (optional) length/ dimension of the column before returning result set
-                        Matcher regexMatcher = TYPE_WITH_SIZE.matcher(colType);
-                        if (regexMatcher.find()) {
-                            colType = regexMatcher.replaceFirst("$1").trim();
+                            // "TYPE_NAME" (colType) is without the length/ dimension
+                            colType = colType.substring(0, iStartOfDimension).trim();
                         }
 
                         int colGenerated = "2".equals(colHidden) ? 1 : 0;
--- a/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
@@ -842,6 +842,8 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
     protected static final Pattern TYPE_INTEGER = Pattern.compile(".*(INT|BOOL).*");
     protected static final Pattern TYPE_VARCHAR = Pattern.compile(".*(CHAR|CLOB|TEXT|BLOB).*");
     protected static final Pattern TYPE_FLOAT = Pattern.compile(".*(REAL|FLOA|DOUB|DEC|NUM).*");
+    protected static final Pattern TYPE_WITH_SIZE = Pattern.compile("^(.*)\\(\\s*\\d+\\s*(\\s*,\\s*\\d+\\s*)?\\)$");
+
 
     /**
      * @see java.sql.DatabaseMetaData#getColumns(java.lang.String, java.lang.String,
@@ -1070,6 +1072,12 @@ public abstract class JDBC3DatabaseMetaData extends org.sqlite.core.CoreDatabase
                                     // just ignore invalid dimension formats here
                                 }
                             }
+                        }
+
+                        // remove (optional) length/ dimension of the column before returning result set
+                        Matcher regexMatcher = TYPE_WITH_SIZE.matcher(colType);
+                        if (regexMatcher.find()) {
+                            colType = regexMatcher.replaceFirst("$1").trim();
                         }
 
                         int colGenerated = "2".equals(colHidden) ? 1 : 0;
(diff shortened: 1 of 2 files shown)


gotson:
> This is consistent with how other JDBC drivers return the `TYPE_NAME` field in [`DatabaseMetaData.getColumns()`](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-).

Can you clarify?

sualeh:
> This is consistent with how other JDBC drivers return the TYPE_NAME field 

> Can you clarify?

@gotson The `TYPE_NAME` field in [`DatabaseMetaData.getColumns()`](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-) should return the type name only, without the length/ dimension of the column data types. This information is returned in other fields such as `COLUMN_SIZE` and `DECIMAL_DIGITS`. The Xerial SQLite JDBC Driver returns `COLUMN_SIZE` and `DECIMAL_DIGITS` correctly, but the `TYPE_NAME` field additionally includes these values in parentheses, which is not required.

gotson:
> > This is consistent with how other JDBC drivers return the TYPE_NAME field
> 
> > Can you clarify?
> 
> @gotson The `TYPE_NAME` field in [`DatabaseMetaData.getColumns()`](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-) should return the type name only, without the length/ dimension of the column data types. This information is returned in other fields such as `COLUMN_SIZE` and `DECIMAL_DIGITS`. The Xerial SQLite JDBC Driver returns `COLUMN_SIZE` and `DECIMAL_DIGITS` correctly, but the `TYPE_NAME` field additionally includes these values in parentheses, which is not required.

Can you supplement this with some documentation ? The Javadoc does not explain that, and does not provide any example.

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -1072,6 +1074,12 @@ public ResultSet getColumns(String c, String s, String tblNamePattern, String co
                             }
                         }
 
+                        // remove (optional) length/ dimension of the column before returning result set
+                        Matcher regexMatcher = TYPE_WITH_SIZE.matcher(colType);
+                        if (regexMatcher.find()) {
+                            colType = regexMatcher.replaceFirst("$1").trim();
```
gotson, line 1080: I'm a bit dubious as to why this syntax is used, why not just use `regexMatcher.group(1).trim()` ?

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -1072,6 +1074,12 @@ public ResultSet getColumns(String c, String s, String tblNamePattern, String co
                             }
                         }
 
+                        // remove (optional) length/ dimension of the column before returning result set
+                        Matcher regexMatcher = TYPE_WITH_SIZE.matcher(colType);
```
gotson, line 1078: can we use a more contextual variable name here?

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -842,6 +842,8 @@ public ResultSet getColumnPrivileges(String c, String s, String t, String colPat
     protected static final Pattern TYPE_INTEGER = Pattern.compile(".*(INT|BOOL).*");
     protected static final Pattern TYPE_VARCHAR = Pattern.compile(".*(CHAR|CLOB|TEXT|BLOB).*");
     protected static final Pattern TYPE_FLOAT = Pattern.compile(".*(REAL|FLOA|DOUB|DEC|NUM).*");
+    protected static final Pattern TYPE_WITH_SIZE = Pattern.compile("^(.*)\\(\\s*\\d+\\s*(\\s*,\\s*\\d+\\s*)?\\)$");
```
gotson, line 845: maybe using named capturing group would be clearer ? Adding a comment on what the regex does and why would be helpful too.

src/test/java/org/sqlite/DBMetaDataTest.java
```diff
@@ -35,7 +35,7 @@ public void connect() throws Exception {
         conn = DriverManager.getConnection("jdbc:sqlite:");
         stat = conn.createStatement();
         stat.executeUpdate(
-                "create table test (id integer primary key, fn float default 0.0, sn not null, intvalue integer(5), realvalue real(8,3));");
+                "create table test (id integer primary key, fn float default 0.0, sn not null, intvalue integer(5), realvalue real(8,3), charvalue varchar(21));");
```
gotson, line 38: what would the added column test compared to the existing `interger(5)` one, in the context of that change?

gotson:
> @gotson I do not have any documentation, but only practical experience. If you look at the https://github.com/schemacrawler/SchemaCrawler tests, they run against multiple databases and JDBC drivers (Oracle, Microsoft SQL Server, IBM DB2, MySQL, PostgreSQL, HyperSQL (HSQLDB), H2, DuckDB, Trino, Cassandra, Calcite, CockroachDB, Derby, Firebird, Informix, MariaDB), and the SQLite JDBC driver is the only outlier in this regard of reporting just the data-type name in the `TYPE_NAME` field in [`DatabaseMetaData.getColumns()`](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-), without any size information in parentheses.

Thanks for providing context, this is helpful.

Looking at the code we already have a code block to find the length/dimension, can't we reuse that to isolate the type, instead of adding one more regex?

src/test/java/org/sqlite/DBMetaDataTest.java
sualeh: @gotson True. It does not add a test case for this change. All it does is test something else, which is an explicit versus implicit (`sn not null`) VARCHAR.

gotson: No that's fine, more tests is good!

src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
sualeh: Please see updated code, which does not use regular expressions.

gotson:
Thanks @sualeh , i like how the code is much cleaner now !

PR #875, opened 2023-04-05
refactor

GIRISHARAN:


gotson:
Without any description or explanations I won't be accepting this

GIRISHARAN:
Refactoring techniques like 
Introduce explaining variable, Extract Method, Extract Class, Decompose Condition was applied for code re-usability and code readability

PR #876, opened 2023-04-07
Code Refactor

GIRISHARAN:
Refactoring techniques like
Introduce explaining variable, Extract Method, Extract Class, Decompose Condition, Pull down method was applied for code re-usability and code readability


GIRISHARAN:
Still  changed some code

PR #882, opened 2023-04-15
fix(jdbc): PreparedStatement getParameterType and getParameterTypeName

jgoday:
Implements JDBC3PreparedStatement getParameterType and getParameterTypeName.
It returns the JDBC type based on the parameter class.


gotson:
thanks a lot for your first PR!

PR #883, opened 2023-04-15
Update Workflow

austinarbor:


gotson:
@austinarbor what are you doing? That's the second PR you open and close in a day with no description. Please stop doing this.

austinarbor:
@gotson sorry, it is not on purpose. default GitHub behavior is to create PR against upstream repo instead of base repo of the PR and I forgot to switch it. There's no way I can see in the UI to disable this behavior. Will try to not let it happen again

Issue #911, opened 2023-05-30
NPE in CoreStatement.checkIndex when no parameter has been set client side


jcgay:
I'm facing an issue since version `v3.41.2.2`.

**Describe the bug**
NullPointerException for `org.sqlite.jdbc3.JDBC3PreparedStatement.getParameterType` when no parameter has been set yet.

**To Reproduce**
```
    @Test
    void getParameterTypeTest_when_no_parameter_set() throws SQLException {
        Connection conn = DriverManager.getConnection("jdbc:sqlite:");
        Statement stat = conn.createStatement();
        stat.executeUpdate("create table t_int(i INT)");

        try (PreparedStatement ps = conn.prepareStatement("INSERT INTO t_int VALUES(?)")) {
            assertThatThrownBy(() -> ps.getParameterMetaData().getParameterType(1))
                    .isInstanceOf(NullPointerException.class);
            assertThatThrownBy(() -> ps.getParameterMetaData().getParameterTypeName(1))
                    .isInstanceOf(NullPointerException.class);
        }
    }
```

**Expected behavior**
A SQLException should be thrown instead indicating that parameter type cannot be determined.

**Logs**
```
java.lang.NullPointerException
	at org.sqlite.core.CoreStatement.checkIndex(CoreStatement.java:145)
	at org.sqlite.jdbc3.JDBC3PreparedStatement.getParameterType(JDBC3PreparedStatement.java:165)
	at xxx.dao.stubbing.SqliteBinderConfiguration.getBinder(SqliteBinderConfiguration.java:25)
	at com.ninja_squad.dbsetup.operation.Insert.initializeBinders(Insert.java:286)
	at com.ninja_squad.dbsetup.operation.Insert.execute(Insert.java:210)
	at com.ninja_squad.dbsetup.operation.CompositeOperation.execute(CompositeOperation.java:92)
	at com.ninja_squad.dbsetup.DbSetup.launch(DbSetup.java:107)
```

**Environment (please complete the following information):**
 - OS: all
 - CPU architecture: x86_64, arm64
 - sqlite-jdbc version 3.41.2.2

**Additional context**
The bug appears in my application tests using http://dbsetup.ninja-squad.com/ when a binder try to guess the database types (http://dbsetup.ninja-squad.com/user-guide.html#data-formats).

[...]

PR #912, opened 2023-05-30
fix: Check parameters presence in getParameterType

jcgay:
Before this commit, the code was failing with NPE when getParameterType was called before any PreparedStatement#setX was done.

Fix for #911


gotson:
Thank you !

PR #939, opened 2023-07-20
fix: use dylib extension instead of jnilib for MacOS library names

kkriske:
copy from https://github.com/xerial/sqlite-jdbc/pull/922#issuecomment-1595180965:
The libraries for MacOS are packaged as a .jnilib file. This seems legacy from pre java 7 (https://bugs.openjdk.org/browse/JDK-7134701). The default naming now seems to be .dylib.


src/main/resources/org/sqlite/native/Windows/armv7/sqlitejdbc.dll
gotson, line 1: we probably don't need that one

src/main/resources/org/sqlite/native/Windows/x86/sqlitejdbc.dll
gotson, line 1: we probably don't need that one

src/main/resources/org/sqlite/native/Windows/x86_64/sqlitejdbc.dll
gotson, line 1: we probably don't need that one

src/main/resources/org/sqlite/native/Windows/aarch64/sqlitejdbc.dll
gotson, line 1: we probably don't need that one

src/main/resources/org/sqlite/native/Mac/aarch64/libsqlitejdbc.dylib
gotson, line 1: was that regenerated or just renamed ? file size has changed

src/main/resources/org/sqlite/native/Mac/x86_64/libsqlitejdbc.dylib
gotson, line 1: was that regenerated or just renamed ? file size has changed

gotson (changes requested):

src/main/resources/org/sqlite/native/Mac/aarch64/libsqlitejdbc.dylib
kkriske: I renamed them initially, and then ran the CI to validate if they are still created correctly.
I don't know why the file size is different.

src/main/resources/org/sqlite/native/Windows/armv7/sqlitejdbc.dll
kkriske: The change does nothing indeed, but windows is unable to create the exact same binary twice due to embedded metadata like creation timestamps. I can revert the changes to the dlls if you want.

gotson: > I can revert the changes to the dlls if you want.

that's what i was implying. This PR should not modify those files.

src/main/resources/org/sqlite/native/Mac/aarch64/libsqlitejdbc.dylib
gotson: can we keep the original file and just rename them in that case?

kkriske: I reverted the CI run, that should do it.

PR #975, opened 2023-09-13
De-duplicate dockcross scripts by sourcing common code

05nelsonm:
This PR de-duplicates dockcross scripts by moving all common code to a separate file and sourcing it from each script.


gotson:
Those are generated by dockcross images, everytime we will update the dockcross image they would be regenerated.

PR #985, opened 2023-10-08
fix(native-image): do not initialize logging framework at build-time

kkriske:
The logging framework (slf4j) should not be configured to be initialized at build-time:
* It is bad form to dictate native-image configuration for classes that are not our own. This can cause configuration conflicts if either the end-user or the maintainer of the external class have/add their own configuration.
* end-users often configure logging frameworks to either redirect to each other, or have custom formatting rules. Initializing the logging framework at build-time can cause this configuration to be missed and partially or completely ignored.

This PR makes sure the logging framework is again initialized at run-time, by:
* Using in-line declarations of single-use loggers: `org.sqlite.SQLiteJDBCLoader$VersionHolder`
* Creating a `LogHolder` subclass which holds the logger instance, if the parent class is configured to be initialized at build-time. The holder wrapper itself will be initialzed at run-time which delays the logger instantiation to its first use: `org.sqlite.jdbc3.JDBC3DatabaseMetaData$LogHolder` & `org.sqlite.util.OSInfo$LogHolder`
* Although not used in this PR, an alternative is to invert the initialization config of a class like `JDBC3DatabaseMetaData`. The parent class (with static logger instance) would be configured at run-time, and the build-time initialization part would be put in an internal wrapper class which is marked for build-time initialization. An example of where this is already set up like this: `org.sqlite.SQLiteJDBCLoader$VersionHolder`.

### DIFF
--- a/src/main/java/org/sqlite/SQLiteJDBCLoader.java
+++ b/src/main/java/org/sqlite/SQLiteJDBCLoader.java
@@ -398,7 +398,6 @@ public class SQLiteJDBCLoader {
      * executable, and we're eliminating the IO operations as well.
      */
     public static final class VersionHolder {
-        private static final Logger logger = LoggerFactory.getLogger(VersionHolder.class);
         private static final String VERSION;
 
         static {
@@ -420,7 +419,10 @@ public class SQLiteJDBCLoader {
                     version = version.trim().replaceAll("[^0-9\\.]", "");
                 }
             } catch (IOException e) {
-                logger.atError()
+                // inline creation of logger to avoid build-time initialization of the logging
+                // framework in native-image
+                LoggerFactory.getLogger(VersionHolder.class)
+                        .atError()
                         .setCause(e)
                         .log("Could not read version from file: {}", versionFile);
             }
--- a/src/main/java/org/sqlite/util/OSInfo.java
+++ b/src/main/java/org/sqlite/util/OSInfo.java
@@ -40,7 +40,6 @@ import org.slf4j.LoggerFactory;
  * @author leo
  */
 public class OSInfo {
-    private static final Logger logger = LoggerFactory.getLogger(OSInfo.class);
     protected static ProcessRunner processRunner = new ProcessRunner();
     private static final HashMap<String, String> archMapping = new HashMap<>();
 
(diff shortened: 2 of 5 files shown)


src/main/java/org/sqlite/SQLiteJDBCLoader.java
```diff
@@ -420,7 +419,10 @@ public static final class VersionHolder {
                     version = version.trim().replaceAll("[^0-9\\.]", "");
                 }
             } catch (IOException e) {
-                logger.atError()
+                // inline creation of logger to avoid build-time initialization of the logging
+                // framework in native-image
+                LoggerFactory.getLogger(VersionHolder.class)
```
kkriske, line 424: It is possible to reference the `SQLiteJDBCLoader#logger` field here.
A caveat of the logger usage here is that if it ever happens in a native-image build, you will end up with the same `Error: Classes that should be initialized at run time got initialized during image building:` instead of the configured log message.

I would argue this is not an issue, as it should never occur in a native-image build. If it happens, normal JVM execution and tests would've already failed anyway.

src/main/java/org/sqlite/util/OSInfo.java
```diff
@@ -161,7 +160,7 @@ static String getHardwareName() {
         try {
             return processRunner.runAndWaitFor("uname -m");
         } catch (Throwable e) {
-            logger.atError().setCause(e).log("Error while running uname -m");
+            LogHolder.logger.atError().setCause(e).log("Error while running uname -m");
```
gotson, line 163: i guess a problem would be that the log line would reference the wrong class name, since it is using a logger generated in another class.

gotson:
> * Creating a `LogHolder` subclass which holds the logger instance, if the parent class is configured to be initialized at build-time. The holder wrapper itself will be initialzed at run-time which delays the logger instantiation to its first use: `org.sqlite.jdbc3.JDBC3DatabaseMetaData$LogHolder` & `org.sqlite.util.OSInfo$LogHolder`

to be honest i don't understand why GraalVM does not complain about this, but complains about the initial code. Would you be able to explain ? It seems to be very similar fields, the LogHolder being static, like the Logger.

gotson:
thanks a lot for the explanation!

PR #993, opened 2023-10-16
fix: downgrade SLF4J to 1.7

### LINKED ISSUE #990: New SQLite 3.43.2.0 fails when logging exceptions if the logger framework is SLF4J 1.7
**Describe the bug**
I'm using SLF4J 1.7 and updated to the latest SQLite 3.43.2.0. When opening a connection, a `java.lang.NoSuchMethodError` exception is produced at `org.slf4j.Logger.atTrace()`

The possible reason is that SQLite 3.43.2.0 (21c77a481f6970622b99d712d36d0336d63fd554) uses the new SLF4J 2.0 fluent api (e.g. `org.slf4j.Logger.atTrace()`) that is not available in SLF4J 1.7

**To Reproduce**
Open a connection using SQLite 3.43.2.0 using SLF4J 1.7 as your logging framework: the exception is raised

**Expected behavior**
No exception should be raised

**Logs**
Log trace:
```
'org.slf4j.spi.LoggingEventBuilder org.slf4j.Logger.atTrace()'
java.lang.NoSuchMethodError: 'org.slf4j.spi.LoggingEventBuilder org.slf4j.Logger.atTrace()'
at org.sqlite.core.NativeDB.prepare(NativeDB.java:129)
at org.sqlite.core.DB.prepare(DB.java:264)
at org.sqlite.jdbc3.JDBC3Statement.lambda$execute$0(JDBC3Statement.java:53)
at org.sqlite.jdbc3.JDBC3Statement.withConnectionTimeout(JDBC3Statement.java:459)
at org.sqlite.jdbc3.JDBC3Statement.execute(JDBC3Statement.java:42)
at org.sqlite.SQLiteConfig.apply(SQLiteConfig.java:219)
at org.sqlite.SQLiteConnection.<init>(SQLiteConnection.java:70)
at org.sqlite.JDBC.createConnection(JDBC.java:106)
at org.sqlite.JDBC.connect(JDBC.java:79)
at java.sql/java.sql.DriverManager.getConnection(Unknown Source)
at java.sql/java.sql.DriverManager.getConnection(Unknown Source)
```

 - OS: windows 11 and ubuntu in Github Actions
 - CPU architecture: x86_64
 - sqlite-jdbc version 3.43.2.0
 - slf4j-api version 1.7.36

vietk:
Fixes #990

### DIFF
--- a/src/main/java/org/sqlite/core/NativeDB.java
+++ b/src/main/java/org/sqlite/core/NativeDB.java
@@ -90,11 +90,10 @@ public final class NativeDB extends DB {
     /** @see org.sqlite.core.DB#_exec(java.lang.String) */
     @Override
     public synchronized int _exec(String sql) throws SQLException {
-        logger.atTrace()
-                .setMessage("DriverManager [{}] [SQLite EXEC] {}")
-                .addArgument(() -> Thread.currentThread().getName())
-                .addArgument(sql)
-                .log();
+        if (logger.isTraceEnabled()) {
+            logger.trace(
+                    "DriverManager [{}] [SQLite EXEC] {}", Thread.currentThread().getName(), sql);
+        }
         return _exec_utf8(stringToUtf8ByteArray(sql));
     }
 
@@ -126,11 +125,10 @@ public final class NativeDB extends DB {
     /** @see org.sqlite.core.DB#prepare(java.lang.String) */
     @Override
     protected synchronized SafeStmtPtr prepare(String sql) throws SQLException {
-        logger.atTrace()
-                .setMessage("DriverManager [{}] [SQLite EXEC] {}")
-                .addArgument(() -> Thread.currentThread().getName())
-                .addArgument(sql)
-                .log();
+        if (logger.isTraceEnabled()) {
+            logger.trace(
+                    "DriverManager [{}] [SQLite EXEC] {}", Thread.currentThread().getName(), sql);
+        }
         return new SafeStmtPtr(this, prepare_utf8(stringToUtf8ByteArray(sql)));
     }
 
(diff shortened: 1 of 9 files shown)


src/main/java/org/sqlite/core/NativeDB.java
```diff
@@ -90,11 +90,7 @@ protected synchronized void _open(String file, int openFlags) throws SQLExceptio
     /** @see org.sqlite.core.DB#_exec(java.lang.String) */
     @Override
     public synchronized int _exec(String sql) throws SQLException {
-        logger.atTrace()
-                .setMessage("DriverManager [{}] [SQLite EXEC] {}")
-                .addArgument(() -> Thread.currentThread().getName())
-                .addArgument(sql)
-                .log();
+        logger.trace("DriverManager [{}] [SQLite EXEC] {}", Thread.currentThread().getName(), sql);
```
gotson, line 93: shouldn't we need to check the log level is trace with an `if` to avoid the performance of getting the thread name?

vietk: Ah thanks for the heads-up I haven't noticed the lambda, I will go through the PR and fix if there are other

vietk: This is fixed in PR update https://github.com/xerial/sqlite-jdbc/compare/e33f9c8441ffb77d722d8dcddbdc18866bd4588e..221b71a762f9db3a078e30d860c5120f7c74ddfe

gotson:
I have added a commit to update the SLF4J version in the 2 makefile, which is used to build the OSInfo class.

Thanks for the PR, i think it will make it easier for dependent project before they can migrate to SLF4J 2.

Issue #1002, opened 2023-10-24
Inaccurate floating point values since 3.43.0.0


InitFlo:
**Describe the bug**
Since 3.43.0.0 some precision of numbers is lost.

**To Reproduce**
```java
@Test
public void sqliteTest() throws SQLException {
	var testValue = new BigDecimal("3.14");

	try (var conn = DriverManager.getConnection("jdbc:sqlite::memory:")) {
		try (var createTblStmt = conn.prepareStatement("create table foo(nr number(10,2));")) {
			createTblStmt.execute();
		}
		try (var insStmt = conn.prepareStatement("insert into foo(nr) values(?);")) {
			insStmt.setBigDecimal(1, testValue);
			insStmt.execute();
		}

		try (var selStmt = conn.prepareStatement("select nr from foo;")) {
			if (selStmt.execute()) {
				try (var selResult = selStmt.getResultSet()) {
					var col1Value = selResult.getBigDecimal(1);
					assertEquals(testValue, col1Value);
				}
			}
		}
	}
}
```

**Expected behavior**
I expected 3.14 as the returned column value, but it is 3.1399999999999997.

**Environment (please complete the following information):**
 - OS: Windows 11
 - CPU architecture: x86_64
 - sqlite-jdbc version 3.43.2.1

gotson:
Maybe check https://github.com/xerial/sqlite-jdbc/issues/927

It's unlikely that it's coming from the driver. Can you try with sqlite cli and compare between versions using the cli? It's most probably coming from sqlite itself.

gotson:
Why is your table created with a number type and not real/double/float affinity if you want to store doubles in it?

gotson:
You could also add any slf4j implementation and enable trace logs to show what exact sql statements are used.

InitFlo:
The issue is not present in [sqlite-tools-win32-x86-3420000.zip ](https://sqlite.org/2023/sqlite-tools-win32-x86-3420000.zip) and [sqlite-tools-win32-x86-3430200.zip](https://sqlite.org/2023/sqlite-tools-win32-x86-3430200.zip).

Yes, there were changes in sqlite: https://www.sqlite.org/changes.html. Probably

> 6. Enhancements to the [decimal extension](https://www.sqlite.org/floatingpoint.html#decext)
c) If X is a floating-point value, then the decimal(X) function now does a full expansion of that value into its exact decimal equivalent.

is the cause. There is the behaviour of both cli-Versions different.

[...]

PR #1003, opened 2023-10-29
fix: avoid double precision in getBigDecimal

InitFlo:
This solves issue #1002.

### DIFF
--- a/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
@@ -160,7 +160,7 @@ public abstract class JDBC3ResultSet extends CoreResultSet {
             case SQLITE_INTEGER:
                 return BigDecimal.valueOf(safeGetLongCol(col));
             case SQLITE_FLOAT:
-            //avoid double precision
+                // avoid double precision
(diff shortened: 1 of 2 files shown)


gotson:
Care to explain the reasoning?

InitFlo:
If a column has the real type affinity and the value can be represented with the limited accuracy of a double, the function `safeGetColumnType(checkCol(col))` evalutes to `SQLITE_FLOAT`. Getting the value using `safeGetDoubleCol(col)` causes an implicit conversion to a double. Since version 3.43.0 the sqlite source doesn't round the value anymore, but returned the value including the 8-byte IEEE floating point number inaccuracies. Somehow different operating systems behave differently. This explains, why the test only faled on Windows.
Using `safeGetColumnText(col)` avoids this lossy conversion.

I debugged the java code. The C part depends on the compiler and platform architecture.

The javadoc of `java.sql.ResultSet#getBigDecimal(int)` demands full precision, which cannot bei ensured by 8-byte IEEE floating point.

gotson:
Thanks for the explanation, i pulled the changed into [my branch with more tests](https://github.com/gotson/sqlite-jdbc/actions/runs/6688113249), and indeed it works well.

PR #1005, opened 2023-10-30
feat(sqlite): support serialize/deserialize sqlite APis

yuvalp-k2view:


yuvalp-k2view:
I see some warning about using DirectByteBuffer and recommendations to use byte[]. Reading up on it ... a fix is coming

src/main/java/org/sqlite/core/NativeDB.java
```diff
@@ -559,4 +559,24 @@ long getUpdateListener() {
...
         return progressHandler;
     }
+
+    @Override
+    public native synchronized long serializeSize(String schema);
+
+    @Override
+    public synchronized ByteBuffer serialize(String schema) {
+        long size = serializeSize(schema);
+        if (size > Integer.MAX_VALUE) {
```
gotson, line 569: why is that a problem?

src/test/java/org/sqlite/SerializeTest.java
```diff
@@ -0,0 +1,176 @@
...
+        ByteBuffer bb = serialize();
+        assertThat(bb.isDirect()).isTrue();
+        byte[] buff = new byte[bb.remaining()];
+        bb.get(buff);
+        try {
+            deserializeAndAssert(ByteBuffer.wrap(buff));
+            fail("exception expected");
+        } catch (IllegalArgumentException ignore) {
+            // has to be a direct buffer
+        }
```
gotson, line 48: probably better to use `assertThatThrownBy`

src/test/java/org/sqlite/SerializeTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+        try (SQLiteConnection connection = (SQLiteConnection) DriverManager.getConnection("jdbc:sqlite:")) {
+            execute(connection, "ATTACH ? AS ?", ":memory:", "a_schema");
+            connection.deserialize("a_schema", bb);
+            assertThat(fetch(connection, "SELECT count(*) FROM a_schema.a_table")).isEqualTo(10001);
+        }
+    }
+
+    @Test
+    public void testMultiFirstTimeSerialize() throws SQLException {
```
gotson, line 82: what does that test ? there's no assertions

src/test/java/org/sqlite/SerializeTest.java
```diff
@@ -0,0 +1,176 @@
...
+        ByteBuffer bb = ByteBuffer.allocateDirect(3);
+        bb.put((byte)1);
+        bb.put((byte)2);
+        bb.put((byte)3);
+        try {
+            deserializeAndAssert(bb);
+            fail("exception expected");
+        } catch (SQLException e) {
+            assertThat(e.getErrorCode()).isEqualTo(SQLiteErrorCode.SQLITE_NOTADB.code);
+        }
```
gotson, line 99: should use `assertThatThrownBy` for clarity

gotson (changes requested):

src/main/java/org/sqlite/core/NativeDB.java
yuvalp-k2view: ByteBuffer is limited to max int

gotson: can we use something else in that case ? that seems very small

yuvalp-k2view: It's all meant for in memory databases so not that small.
We could use an array of buffers. That would not be a very elegant API though.
I could wrap it with an InputStream so the end user will access it with a more standard approach. Could support both APIs. buffer for small databases and InputStream for larger?
wdyt?

gotson: > It's all meant for in memory databases so not that small.

Not sure to understand this statement. This can be used to serialize any database, not just in-memory, no? Or did i misunderstood something?

>  We could use an array of buffers. That would not be a very elegant API though.

do we have to use a Buffer ? Why not just a byte array?

yuvalp-k2view: Theoretically you could Serialize a file into memory but that would be very inefficient. Makes much more sense to just read the file. Deserialize is by definition memory only.
The max sqlite default for this mode is only 1,073,741,824 (1gb).
I am checking if I can increase it.
There are good reasons to switch to byte[] (might be faster oveall, see comment below). However, it is also limited to max int size.

yuvalp-k2view: More info here: I've managed to raise the limit but anything beyond 2gb results in a malloc error from sqlite.
Turns out it does want to allocate more than 2gb of continuous space. I will ask the sqlite folks if there is a way around it but for now I think we should leave it at that.

src/test/java/org/sqlite/SerializeTest.java
yuvalp-k2view: done

src/test/java/org/sqlite/SerializeTest.java
yuvalp-k2view: Added assert that size doesn't change across runs

src/test/java/org/sqlite/SerializeTest.java
yuvalp-k2view: Done

gotson:
it seems there are some CI errors, can you have a look?

src/main/java/org/sqlite/core/NativeDB.java
yuvalp-k2view: Asked and answered by sqlite maintainer. 2gb is a hard coded limit in the library, no way to change it, not even with a compilation flag.

yuvalp-k2view:
> it seems there are some CI errors, can you have a look?

ran spotless:apply o that should be okay.

The unsatisfied link error ... I'm baffled by. I was able to link and run all the tests locally without a problem and of course the native code contains the Java_org_sqlite_core_NativeDB_serializeSize function.

Any ideas? Anything I can do locally to reproduce?

yuvalp-k2view:
It seems the "make" artifacts are committed... didn't occur to me. Running a clean 'make' and will commit them.

yuvalp-k2view:
Binaries comitted

src/main/java/org/sqlite/core/NativeDB.java
gotson: might be worth to leave a comment in the code about it

gotson:
yes, binaries are committed to the repo, and can be built [using a PR comment](https://github.com/xerial/sqlite-jdbc/blob/master/CONTRIBUTING.md#build-from-ci) too.

gotson:
@pyckle may i ask whether you would have a bit of time to glance over the C code of that PR, and provide your insights ?

src/main/java/org/sqlite/core/NativeDB.java
yuvalp-k2view: Done. Also added the limit to the javadoc comment.

yuvalp-k2view:
@pyckle @gotson 
I see failures only on the graalvm tests. Is there anything I should do to create artifacts or test this?

src/main/java/org/sqlite/core/NativeDB.java
pyckle: I believe this limit is slightly wrong.
See: https://stackoverflow.com/questions/3038392/do-java-arrays-have-a-maximum-size
See: https://github.com/OpenHFT/Chronicle-Bytes/blob/a02a79ed8ce9900f8748ad6f3ce801fa241844d1/src/main/java/net/openhft/chronicle/bytes/Bytes.java#L110

I would go with the more conservative limit that openhft suggests. The real limit of a byte array size is JVM dependent. In addition, I don't think this is reachable anyways due to the sqlite malloc limit.

pyckle:
> @pyckle may i ask whether you would have a bit of time to glance over the C code of that PR, and provide your insights ?

I took a quick look. Looks good to me.

> I see failures only on the graalvm tests.

Most likely this is graal native image failures because https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java needs to be updated to list the newly added JNI calls

src/main/java/org/sqlite/core/NativeDB.java
```diff
@@ -559,4 +559,24 @@ long getUpdateListener() {
...
+    public synchronized native long serializeSize(String schema) throws SQLException;
+
+    @Override
+    public synchronized byte[] serialize(String schema) throws SQLException {
+        long size = serializeSize(schema);
+        if (size > Integer.MAX_VALUE) {
+            // Extra precaution as the sqlite side is also limited to 2gb.
+            throw new IllegalStateException("Database is bigger than byte[] max size: " + size);
+        }
+        byte[] buff = new byte[(int) size];
```
pyckle, line 573: It seems to me that it would be better to allocate this array in JNI and return it, and call serialize once.
Even though this DB can't be manipulated here due to the synchronize, there's nothing stopping another connection (or even another process) from manipulating the DB between these calls.

I don't think this is critical, as the JNI code in serialize double checks this and as you said, this is really more for inmemory databases. But it seems minor to fix and would make this more robust.

src/main/java/org/sqlite/core/NativeDB.java
yuvalp-k2view: @pyckle I think every module should take care of its own limitations. It is true that the limitation on the sqlite is a bit more lower than the java side so this shouldn't happen (added a comment to that effect).

I did develop a unit test that shows the database can grow up to ~2gb. It of course needs at least 4gb of free memory and takes around 30 seconds. Do you think I should commit it? I thought it was a bit extreme and could fail on some test environments.

yuvalp-k2view:
@pyckle I looked at https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
It seems to only list callback methods and not jni calls.
I did add byte[] line as it seems that any type of class to make the bridge is listed there.
What am I missing? Worth a rerun?

gotson:
> @pyckle I looked at [`master`/src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java](https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java?rgh-link-date=2023-11-02T13%3A48%3A17Z) It seems to only list callback methods and not jni calls. I did add byte[] line as it seems that any type of class to make the bridge is listed there. What am I missing? Worth a rerun?

@kkriske would you be able to have a quick look at this PR, see if we need to adjust anything for the GraalVM native configuration?

src/main/java/org/sqlite/core/NativeDB.java
gotson: > I did develop a unit test that shows the database can grow up to ~2gb. It of course needs at least 4gb of free memory and takes around 30 seconds. Do you think I should commit it? I thought it was a bit extreme and could fail on some test environments.

the github workers only have 7gb of ram and 15gb of disk space (total, not available)

src/main/java/org/sqlite/core/NativeDB.java
pyckle: > I found no way to this other than allocating DirectByteBuffer in jni
Why not use `NewByteArray()`? It should yield the same performance as allocating the byte array in java. See: `https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#New_PrimitiveType_Array_routines`
You can change the native function to return `byte[]`.

For InMemory DBs, concurrent modification obviously isn't an issue, but this may eventually be used for file dbs.

src/main/java/org/sqlite/core/NativeDB.java
pyckle: This is up to @gotson as he's the maintainer, but I would suggest committing it with an `@Ignore` annotation. That way we won't have stability issues on CI and if issues with large DB size arise in the future, one can run this test locally.

pyckle:
> @pyckle I looked at https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java It seems to only list callback methods and not jni calls. I did add byte[] line as it seems that any type of class to make the bridge is listed there. What am I missing? Worth a rerun?

I don't have time to look at this in detail hands on until later next week. If it's still broken then, I'll investigate.

src/main/java/org/sqlite/core/NativeDB.java
yuvalp-k2view: Done as you suggested, testVeryLarge is committed and Disabled

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
```diff
@@ -124,6 +124,7 @@ private void registerJNICalls() {
 
         // Other JNI calls
         RuntimeJNIAccess.register(boolean[].class);
+        RuntimeJNIAccess.register(byte[].class);
```
kkriske, line 127:
```suggestion
```

src/test/java/org/sqlite/SerializeTest.java
```diff
@@ -0,0 +1,245 @@
+package org.sqlite;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.sql.*;
+import org.junit.jupiter.api.Disabled;
```
kkriske, line 10:
```suggestion
import java.sql.*;
import org.assertj.core.api.InstanceOfAssertFactories;
import org.junit.jupiter.api.Disabled;
```

src/test/java/org/sqlite/SerializeTest.java
```diff
@@ -0,0 +1,245 @@
...
+            size = b.length;
+        }
+    }
+
+    @Test
+    public void testErrorCorrupt() {
+        byte[] bb = {1, 2, 3};
+        assertThatThrownBy(() -> deserializeAndAssert(bb))
+                .isInstanceOf(SQLException.class)
+                .hasFieldOrPropertyWithValue("errorCode", SQLiteErrorCode.SQLITE_NOTADB.code);
```
kkriske, line 86:
```suggestion
                .asInstanceOf(InstanceOfAssertFactories.type(SQLException.class))
                .extracting(SQLException::getErrorCode)
                .isEqualTo(SQLiteErrorCode.SQLITE_NOTADB.code);
```

src/test/java/org/sqlite/SerializeTest.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    @Test
+    public void testErrorNoSuchSchema() throws SQLException {
+        byte[] bb = serialize();
+        try (SQLiteConnection connection =
+                (SQLiteConnection) DriverManager.getConnection("jdbc:sqlite:")) {
+            assertThatThrownBy(() -> connection.deserialize("a_schema", bb))
+                    .isInstanceOf(SQLiteException.class)
+                    .hasFieldOrPropertyWithValue("resultCode", SQLiteErrorCode.SQLITE_ERROR);
```
kkriske, line 117:
```suggestion
                    .asInstanceOf(InstanceOfAssertFactories.type(SQLiteException.class))
                    .extracting(SQLiteException::getResultCode)
                    .isEqualTo(SQLiteErrorCode.SQLITE_ERROR);
```

src/test/java/org/sqlite/SerializeTest.java
```diff
@@ -0,0 +1,245 @@
...
+                                //noinspection InfiniteLoopStatement
+                                while (true) {
+                                    execute(
+                                            connection,
+                                            "INSERT INTO a_schema.a_table (x) values (?)",
+                                            s);
+                                }
+                            })
+                    .isInstanceOf(SQLException.class)
+                    .hasFieldOrPropertyWithValue("errorCode", SQLiteErrorCode.SQLITE_FULL.code);
```
kkriske, line 197:
```suggestion
                    .asInstanceOf(InstanceOfAssertFactories.type(SQLException.class))
                    .extracting(SQLException::getErrorCode)
                    .isEqualTo(SQLiteErrorCode.SQLITE_FULL.code);
```

kkriske (changes requested):
I don't have write rights so couldn't add this as a commit. I've added suggestions where needed.

The problem is purely in tests, `hasFieldOrPropertyWithValue` uses reflective lookups to the relevant field, so this would require native-image configuration. Either way, using such reflection is bad form imo as you're testing hidden internals and not exposed API. The suggestions remove the `hasFieldOrPropertyWithValue` calls and replace them with proper casts and method calls.

src/main/java9/org/sqlite/nativeimage/SqliteJdbcFeature.java
yuvalp-k2view: removed

src/test/java/org/sqlite/SerializeTest.java
yuvalp-k2view: did it a bit different

src/test/java/org/sqlite/SerializeTest.java
yuvalp-k2view: did it a bit different

src/main/java/org/sqlite/core/NativeDB.java
yuvalp-k2view: Done as you suggested, wasn't aware. Benchmarked, all is well. I also removed the serializedSize method. Don't need it + worried that someone will use it to find out the db size which is very wasteful compared to querying sqlite.

src/test/java/org/sqlite/SerializeTest.java
kkriske: That's fine, one minor note: All the `isInstanceOf` checks should probably check against `SQLiteException` and not `SQLException` to be in line with the casts to `SQLiteException`.

yuvalp-k2view:
I've removed reflection from the tests. We can try to run again.

src/test/java/org/sqlite/SerializeTest.java
yuvalp-k2view: I started of testing against the contract, but you are right, if I cast, I should check it correctly. done.

gotson:
@yuvalp-k2view could you rebase on the latest `master` ? I have updated sqlite to 3.44.0.

Once done we can rebuild the native libs [with a PR comment](https://github.com/xerial/sqlite-jdbc/blob/master/CONTRIBUTING.md#build-from-ci)

yuvalp-k2view:
I rebased but did not rebuild. I am assuming you will override the binaries

gotson:
it seems the bot can't push to your remote https://github.com/xerial/sqlite-jdbc/actions/runs/6768202925/job/18392278956?pr=1005

yuvalp-k2view:
I rebuilt and pushed

yuvalp-k2view:
Hi @gotson, are there any outstanding issues? Would love to see this feature as part of the 3.44 release.

 (3.44 removes a global lock when creating transaction for in-memory databases. combination of this and serialize is a very serious performance boost).

src/main/java/org/sqlite/core/NativeDB.java
pyckle: Looks good to me. Nice work (:

gotson (approved):
LGTM

PR #1027, opened 2023-11-28
Make ConnectionConfig#autoCommit configurable via fake pragma

exoego:
## Problem

I am using Kotlin's SQL Framework https://github.com/JetBrains/Exposed with sqlite-jdbc.
```kotlin
import org.jetbrains.exposed.sql.Database
import org.jetbrains.exposed.sql.transactions.transaction
(snip)

val sqLiteConfig = SQLiteConfig()
val ds = SQLiteDataSource(sqLiteConfig)
val database = Database.connect(ds)

import org.jetbrains.exposed.sql.transactions.transaction
import org.sqlite.SQLiteErrorCode
import org.sqlite.SQLiteException
import java.time.temporal.Temporal

transaction(db) {
  connection.autoCommit = false
  // do something with DB
}
```

I want to disable auto-commit for better performance, but it appears that it is not configurable via SQLiteConfig nor SQLiteDataSource.
SQLiteConnectionConfig offers `setAutoCommit`, but `SQLiteConfig#defaultConnectionConfig` always use `autoCommit = true`.
https://github.com/xerial/sqlite-jdbc/blob/263c296f0057b1de9d758fbdef5ccfd2762add26/src/main/java/org/sqlite/SQLiteConnectionConfig.java#L21

One can specify `connection.autoCommit = false` in `transaction { ... }` DSL, but it is error-prone because one might forget.

## Proposal

This PR adds "fake" pragma `jdbc.auto_commit`, so auto-commit can be configurable via SQLiteConfig.

```kotlin
val prop = new Properties()
prop.setProperty(SQLiteConfig.Pragma.JDBC_AUTO_COMMIT.pragmaName, "true")
val sqLiteConfig = SQLiteConfig(prop)
val ds = SQLiteDataSource(sqLiteConfig)
val database = Database.connect(ds)
```

### DIFF
--- a/src/main/java/org/sqlite/SQLiteConfig.java
+++ b/src/main/java/org/sqlite/SQLiteConfig.java
@@ -535,7 +535,9 @@ public class SQLiteConfig {
 
         // extensions: "fake" pragmas to allow conformance with JDBC
         JDBC_AUTO_COMMIT(
-                "jdbc.auto_commit", "Configure explicit read only transactions", new String[] {"true", "false"}),
+                "jdbc.auto_commit",
+                "Configure explicit read only transactions",
+                new String[] {"true", "false"}),
         JDBC_EXPLICIT_READONLY(
                 "jdbc.explicit_readonly", "Set explicit read only transactions", null);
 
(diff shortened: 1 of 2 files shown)


gotson:
It seems JDBC expects the default to always be true, according to https://stackoverflow.com/a/11022406

It would probably make sense for the exposed framework to allow for such option so that new connections are always set to false.

exoego:
I agree autoCommit = true should be the default.
This PR allows users to configure autoCommit so they can use this driver for any framework in `autoCommit = false` mode.

exoego:
Please take a look.

Issue #1018, opened 2023-11-13
Simply removing JDBC generated key support is NOT a fix


barrycaceres:
Version 3.43.0.0 and later breaks all support for obtaining generated keys because it was not "safe" in some cases.
This is NOT a fix.  This is crippling SQLite access via JDBC.

This is supposedly a resolution to Issue #329 , but removing a feature does not resolve the problem.

We still need a reliable way to obtain the generated ID after a single INSERT statement.

There is discussion of using a "RETURNING" clause but even this says its not a complete solution.

The previous functionality worked JUST FINE if you were smart enough to obtain the generated ID after each individual insert and not let any other thread obtain access to your Connection object until you had atomically inserted and obtained the ID.

Just because this did not work with batch inserts or poorly crafted multi-threaded applications did not mean that the functionality should be removed.

Please restore support for obtaining generated keys via JDBC.

ADDITIONALLY: Removing this support in a minor release (i.e.: not 4.x but going from 3.42.x to 3.43.x) breaks backwards compatibility and therefore is not adhering to proper semantic versioning.

sjlombardo:
+1

 I fully understand that:
1. the previous behavior was not ideal, and that it broke down in more complicated scenarios featuring multi-row inserts, cross thread use, etc.
2. in a perfect world every application would check driver capabilities and gracefully handle that `getGeneratedKeys` is not available according to the JDBC spec

On the other hand, the sqlite-jdbc project supported (if not perfectly) `getGeneratedKeys` in for **_at least_** 16 years. It was a long term, well established, and generally understood feature of the software with a clear precedent of support. 

The abrupt removal of this capability has broken integrations with everything from Spring to Hibernate and has clearly caused problems for dozens of applications, plus more that we don't know about. I know first hand that it has caused projects to completely stop including upstream versions of sqlite-jdbc because it will irreparably break applications.

It would really be infinitely better to restore the old `getGeneratedKeys` behavior and document the limitations. The project could then pursue a long-term investigation for a way to improve the behavior of `getGeneratedKeys` without breaking backwards compatibility.

barrycaceres:
In my searches, I thought I read that Microsoft/Mojang's Java-based Minecraft was also broken by this update.

Clearly not something that should have happened in version `3.43.0.0` -- definitely should have been saved for version `4.x`.

In fact the comments in Issue #329 say that the functionality should be resolved in "the next **major** version"

[...]

Issue #468, opened 2019-12-16
getGeneratedKeys() only returns one ID on multiple insertions


alesimula:
If i make multiple insertions in one query

`INSERT INTO 'mytable' ('id', 'name') VALUES
  ('1', 'alex'),
  ('2', 'jean');`

The ResultSet of getGeneratedKeys() will only hold the ID of the last entry

gotson:
I think that's the default behaviour of SQLite to return a single value. You can check about [RETURNING](https://www.sqlite.org/lang_returning.html).

gotson:
Closing this in favor of #329

Issue #329, opened 2018-03-02
Proper / secure way to receive last_insert_rowid


b3nn0:
Hi,
right now there are multiple ways to retrieve the last inserted rowid of an insert.
None of them is very good.
- Query for "SELECT last_insertrowid()":
-> Doesn't work in multithreaded applications

- Add the select statement directly behind the INSERT query, e.g.
"INSERT INTO a VALUES('x'); SELECT last_insert_rowid()"
-> Doesn't work when the insert is a prepared statement. SQLite will only respect the first statement.

- Call stmt.getGeneratedKeys()
-> Just as unreliable. It seems to do a lazy call to last_insert_rowid(), which is, in my opinion, a bug.

As SqliteJdbc seems to synchronize everything going to SQLite, it would be quite easy to fix.
To do so, the library simply would need to call sqlite3_last_insert_rowid() right after the query.
JDBC already provides the API, see Statement.RETURN_GENERATED_KEYS constant.

Implementing it should be quite simple...
I could potentially provide a patch, if it has the chance to be accepted.

gotson:
Is this still happening on the latest version?

gotson:
I've been thinking about this, and I think option 2 would require to have a sql parser. We would need to be able to find out whether there is already a RETURNING clause, and also to find out what's the table name so that we can infer the generated keys too.

gotson:
@b3nn0 in the next major version the driver will not support anymore `getGeneratedKeys`, as there is no reliable way to retrieve those. The good way would be to use the SQLite `RETURNING` clause.

b3nn0:
Alright, makes sense. Closing this then.

sproket:
How is removing a feature resolved? Anyway it breaks Persism but I'm doing it the new way for SQLite and it's working so I'll put out an update soon.

barrycaceres:
If I am reading this correctly, the ability to obtain the generated keys has been removed entirely with no "proper" way to do it?

Those of us who were managing our connections and atomically doing single inserts and then immediately obtaining the generated keys were having no problem with the existing functionality because we understood that using it meant doing a single insert and before letting any other thread get your `Connection` object you had to get the generated key.

Simply removing this functionality because it did not work for those who did not understand how to do it safely seems like a non-fix but like crippling access to SQLite via JDBC.

barrycaceres:
[...]

PR #1035, opened 2023-12-06
feat(jdbc): reintroduce improved support for Statement#getGeneratedKeys

sjlombardo:
Hello @gotson,

Over in https://github.com/xerial/sqlite-jdbc/issues/1018#issuecomment-1813516287 you mentioned that you'd be willing to look at a PR that restores getGeneratedKeys functionality if it could resolve the safety issues originally reported in #329. 

This PR seeks to do that by adding back support for Statement#getGeneratedKeys with a new implementation that tightly scopes the generated keys to the Statement object. This makes it safe to use in an interleaved manner, across threads, and with multiple Statements.

In the interest of transparency, this PR does not address tracking keys for multiple inserts on the same statement (#468). However, that functionality was never previously available. That it could potentially be added in future enhancements to this model if it became a requirement.

Restoring this functionality would be a huge benefit to the users of this JDBC driver. Many applications and integrations were adversely affected by the removal of the basic getGeneratedKeys functionality that they came to depend on it over the past 16 years. This PR should restore the core functionality with critical improvements to scope and safety.

Please let me know if you have any questions or suggestions for improvements. Thank you for your consideration.

### DIFF
--- a/src/main/java/org/sqlite/core/CoreStatement.java
+++ b/src/main/java/org/sqlite/core/CoreStatement.java
@@ -193,8 +193,7 @@ public abstract class CoreStatement implements Codes {
         // up a new result set without any contents by issuing a query with a false where condition
         if (generatedKeysRs == null) {
             generatedKeysStat = conn.createStatement();
-            generatedKeysRs =
-                    generatedKeysStat.executeQuery("SELECT 1 WHERE 1 = 2;");
+            generatedKeysRs = generatedKeysStat.executeQuery("SELECT 1 WHERE 1 = 2;");
         }
         return generatedKeysRs;
     }
--- a/src/main/java/org/sqlite/core/CoreStatement.java
+++ b/src/main/java/org/sqlite/core/CoreStatement.java
@@ -18,7 +18,6 @@ package org.sqlite.core;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
-
 import org.sqlite.SQLiteConnection;
 import org.sqlite.SQLiteConnectionConfig;
 import org.sqlite.jdbc3.JDBC3Connection;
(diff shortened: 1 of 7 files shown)


sjlombardo:
Hi @gotson - just checking in - do you have any thoughts about this PR?

gotson:
> Hi @gotson - just checking in - do you have any thoughts about this PR?

I didn't had time to look at it yet, but will do in January.

src/main/java/org/sqlite/core/CoreStatement.java
```diff
@@ -149,4 +153,49 @@ protected void checkIndex(int index) throws SQLException {
...
+     * @see java.sql.Statement#getGeneratedKeys()
+     */
+    public ResultSet getGeneratedKeys() throws SQLException {
+        // getGeneratedKeys is required to return an EmptyResult set if the statement
+        // did not generate any keys. Thus, if the generateKeysResultSet is NULL, spin
+        // up a new result set without any contents by issuing a query with a false where condition
+        if (generatedKeysRs == null) {
+            generatedKeysStat = conn.createStatement();
+            generatedKeysRs =
+                    generatedKeysStat.executeQuery("SELECT last_insert_rowid() WHERE 0 <> 0;");
```
gotson, line 197: Wouldn't that execute the `last_insert_rowid()` function, even if the condition is false ?

If we need an empty result set, we can do something simpler like `select 1 where 1 = 2;`

sjlombardo: This is a good suggestion, I've updated the empty result set branch accordingly.

gotson:
seems like code formatting failed, could you run `mvn spotless:apply` ?

sjlombardo:
all set, code formatting should be good now!

gotson:
Thanks a lot for the quality PR!

PR #1041, opened 2023-12-20
[experimental] Add native lib for Linux loongarch64

Panxuefeng-loongson:
This PR tries to add native lib for Linux loongarch64


gotson:
we don't integrate any compiled libraries. We only add support when the libraries can be recompiled on demand, usually using Docker.

If you need to use any unsupported architecture, you can compile yourself and [explicitly load that library](https://github.com/xerial/sqlite-jdbc/blob/master/USAGE.md#how-to-use-a-specific-native-library).

PR #1056, opened 2024-01-23
fix(jdbc): support CTE and REPLACE syntax with Statement#getGeneratedKeys

sjlombardo:
This pull request improves `getGeneratedKeys` to support SQL statements that:

1. have a [CTE](https://www.sqlite.org/lang_with.html) preceding the `INSERT` (e.g. `WITH cte AS (SELECT 1 AS a) INSERT INTO t1(a) SELECT a FROM cte`)
2. use the more esoteric [`REPLACE`](https://www.sqlite.org/lang_insert.html) keyword instead of `INSERT OR REPLACE` (which SQLite supports for MySQL compatibility)

The improved matching will catch these cases and ensure that `genGeneratedKeys` returns an appropriate result set for them.

The tests are also expanded to cover the greater range of statement varieties.

### DIFF
--- a/src/main/java/org/sqlite/core/CoreStatement.java
+++ b/src/main/java/org/sqlite/core/CoreStatement.java
@@ -18,6 +18,7 @@ package org.sqlite.core;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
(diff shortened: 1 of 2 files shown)


gotson:
thanks!

Issue #938, opened 2023-07-13
jdbc potential memory leak


mhanna:
**Describe the bug**
Using a yourkit to capture memory used by java application that uses thousands of prepared statements and non prepared statements to sqlite, i can see that the NativeDB's safeStmt set is much more than the statements is should track.. see screenshot

**To Reproduce**
Maybe run many prepare/non-prepare statements 

**Expected behavior**

**Logs**
![image](https://github.com/xerial/sqlite-jdbc/assets/4300364/fa95c1a1-5168-41c3-8c58-2755f62f4750)

**Environment (please complete the following information):**
 - OS: Centos
 - CPU architecture: x86_64]
 - sqlite-jdbc version [e.g. 3.39.2.0]

Do you see any issue here?

rhyswilliamsza:
I am experiencing a similar issue on Java 11.0.19, but not Java 11.0.16. Will confirm if it's the same issue soon.

rhyswilliamsza:
I am potentially experiencing a similar issue on Java 11.0.19, but not Java 11.0.16. Which version of Java are you running?

mhanna:
openjdk version "17" 2021-09-14 OpenJDK Runtime Environment (build 17+35-2724) OpenJDK 64-Bit Server VM (build 17+35-2724, mixed mode, sharing)
________________________________

mhanna:
@gotson can anyone shed light on this?

gotson:
> @gotson can anyone shed light on this?

Someone would need to provide evidence of where the leak happens.

noahboegli:
@mhanna I had exactly the same problem as you. A program running for 6+ hours doing requests and requests to an SQLite db. 

The debian box has 6 GB of RAM and a variant of the program that does not make use of SQLite runs with less than 512MB.

After inspecting with VisualVM I found out that the instances of `SafeStmtPtr` were through the roof.

Simply closing the Statement calling `close()` solved my issue. I believe you should also be able to set them to automatically close on complete, although I haven't tested this.

I assume not closing the statement leaves some memory out of the JVM to remain allocated and when the `SafeStmtPtr` is collected the memory is leaked. Although this is a wild guess and I don't really know about JDBC's or SQLite's internals.

noahboegli:
On my end it was less practical to use a try-with-resources due to a catch and error forwarding nightmare, there was a `close()` call in the catch but not if the statement was successful. 

But the JavaDoc for Statement/Prepared statement is confusing

[...]

PR #1073, opened 2024-02-20
Docs: Adding try-with-resources to examples and demo. related #938

hbobenicio:
related #938.

- added try-with-resources to examples (README, USAGE and demo/Sample.java)
- changed exception handling to not ignore root causes
- removed finally blocks (not needed anymore)


gotson:
thanks a lot !

Issue #1049, opened 2024-01-09
Add info about which signing keys will be used for published artifacts


yogurtearl:
### Add info about which signing keys will be used for published artifacts. 

For security purposes, it would be great if you were able to publish details (in the project docs) about gpg public keys that are "valid" for use when verifying signing artifacts uploaded to maven central.

This allows for "out of band" verification of the expected signing key. 

Some examples of other libs publishing their signing keys:

https://square.github.io/okhttp/security/security/#verifying-artifacts

https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/KEYS.txt
https://downloads.apache.org/commons/KEYS
https://downloads.apache.org/logging/KEYS

gotson:
Up to @xerial since his key is being used in CI.

xerial:
I think this key (fingerprint) has been used for releasing sqlite-jdbc:
```
Taro L. Saito (For GitHub Actions) <leo@xerial.org> 05A7 776A A9B3 00D8 F5A0  966F 8857 595B 73BF D468
```

gotson:
@yogurtearl did you test the manual verification with the above key ?

gotson:
> yep, that is the key that is being used to sign the latest binaries, would be helpful to put it in the docs, on the website or in the README.

we would accept a PR in the readme, as we don't have a website.

prubel:
I ran into the same problem and just added https://github.com/xerial/sqlite-jdbc/pull/1076 with an update for the key.

PR #1076, opened 2024-03-04
docs: add gpg key to README

prubel:
https://github.com/xerial/sqlite-jdbc/issues/1049 requested a PR to add the signing key to the README.  This is that PR.


gotson:
Thanks !

PR #1080, opened 2024-03-12
perf: CoreStatement uses optimize regex for generated key matches

schlosna:
While profiling some software that makes heavy use of prepared statements and SQLite, I noticed that each call to `java.sql.Connection#prepareStatement(java.lang.String)` ended up compiling a regex pattern as part of `org.sqlite.core.CoreStatement.<init>(SQLiteConnection)` construction. This appears to be a recent addition as part of https://github.com/xerial/sqlite-jdbc/pull/1056  with example call trace showing up as over 75% of total allocations in JFR:

image

Because JDK `Pattern` instances are immutable and thread-safe, we can compile the regex on `CoreStatement` class load to a static field for reuse, and adjust the regex to avoid additional possible allocations from trimming and converting the SQL to lower case, as well as use non-capturing groups to avoid `Pattern#matcher` allocating the `int[]` arrays associated with capture groups.


gotson:
thanks a lot for the PR !

PR #1082, opened 2024-03-12
build: use BC signer

cstamas:
Changes:
* upgrade to 3.2.0 GPG plugin
* do NOT import key in home of runner (and leave it there once job done?)
* instead, make them both env variables
* and on CI force BC signer

This leaves GPG on workstation work unchanged.

Related to https://issues.apache.org/jira/browse/MGPG-90


gotson:
Thanks a lot @cstamas for the PR, let me merge this and see if the deploy works fine for snapshots!

PR #1088, opened 2024-03-20
Do not print the error message when failed to delete the old native lib.

fcyu:
There is no harm not deleting the old lib but the error message is quite annoying.

Thanks,

Fengchao

### DIFF
--- a/src/main/java/org/sqlite/SQLiteJDBCLoader.java
+++ b/src/main/java/org/sqlite/SQLiteJDBCLoader.java
@@ -99,9 +99,8 @@ public class SQLiteJDBCLoader {
                                 Path lckFile = Paths.get(nativeLib + LOCK_EXT);
                                 if (Files.notExists(lckFile)) {
                                     try {
-                                        Files.delete(nativeLib);
+                                        Files.deleteIfExists(nativeLib);
                                     } catch (Exception e) {
-                                        logger.error("Failed to delete old native lib", e);
                                     }
                                 }
                             });


gotson:
but why?

fcyu:
One of our software uses sqlite-jdbc and prints a bunch of exception stacktrace when it couldn't delete the old native lib, which confuse users. 

Thanks,

Fengchao

gotson:
i don't understand how users can be confused by log files, do users read logs as part of the application's normal behaviour ?

logs files are for investigation, not logging errors is not a good practice

gotson:
also your code does not make sense to me, as the loop is on files that are on disk and found matching the pattern, so using delete if exist is redundant, since the file exists.

You are just removing the logging in the catch block.

fcyu:
> do users read logs as part of the application's normal behaviour ?

That's correct.

> also your code does not make sense to me, as the loop is on files that are on disk and found matching the pattern, so using delete if exist is redundant, since the file exists.

Just in case that file was remove by other program right before executing `delete()`. There is almost no harm using `deleteIfExists()`, but feel free to change it.

Best,

Fengchao

fcyu:
Also, rather than throwing an exception, it prints the error message and stacktrace, which is hard to customize the behavior. If changing to throwing an exception, I guess I can catch it in my function and "mute" it.

Best,

Fengchao

gotson:
we are using SLF4J for logging, just change the logger configuration to disable that log statement.

I will close this, as this is not something we want to merge.

PR #1128, opened 2024-06-14
fix: never return arch as aarch64 when running in a 32-bit JVM

### LINKED ISSUE #1127: aarch64 native library is tried to be loaded when using a 32-bit JVM
**Describe the bug**
When running inside a 32-bit JVM on a aarch64 processor, sqlite-jdbc tries to load the aarch64 native library. This of course fails.

**To Reproduce**
```
package org.example;

public class Main {
  public static void main(String[] args) throws Exception {
    System.out.println("Arch: " + System.getProperty("sun.arch.data.model"));

    try (var connection = DriverManager.getConnection("jdbc:sqlite:/tmp/test.db");
      var statement = connection.createStatement();
      var result = statement.executeQuery("SELECT RANDOM();")) {
      while (result.next()) {
        System.out.println("Output: " + result.getLong(1));
      }
    }
  }
}
```

**Expected behavior**
As the code is running inside a 32-bit JVM, the `arm`, `armv6` or `armv7` native library should be loaded.

**Logs**
Output:
```
Arch: 32
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
OpenJDK Client VM warning: You have loaded library /tmp/sqlite-3.46.0.0-51c6d6c2-c417-498d-9da6-fb79c3df716a-libsqlitejdbc.so which might have disabled stack guard. The VM will try to fix the stack guard now.
It's highly recommended that you fix the library with 'execstack -c <libfile>', or link it with '-z noexecstack'.
Exception in thread "main" java.sql.SQLException: Error opening connection
	at org.sqlite.SQLiteConnection.open(SQLiteConnection.java:283)
	at org.sqlite.SQLiteConnection.<init>(SQLiteConnection.java:67)
	at org.sqlite.JDBC.createConnection(JDBC.java:106)
	at org.sqlite.JDBC.connect(JDBC.java:79)
	at java.sql/java.sql.DriverManager.getConnection(DriverManager.java:681)
	at java.sql/java.sql.DriverManager.getConnection(DriverManager.java:252)
	at org.example.Main.main(Main.java:9)
Caused by: org.sqlite.NativeLibraryNotFoundException: No native library found for os.name=Linux, os.arch=aarch64, paths=[/org/sqlite/native/Linux/aarch64:/usr/java/packages/lib:/usr/lib/arm-linux-gnueabihf/jni:/lib/arm-linux-gnueabihf:/usr/lib/arm-linux-gnueabihf:/usr/lib/jni:/lib:/usr/lib]
	at org.sqlite.SQLiteJDBCLoader.loadSQLiteNativeLibrary(SQLiteJDBCLoader.java:361)
	at org.sqlite.SQLiteJDBCLoader.initialize(SQLiteJDBCLoader.java:75)
	at org.sqlite.core.NativeDB.load(NativeDB.java:69)
	at org.sqlite.SQLiteConnection.open(SQLiteConnection.java:280)
	... 8 more
```

 - OS: Raspberry Pi OS
 - CPU architecture: aarch64
 - sqlite-jdbc version: 3.46.0.0

**Additional context**

tyilo:
Fixes #1127

### DIFF
--- a/src/main/java/org/sqlite/util/OSInfo.java
+++ b/src/main/java/org/sqlite/util/OSInfo.java
@@ -175,7 +175,7 @@ public class OSInfo {
 
             // for Android, we fold everything that is not aarch64 into arm
             if (isAndroid()) {
-                if (armType.startsWith("aarch64") && !is32bit) {
+                if (armType.startsWith("aarch64")) {
                     // Use arm64
                     return "aarch64";
                 } else {
--- a/src/main/java/org/sqlite/util/OSInfo.java
+++ b/src/main/java/org/sqlite/util/OSInfo.java
@@ -171,9 +171,11 @@ public class OSInfo {
             // armType (uname -m) can be armv5t, armv5te, armv5tej, armv5tejl, armv6, armv7, armv7l,
             // aarch64, i686
 
+            boolean is32bit = "32".equals(System.getProperty("sun.arch.data.model"));
+
             // for Android, we fold everything that is not aarch64 into arm
             if (isAndroid()) {
-                if (armType.startsWith("aarch64")) {
+                if (armType.startsWith("aarch64") && !is32bit) {
                     // Use arm64
                     return "aarch64";
                 } else {
@@ -190,7 +192,7 @@ public class OSInfo {
             } else if (armType.startsWith("armv5")) {
                 // Use armv5, soft-float ABI
                 return "arm";
-            } else if (armType.startsWith("aarch64")) {
+            } else if (armType.startsWith("aarch64") && !is32bit) {
                 // Use arm64
                 return "aarch64";
             }
--- a/src/main/java/org/sqlite/util/OSInfo.java
+++ b/src/main/java/org/sqlite/util/OSInfo.java
@@ -171,8 +171,6 @@ public class OSInfo {
             // armType (uname -m) can be armv5t, armv5te, armv5tej, armv5tejl, armv6, armv7, armv7l,
             // aarch64, i686
 
-            boolean is32bit = "32".equals(System.getProperty("sun.arch.data.model"));
-
             // for Android, we fold everything that is not aarch64 into arm
             if (isAndroid()) {
                 if (armType.startsWith("aarch64")) {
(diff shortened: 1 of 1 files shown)


gotson:
Seems like this is a sun specific property. Could we use anything else ?

tyilo:
> Seems like this is a sun specific property. Could we use anything else ?

It doesn't seem like it...

src/main/java/org/sqlite/util/OSInfo.java
```diff
@@ -171,9 +171,11 @@ static String resolveArmArchType() {
             // armType (uname -m) can be armv5t, armv5te, armv5tej, armv5tejl, armv6, armv7, armv7l,
             // aarch64, i686
 
+            boolean is32bit = "32".equals(System.getProperty("sun.arch.data.model"));
+
             // for Android, we fold everything that is not aarch64 into arm
             if (isAndroid()) {
-                if (armType.startsWith("aarch64")) {
+                if (armType.startsWith("aarch64") && !is32bit) {
```
gotson, line 178: i don't think we need the change in Android, not even sure those system properties would exist on Android

src/main/java/org/sqlite/util/OSInfo.java
```diff
@@ -190,7 +192,7 @@ static String resolveArmArchType() {
             } else if (armType.startsWith("armv5")) {
                 // Use armv5, soft-float ABI
                 return "arm";
-            } else if (armType.startsWith("aarch64")) {
+            } else if (armType.startsWith("aarch64") && !is32bit) {
```
gotson, line 195: by skipping this check the code will try the other methods, but it's not guaranteed that the correct type would be returned.

the following code could return `armv7` or `arm`, but not `armv6` for instance. As we have different `.so` files for each, it may not always work, as it could return an incorrect architecture.

It seems a better way to fix this would be to use something different than `uname -m` to get the arch.

tyilo: I have updated the code to always return `armv7` when running on a 32-bit JVM on an `aarch64` architecture.

I think that all `aarch64` architectures (if they support 32-bit instructions) will always support `armv7`.

gotson: > I think that all aarch64 architectures (if they support 32-bit instructions) will always support armv7.

i think that's acceptable indeed.

gotson:
thanks for the PR

PR #1133, opened 2024-06-18
fix(jdbc): generated columns with stored in SQLite are not marked as generated

### LINKED ISSUE #1132: Generated columns with `stored` in SQLite are not marked as generated
**Description**
Generated columns are not reported as such by database metadata.

**To Reproduce**
Create the following table with SQLite :

```sql
create table foo (
  id integer primary key,
  bar int not null generated always as (id + 1) stored
);
```

On the `bar` column `IS_GENERATEDCOLUMN` is `NO` with `DatabaseMetaData.getColumns`. Without the stored keyword, it is `YES`.

**Expected Behavior**
On the `bar` column `IS_GENERATEDCOLUMN` should be `YES` with `DatabaseMetaData.getColumns`.

**Additional Context**
See [JDBC3DatabaseMetaData](https://github.com/xerial/sqlite-jdbc/blob/99aa8d9e41113d6bec2f77e0084a328cacbc98c7/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java#L1081). It only checks for a value of "2", when according to the [SQLite documentation](https://sqlite.org/pragma.html#pragma_table_xinfo:~:text=a%20dynamic%20or%20stored%20generated%20column%20%282%20or%203%29), it should be checking for both 2 and 3.

sualeh:
Fixes #1132 - Generated columns with stored in SQLite are not marked as generated

### DIFF
--- a/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
@@ -1078,7 +1078,7 @@ public abstract class JDBC3DatabaseMetaData extends CoreDatabaseMetaData {
                             colType = colType.substring(0, iStartOfDimension).trim();
                         }
 
-                        int colGenerated = "2".equals(colHidden) ? 1 : 0;
+                        int colGenerated = Arrays.binarySearch(new String[] {"2", "3"}, colHidden) >= 0 ? 1 : 0;
 
                         sql.append("select ")
                                 .append(i + 1)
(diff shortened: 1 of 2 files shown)


src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
```diff
@@ -1078,7 +1078,10 @@ public ResultSet getColumns(String c, String s, String tblNamePattern, String co
                             colType = colType.substring(0, iStartOfDimension).trim();
                         }
 
-                        int colGenerated = "2".equals(colHidden) ? 1 : 0;
+                        int colGenerated =
+                                Arrays.binarySearch(new String[] {"2", "3"}, colHidden) >= 0
```
gotson, line 1082: it seems convoluted instead of doing a OR condition

sualeh: Modified. @gotson Please take another look.

PR #1150, opened 2024-07-29
feat: add riscv64 binaries

qyl27:


gotson:
some context would help. I see it's using Dockcross, so that seems alright, but why did the windows dll changed?

gotson:
> Oh, the changes to windows dll seem like by the bot, should I revert it?

yes please

> The os info of RISC-V did is `riscv64`, I had a misunderstanding with the `OSInfo` class.

no that's fine, i checked the code and if the `os.info` value is the same as the folder name then the loading should be good.

qyl27:
All done.

gotson:
Thanks for the PR, a good addition to the project!

sualeh:
Also see https://github.com/schemacrawler/SchemaCrawler/discussions/1624

Issue #1141, opened 2024-07-11
Bugs in ExtendedCommand.removeQuotation cause possible StringIndexOutOfBoundsException


arthurscchan:
**Describe the bug**
The `org.sqlite.ExtendedCommand::removeQuotation(String)` could throw an unexpected StringIndexOutOfBoundsException with invalid input. The method checks if the provided string starts and ends with the same quotation mark and removes them. But if the string is invalid with just a single quotation mark, the conditional check still passes but the substring method will throw a StringIndexOutOfBoundsException because the second argument for the substring method will be less than the first argument. If this method is being used in part of more complicated command string processing, then the unexpected exception could crash the run unexpectedly.

**To Reproduce**
Here is a proof of concept code for triggering the bug. Just compile and run it is enough to trigger the bug.
```java
import org.sqlite.ExtendedCommand;
public class ProofOfConcept {
  public static void main(String...args) {
    ExtendedCommand.removeQuotation("\"");
  }
}
```

**Expected behaviour**
It should return the original string since in theory, it should fail in the conditional check.

**Suggested fix**
Changing the conditional check to consider a single quotation string should be better. For example, add a length check to avoid the single quotation character string to pass the existing conditional check.

gotson:
PR Welcome

PR #1161, opened 2024-08-19
fix: possible StringIndexOutOfBoundsException in ExtendedCommand

arthurscchan:
This PR proposes a fix for #1141 .


gotson:
do you think you could add a test covering this ?

arthurscchan:
@gotson Really sorry for missing your comment. I have added in the unit test for the removeQuotation method.

gotson:
> @gotson Really sorry for missing your comment. I have added in the unit test for the removeQuotation method.

the test code doesn't compile

gotson:
i've refactored your test code:

```java
@ParameterizedTest
  @MethodSource
  public void removeQuotation(String input, String expected) throws SQLException {
    assertThat(ExtendedCommand.removeQuotation(input)).isEqualTo(expected);
  }

  private static Stream<Arguments> removeQuotation() {
    return Stream.of(
        Arguments.of(null, null), // Null String
        Arguments.of("'", "'"), // String with one single quotation only
        Arguments.of("\"", "\""), // String with one double quotation only
        Arguments.of("'Test\"", "'Test\""), // String with two mismatch quotations
        Arguments.of("'Test'", "Test"), // String with two matching single quotations
        Arguments.of("\"Test\"", "Test"), // String with two matching double quotations
        Arguments.of("'Te's\"t'", "Te's\"t") // String with more than two quotations
    );
  }
```

arthurscchan:
@gotson, I have changed the parameter name but forget to push in the new one. Sorry for my careless mistake. And thanks for helping me to fix it.

Issue #1155, opened 2024-08-07
[Feature] Extract DB files to a subdirectory within temporary folder


belugabehr:
**Is your feature request related to a problem? Please describe.**

The library performs a cleanup routine of the `/tmp` directory at the start of every execution.

https://github.com/xerial/sqlite-jdbc/blob/61fda53dd39bc31dbf46f5f4523d85334f6c0fb9/src/main/java/org/sqlite/SQLiteJDBCLoader.java#L90C48-L90C58

This means that the duration of time to perform this action depends on the number of files in the `/tmp` directory at the time of startup. 

**Describe the solution you'd like**
This can be mostly mitigated by first creating a sub-directory in the `/tmp` directory (e.g., "xerial-sqlite-jdbc") and only store JDBC native libs in this directory. The size of this directory will be much less at risk of having large numbers of files present.

gotson:
PR welcome

gotson:
I thought about that, and we already have `org.sqlite.tmpdir`, so this seems redundant. What do you think?

joffrey-bion:
I think it's still relevant. `org.sqlite.tmpdir` is a nice way to _customize_ this, but its default value could be `{java.io.tmpdir}/something`.

Issue #1154, opened 2024-08-07
More Fine Grained Control for Loading Native JDBC Lib


belugabehr:
**Is your feature request related to a problem? Please describe.**
Follow-up to #1140.

For the greatest amount of deployment safety, I have configured my application to use a static copy of `sqlitejdbc.so` via `-Dorg.sqlite.lib.path` that gets deployed alongside the application. There are however some issues with this approach:

1. The [cleanup](https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java/org/sqlite/SQLiteJDBCLoader.java#L73C13-L73C23) routine gets run every time, even though the JDBC native lib will never be present there - this adds startup time that is variable as it depends on the number of items that happen to be at `/tmp` at any given time
2. Logging is not sufficient and therefore if the `org.sqlite.lib.path` option is misconfigured, it is not easy to identify. That is to say, if the JDBC native lib is missing, or if the `lib.path` is incorrect, then it silently fails and falls-back to loading from the JAR resources (and using `/tmp`) and there is no indication that this has happened.

**Describe the solution you'd like**
Make the default to load from JAR resources, but allow users to explicitly specify (e.g., `-Dorg.sqlite.lib.location` ) where the native JDBC lib should be loaded from. If it cannot be loaded, the behavior should be to fail-fast, not fall-back.

Added logging (at least DEBUG, maybe INFO) where the Native JDBC lib file is loaded from.

gotson:
> If it cannot be loaded, the behavior should be to fail-fast, not fall-back.

Can you explain why you would not want fallback ?

> Added logging (at least DEBUG, maybe INFO) where the Native JDBC lib file is loaded from.

PR welcome

belugabehr:
> Can you explain why you would not want fallback ?

Sure! Basically, if I misconfigure my application, I would rather it fail-fast than to fall-back. I very much want it to use the static copy of the JDBC native library I have deployed alongside my app. If it falls-back, then the fact that it is misconfigured is hidden from me and everything _looks_ like it is working as expected.

gotson:
IMHO if you set both `-Dorg.sqlite.lib.path` and `-Dorg.sqlite.lib.name`, then we should not fallback, and disable cleanup. That would be a breaking change though.

Or we need a new option to disable cleanup, and disable fallback.

belugabehr:
> Or we need a new option to disable cleanup, and disable fallback.

[...]

Issue #1140, opened 2024-07-10
Do Not Use Lock File


belugabehr:
**Describe the bug**
The SQLiteJDBCLoader class uses a "lock file" to determine when a process is using a SQLite binary library file. The lock file is unreliable, it may leak the lock file, and therefore may allow many instances of the SQLite binary library files to accumulate in the `/tmp` directory.

> Deletion will be attempted only for normal termination of the virtual machine, as defined by the Java Language Specification.
>
> Note: this method should not be used for file-locking, as the resulting protocol cannot be made to work reliably. The [FileLock](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/FileLock.html) facility should be used instead.

https://docs.oracle.com/javase/8/docs/api/java/io/File.html#deleteOnExit--

**Expected behavior**
Unused copies of the SQLite binary files are cleaned up reliably.

belugabehr:
For additional context, I came across a scenario where a service on a host was stuck in a fail-retry loop. It ended up filling up the drive of the host with these files and knocked out all other services on the host as a result.

belugabehr:
I've been looking at this question quite a bit in-depth.

The biggest issue in my mind is trying to cleanup after oneself in failure scenarios. I think the answer is: you don't.

As things stand, every time the application starts, this library scans the entire contents of the `/tmp` directory. From a traditional, long-running, production server standpoint, this is not ideal as it adds some amount of extra startup time and variability into the initialization.  If you are running a long-running traditional production server, you are best served by shipping your application with a copy of the `sqlitejdbc` binary and you configure `sqlite-jdbc` to use that explicit copy instead of trying to use this auto-deploy feature. Also, as I learned the hard way, the current cleanup code is not all that robust and will leak files if both the `jdbcsqlite` binary _and_ its corresponding `lock` file are not cleaned up during an abnormal termination of the JVM (i.e., `System.exit(-1)`).

If you are using an application that is transient: short-lived docker container, or AWS Lambda, then the auto-discovery feature seems perfectly fine and simplifies things for you.

[...]

PR #1164, opened 2024-08-20
Create native SQLite loader chain

belugabehr:
Related to #1154

I have separated all of the various ways of loading the native SQLite file into their own discrete classes.

Note that all of these classes are in their own package, therefore, I can tune the logging just for this one package to get a more clear idea of what is going on in the loading of the native file.

Also, if I want to implement something like #1140 and/or #1155, then I can create the new feature as a new class, and then provide a System property to toggle which loading mechanism I want to use "default" v.s. "cache" or something like that.

### DIFF
--- a/src/main/java/org/sqlite/NativeLibraryNotFoundException.java
+++ b/src/main/java/org/sqlite/NativeLibraryNotFoundException.java
@@ -4,4 +4,12 @@ public class NativeLibraryNotFoundException extends Exception {
     public NativeLibraryNotFoundException(String message) {
         super(message);
     }
+
+    public NativeLibraryNotFoundException(Throwable cause) {
+        super(cause);
+    }
+
+    public NativeLibraryNotFoundException(String message, Throwable cause) {
+        super(message, cause);
+    }
 }
--- a/src/main/java/org/sqlite/SQLiteJDBCLoader.java
+++ b/src/main/java/org/sqlite/SQLiteJDBCLoader.java
@@ -26,24 +26,18 @@ package org.sqlite;
 
 import java.io.*;
 import java.net.URL;
-import java.net.URLConnection;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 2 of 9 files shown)


gotson:
hi, i don't quite see how you will solve #1140 and #1155 with this ? It seems this is just a stepping stone refactoring without any other feature ?

I am generally against that kind of preparatory work, and would prefer a PR that solves issues or brings new features.

belugabehr:
Hello,

This moves all the code into a new package, so I can set my logger to more verbose just for this one package to get better understand where the native lib is being loaded from.

> Added logging (at least DEBUG, maybe INFO) where the Native JDBC lib file is loaded from.

For the other changes, you pointed out correctly that all of these proposals are breaking backwards compatibility in subtle ways.

So, in order to preserve backwards functionality, and to but new features behind flags, it will be more simple:


```java
if (featureFlag) {
    MyCacheBasedLoader.getInstance().loadSQLiteNative();
} else {
    DefaultSQLiteNativeLoaderChain.getInstance().loadSQLiteNative();
}
```

> and would prefer a PR that solves issues or brings new features.

The feature here is the logging aspect and easy to add new loading functionality.  Also, this is my first engagement with the project, so I am hesitant to spend the time delivering a new feature and then it gets rejected or ignored. I hope you understand that perspective :)

gotson:
if it's just about moving the loader into it's own package for logging purposes, you can do a PR for that without all the refactoring, which IMHO may be useful, but may be bloated also. For instance it breaks the GraalVM native image tests.

Moving SQLiteJSBCLoader into its own `loader` subpackage could maybe suffice for now?

05nelsonm:
Would also like to add any usage of `Paths` will break android API 25 and below. Must use `java.io.File`.

gotson:
> Would also like to add any usage of `Path` will break android API 25 and below. Must use `java.io.File`.

API 25 is 8 years old, this library will not make any stretch to support old android API versions.

PR #1178, opened 2024-09-18
feat: make slf4j optional with fallback on JUL

### LINKED ISSUE #1094: why was slf4j-api added as a dependency?
i read: https://github.com/xerial/sqlite-jdbc/issues/997#issuecomment-1776972487
and: https://github.com/xerial/sqlite-jdbc/commit/32082c0504d02ee4b0b379b6cc4f26fc9fcf95ab
and: https://github.com/xerial/sqlite-jdbc/commit/89dbda1bb21bc7df004419edb32e26f17bbb0e19

but i can not see any reason this was done.

here it says "Read the Changelog", but there is nothing about slf4j-api in the changelog.
why was this added? can you link to documentation about the decision or some code in sqlite that requires it?

why was it not possible to make this optional?

headius:
In order to avoid the hard dependency on slf4j, this patch abstracts the Logger interface and LoggerFactory factory to fall back on a java.util.logging-based implementation if slf4j is not present.

Fixes #1094

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -415,6 +415,7 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
             <version>1.7.36</version>
+            <optional>true</optional>
         </dependency>
         <!--
         This dependency makes compilation on non-GraalVM versions possible.
--- a/src/main/java/org/sqlite/JDBC.java
+++ b/src/main/java/org/sqlite/JDBC.java
@@ -18,12 +18,12 @@ package org.sqlite;
 
 import java.sql.*;
 import java.util.Properties;
-import java.util.logging.Logger;
-import org.slf4j.LoggerFactory;
 import org.sqlite.jdbc4.JDBC4Connection;
+import org.sqlite.util.Logger;
+import org.sqlite.util.LoggerFactory;
 
 public class JDBC implements Driver {
-    private static final org.slf4j.Logger logger = LoggerFactory.getLogger(JDBC.class);
+    private static final Logger logger = LoggerFactory.getLogger(JDBC.class);
     public static final String PREFIX = "jdbc:sqlite:";
 
     static {
@@ -49,7 +49,7 @@ public class JDBC implements Driver {
         return false;
     }
 
-    public Logger getParentLogger() throws SQLFeatureNotSupportedException {
+    public java.util.logging.Logger getParentLogger() throws SQLFeatureNotSupportedException {
         // TODO
         return null;
     }
--- a/README.adoc
+++ b/README.adoc
@@ -29,19 +29,18 @@ SQLite JDBC is a library for accessing SQLite databases through the JDBC API. Fo
 
 . <<Download,Download>> `sqlite-jdbc-{project-version}.jar`
 then append this jar file into your classpath.
-. https://search.maven.org/remotecontent?filepath=org/slf4j/slf4j-api/1.7.36/slf4j-api-1.7.36.jar[Download] `slf4j-api-1.7.36.jar` then append this jar file into your classpath.
 . Open a SQLite database connection from your code. (see the example below)
 
 === Example usage
 
-Assuming `sqlite-jdbc-{project-version}.jar` and `slf4j-api-1.7.36.jar` are placed in the current directory.
+Assuming `sqlite-jdbc-{project-version}.jar` is placed in the current directory.
 
 [source,shell,subs="attributes+"]
 ----
 > javac Sample.java
-> java -classpath ".;sqlite-jdbc-{project-version}.jar;slf4j-api-1.7.36.jar" Sample   # in Windows
+> java -classpath ".;sqlite-jdbc-{project-version}.jar" Sample   # in Windows
 or
-> java -classpath ".:sqlite-jdbc-{project-version}.jar:slf4j-api-1.7.36.jar" Sample   # in macOS or Linux
+> java -classpath ".:sqlite-jdbc-{project-version}.jar" Sample   # in macOS or Linux
 name = leo
 id = 1
 name = yui
--- a/src/main/java/org/sqlite/SQLiteJDBCLoader.java
+++ b/src/main/java/org/sqlite/SQLiteJDBCLoader.java
@@ -39,9 +39,9 @@ import java.util.List;
 import java.util.Properties;
 import java.util.UUID;
 import java.util.stream.Stream;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.sqlite.util.LibraryLoaderUtil;
+import org.sqlite.util.Logger;
+import org.sqlite.util.LoggerFactory;
 import org.sqlite.util.OSInfo;
 import org.sqlite.util.StringUtils;
 
--- a/src/main/java/org/sqlite/core/NativeDB.java
+++ b/src/main/java/org/sqlite/core/NativeDB.java
@@ -19,14 +19,14 @@ package org.sqlite.core;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.sql.SQLException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.sqlite.BusyHandler;
 import org.sqlite.Collation;
 import org.sqlite.Function;
 import org.sqlite.ProgressHandler;
 import org.sqlite.SQLiteConfig;
 import org.sqlite.SQLiteJDBCLoader;
+import org.sqlite.util.Logger;
+import org.sqlite.util.LoggerFactory;
 
 /** This class provides a thin JNI layer over the SQLite3 C API. */
 public final class NativeDB extends DB {
--- a/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
@@ -21,12 +21,12 @@ import java.util.Properties;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.sqlite.SQLiteConnection;
 import org.sqlite.core.CoreDatabaseMetaData;
 import org.sqlite.core.CoreStatement;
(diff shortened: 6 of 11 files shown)


headius:
Note I have only implemented the bare minimum interface necessary to allow this patch to drop into existing code. Additional log levels and log level methods can be implemented in the future if necessary.

gotson:
Hi, thanks for the PR. Can you check the failing CI tests ?

gotson:
looks like you may have forgotten to commit the new Logger class

headius:
> looks like you may have forgotten to commit the new Logger class

Classic! Re-pushed with the missing files.

src/main/java/org/sqlite/JDBC.java
```diff
@@ -49,7 +49,7 @@ public boolean jdbcCompliant() {
         return false;
     }
 
-    public Logger getParentLogger() throws SQLFeatureNotSupportedException {
+    public java.util.logging.Logger getParentLogger() throws SQLFeatureNotSupportedException {
```
gotson, line 52: this needs to return a `java.util.logging.Logger` as per https://docs.oracle.com/javase/8/docs/api/java/sql/Driver.html#getParentLogger--

Makefile.common
```diff
@@ -59,7 +55,7 @@ ifeq (,$(findstring $(strip $(target)),$(known_targets)))
   target := Default
 endif
 
-$(info Will build using target: $(target) (detected os: $(OS_NAME), arch: $(OS_ARCH)))
+$(info Will build using target: $(target) (detected os: $(OS_NAME) arch: $(OS_ARCH)))
```
gotson, line 58: why the change here?

headius: I reverted the Makefile and Makefile.common changes (mostly adding slf4j to the build) because they broke the "amalgamated" job in CI. I assume the slf4j jar is needed for compiling some Java things from the Makefile.

src/main/java/org/sqlite/JDBC.java
headius: It is returning that. I added the package because the new sqlite-jdbc Logger class conflicts.

Makefile.common
gotson: reverting is fine, we shouldn't need slf4j anymore in the makefile if it's optional, i was more referring to that particular `,` 🙃

src/main/java/org/sqlite/JDBC.java
gotson: right, i misread the diff somehow, please ignore

headius:
I pushed two changes to address failures in a previous job run:

* Apply spotless formatting to the two small javadocs in Logger and LoggerFactory.
* Revert the changes to Makefile and Makefile.common as the slf4j jar file appears to be needed in those builds.

gotson:
> Revert the changes to Makefile and Makefile.common as the slf4j jar file appears to be needed in those builds.

it shouldn't be, but it's been a while since i looked at it

headius:
> it shouldn't be, but it's been a while since i looked at it

@gotson Originally the slf4j jar was added to those files during the addition of slf4j logging. Without the jar in CLASSPATH, the "amalgamation" job failed to run its `javac` targets. It is green now with the jar restored.

Makefile.common
headius: The changes I made there were themselves reverting the addition of slf4j jar from previous commits. The comma seems to be present in that commit: https://github.com/xerial/sqlite-jdbc/commit/1663cfec7a7bf167f299f5b93f0af754649e6c82#diff-f190d29e2cad4f32539c8b73d8af07c77f0aac63b5c6553c06b117c08a70b494R62

I am not sure what to say about the slf4j dependency in Makefile and Makefile.common; without it present, the "amalgamation" job fails to `javac` successfully.

headius:
Everything is green now. The job "test aarch64 ubuntu_latest jdk21" took a very long time, but I assume that is because it is running aarch64 tests in an emulator on an x86_64 job runner.

Makefile.common
headius: Note that this change is moot since I reverted to the old Makefile.common (which does have the comma, added in 1663cfe).

gotson:
> Without the jar in CLASSPATH, the "amalgamation" job failed to run its `javac` targets. It is green now with the jar restored.

we need to fix this, it should not be required. I pulled your PR code and tested locally, and the `make native` fails because of some issues in `NativeDB.java`:

```
"$JAVA_HOME/bin/javac"  -d target/common-lib -sourcepath src/main/java -h target/common-lib src/main/java/org/sqlite/core/NativeDB.java
src/main/java/org/sqlite/util/LoggerFactory.java:89: error: package org.slf4j does not exist
        final org.slf4j.Logger logger;
                       ^
src/main/java/org/sqlite/util/LoggerFactory.java:92: error: package org.slf4j does not exist
            logger = org.slf4j.LoggerFactory.getLogger(hostClass);
                              ^
```

I would say it's because you define this field in `SLF4JLogger`: `final org.slf4j.Logger logger;`

gotson:
Not sure we can work around that though, maybe using reflection ?

pom.xml
```diff
@@ -415,6 +415,7 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
             <version>1.7.36</version>
+            <scope>compile</scope>
```
jcgay, line 418: `compile` is already the default scope for a dependency with Maven. Adding it doesn't change anything : https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope

I guess if we want to make it optional it should not be automatically added transitively as a dependency and we should use the `<optional>true</optional>` configuration ? https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html

gotson: i am not very familiar with Maven, but when we reference the SLF4J package, java needs to know the package, even if it is not included in the build, to compile the code.

i think we can leave it as is, and leave slf4j in the makefile to compile NativeDB.java

jcgay: The `<optional>` configuration does not exclude the dependency for the `sqlite-jdbc` build. It will still be added to the classpath when compiling.

The `<optional>` configuration will only affect builds which depends on `sqlite-jdbc`. The `slf4j-api` will not be added as a transitive dependency. It should make the consumption easier like:
 - one that don't want to use SLF4J would add:
```
<dependencies>
    <dependency>
      <groupId>org.xerial</groupId>
      <version>3.46.1.0</version>
    </dependency>
</dependencies>
```
instead of:
```
<dependencies>
    <dependency>
      <groupId>org.xerial</groupId>
      <version>3.46.1.0</version>
    </dependency>
    <exclusions>
       <exclusion>
         <groupId>org.slf4j</groupId>
         <artifactId>slf4j-api</artifactId>
       </exclusion>
     </exclusions>
</dependencies>
```
 - one that want to use SLF4J will add:
```
<dependencies>
    <dependency>
      <groupId>org.xerial</groupId>
      <version>3.46.1.0</version>
    </dependency>
<dependency>
      <groupId>org.slf4j</groupId>
      <artifactId>slf4j-api</artifactId>
      <version>1.7.36</version>
    </dependency>
</dependencies>
```
But if we want to use SLF4J the dependency should be already declared somewhere 😇.

It just triggers me because we say SLF4J is optional but it will still be added in the dependency graph if we don't exclude it explicitly. It may be not what people expect when using `sqlite-jdbc` but this is not necessarily a big deal if this is documented 😇.

gotson: i guess optional would be better than compile then ?

jcgay: In fact this is an addition to the `compile` scope, it would be: 

```
<dependency>
            <groupId>org.slf4j</groupId>
            <artifactId>slf4j-api</artifactId>
            <version>1.7.36</version>
            <scope>compile</scope>
            <optional>true</optional>
</dependency>
```

or just (which is equivalent):

```
<dependency>
            <groupId>org.slf4j</groupId>
            <artifactId>slf4j-api</artifactId>
            <version>1.7.36</version>
            <optional>true</optional>
</dependency>
```

Because with Maven, when a scope is not set, it is `compile` by default.

The other way to reproduce the same transitive mechanism behavior (not adding slf4j-api automatically in the classpath when you depend on sqlite-jdbc) would be to replace the `compile` scope by `provided`. But it does change a little the meaning because it would indicate that we really need `slf4j-api` at runtime and that it should be provided by something else (an application server, etc): https://stackoverflow.com/questions/40393098/when-to-use-optional-dependencies-and-when-to-use-provided-scope

My 2 cents is to make the dependency optional in `sqlite-jdbc` which will not resolve `slf4j-api` transitively when we depend on `sqlite-jdbc` and let the consumer use this logging facade or not 😇.

gotson: > My 2 cents is to make the dependency `<optional>` in `sqlite-jdbc` which will not resolve `slf4j-api` transitively when we depend on `sqlite-jdbc` and let the consumer use this logging facade or not 😇.

i am OK with that

headius: > My 2 cents is to make the dependency <optional>

I did not know this! I will make the change and re-push.

headius:
I have updated the build to make slf4j optional, as @jcgay suggested. This should resolve the last few concerns about this PR, but let me know if anything else needs to change!

gotson:
@headius thanks for the PR. I have added one commit to update the coding tests for loggers. Thanks @jcgay for the maven insights.

headius:
@gotson Thank you! Looking forward to the release.

rmraya:
Hi,
It looks like you forgot to remove the dependency on `org.slf4j` from `src/main/java9/module-info.java`
Regards,
Rodolfo

headius:
@rmraya D'oh! Could you push a PR? I believe it should be `requires static`.

From https://www.oracle.com/corporate/features/understanding-java-9-modules.html:

> There is also a requires static directive to indicate that a module is required at compile time, but is optional at runtime. This is known as an optional dependency and won’t be discussed in this introduction.

rmraya:
Marking as `static` means that it is needed at compile time and what I need is to get rid of slf4j jars.

headius:
I'm not sure what the right option is then. Won't removing it altogether mean it cannot access those APIs at all at runtime?

headius:
It still seems to me like `requires static` is correct. It marks the module as required only at compile time for that module. I don't think it would force you to have slf4j at compile time as a user of sqlite-jdbc.

https://www.baeldung.com/java-modularity#2-requires-static

rmraya:
OK, let me build from source and test. I'll post a PR when I have it working.

rmraya:
Created new PR #1205

PR #1201, opened 2024-10-22
Add support for default SQLite schemas

shybovycha:
According to [SQLite docs](https://www.sqlite.org/lang_createtable.html), there are at least two schemas supported: `main` and `temp`:

> If a `schema-name` is specified, it must be either "main", "temp", or the name of an [attached database](https://www.sqlite.org/lang_attach.html). In this case the new table is created in the named database. If the "TEMP" or "TEMPORARY" keyword occurs between the "CREATE" and "TABLE" then the new table is created in the temp database. It is an error to specify both a schema-name and the TEMP or TEMPORARY keyword, unless the `schema-name` is "temp". If no schema name is specified and the TEMP keyword is not present then the table is created in the main database.

Thus, two things need to happen for this JDBC driver to be compatible with SQLite (at least from this feature perspective):

1. the `DatabaseMetadata#getSchemas()` must return at least those two schemas (`main` and `temp`); hopefully the attached database too, but I am unsure if this is possible to determine at the relevant point in the code
2. the `DatabaseMetadata#supportsSchemasInTableDefinitions()` must return `true` to indicate the consumers the table could be referenced either via its raw name or using a schema prefix

Currently the driver returns an empty result set for the `DatabaseMetadata#getSchemas()` and `false` for `DatabaseMetadata#supportsSchemasInTableDefinitions()`. This breaks some of the consumers (for instance, the applications using 3rd party APIs or the driver directly when trying to determine database schema at runtime).


gotson:
#829 may be related

PR #1205, opened 2024-10-23
fix: update module-info.java to make org.slf4j static requirement

rmraya:


headius:
Looks good to me!

gotson:
thanks!

headius:
@gotson Thank you!

PR #1250, opened 2025-03-18
build(deps-dev): bump org.junit.jupiter:junit-jupiter to 5.12.1

gotson:
Use the JUnit BOM, but it still fails for native-image with GraalVM 22+

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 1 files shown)


kkriske:
I'll take a look

kkriske:
Put this together: https://github.com/xerial/sqlite-jdbc/pull/1252
Looks like official JUnit 5.12 support is still incomplete, but the GraalVM team is working on a more future-proof approach that would not run into this kind of issues.

PR #1255, opened 2025-03-24
refactor: Simplify type checks in `getColumnType` using Sets

priyanshsolanki:
- Replaced complex and repetitive conditions with `Set` collections for type names.
- Improved readability and maintainability by grouping related type names.
- Added sets for boolean, integer, float, text, and blob types.

This change does not affect functionality but makes the code cleaner and easier to maintain.


gotson:
you are going to define those sets every time the function is called. Extracting could be useful if the comparisons were done more than once, but it's not the case. Adds complexity for nothing.

PR #1256, opened 2025-03-24
refactor: Centralize all Pattern constants into `PatternConstants`

priyanshsolanki:
- Moved all regex patterns (`TYPE_INTEGER`, `TYPE_VARCHAR`, `PK_UNNAMED_PATTERN`, etc.) into a single `PatternConstants` class.
- Added patterns for column types, primary/foreign keys, date formats, and SQL commands.
- Updated references to use the centralized constants.

This change improves code organization, reduces duplication, and makes the patterns easier to maintain.

### DIFF
--- a/src/main/java/org/sqlite/ExtendedCommand.java
+++ b/src/main/java/org/sqlite/ExtendedCommand.java
@@ -14,6 +14,9 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.sqlite.core.DB;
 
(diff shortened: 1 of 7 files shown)


gotson:
> Moved all regex patterns (TYPE_INTEGER, TYPE_VARCHAR, PK_UNNAMED_PATTERN, etc.) into a single PatternConstants class.

I don't unserstand this. Instead of having the pattern colocated to where it is used, it's now located in another file that has nothing to do with it.

It centralizes things based on their type instead of their usage, IMO it just adds complexity.

Issue #1260, opened 2025-03-28
Potential endless loop in applications that open a non-existing database file


hofmandl1:
**Describe the bug**

The `open` function https://github.com/xerial/sqlite-jdbc/blob/bd0e99541a70a31da41f7d7c4153e7ac62840cad/src/main/java/org/sqlite/SQLiteConnection.java#L219 has the (undocumented?) side-effect of temporarily creating a 0-bytes file when trying to open a non-existing database which seems to be an in-Java location-writeability test done before handing over control to native code.

The 0-bytes file is immediately deleted after creation, which thus may seem innocent, however our application sometimes tries to open non-existing sqlite database files (our application is working with multiple database files at once, that can be registered and deregistered and manual deletion of database files is something we support).

That attempt to open a non-existing database file then temporarily creates the database-file (with 0 bytes) leading to a filesystem level file-creation-event which our application also reacts to and which leads to another attempt of opening the database-file. As the database-file has already been deleted we have another attempt to open a non-existing database file and thus an endless loop or at least a long sequence of open attempts (the loop eventually breaks, if the open attempt happens before file-deletion).

Of course, there is a way to workaround this problem on application level, e.g. doing an existance check before attempting to open the database, however IMHO that should not be necessary as it
a) is bad practice because it is racy and
b) requires an additional filesystem access

Note, that our application also explicitly resets SQLiteOpenMode.CREATE because it is not supposed to create the databases it is working with on its own.

**To Reproduce**
A reproducer would involve filesystem watching and probably be highly scheduling-dependent, so I hope my explanation above is sufficient.

**Expected behavior**
It should be possible to just try to open the file and from the caught exception determine that the problem was a non-existing file without any side-effects like a temporarily created file.

**Logs**
N/A

**Environment (please complete the following information):**
I think we were observing this problem first (and maybe only) on windows, but it should be system-independent, and probably also depends on the OS scheduling.

**Additional context**
Our solution since a couple of years is to patch the `open`-function and simply remove the whole logic in the `else`-branch that creates the empty file
replacing it by simply doing `fileName = new File(fileName).getAbsolutePath();`
which is working flawlessly for us, since we've been building sqlite-jdbc ourselves for a couple of other reasons.
However, we would very much like to not build sqlite-jdbc ourselves anymore, hence this report.

gotson:
> **Expected behavior**
[...]

PR #1263, opened 2025-04-04
remove an undocumented side-effect from SQLiteConnection::open

hofmandl1:
The side-effect consisted in temporarily creating a non-existing db-file to test for write-permission, see #1260
Also add some test to better document the expected behavior.

### DIFF
--- a/src/test/java/org/sqlite/ConnectionTest.java
+++ b/src/test/java/org/sqlite/ConnectionTest.java
@@ -17,13 +17,13 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
 import org.junit.jupiter.api.io.TempDir;
 import org.sqlite.SQLiteConfig.JournalMode;
 import org.sqlite.SQLiteConfig.Pragma;
 import org.sqlite.SQLiteConfig.SynchronousMode;
-import org.sqlite.util.OSInfo;
 
 /**
  * These tests check whether access to files is working correctly and some Connection.close() cases.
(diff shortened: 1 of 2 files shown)


hofmandl1:
Note, that the PR currently does NOT preserve the status quo error messages. Instead it preserves the orignal SQLiteException containing the SQLiteErrorCode CANT_OPEN.

I *could* add a try-catch-block for `SQLiteException` around the native code that if the error-code is CANT_OPEN tries determining write-permission via e.g. https://docs.oracle.com/javase/8/docs/api/java/io/File.html#canWrite-- which of course is not as exact as actually trying to write, however it won't have that potentially loop-creating side-effect.

I would however, prefer to merge as-is. Providing a more detailed / exact error is IMHO the responsibility of native SQLite via their open error codes and sqlite-jdbc shouldn't try to work around / enhance on this.

gotson:
can you check the code formatting and failing tests ?

Can you also change the assertions in the test to assertJ like in the rest of the project ?

hofmandl1:
> can you check the code formatting and failing tests ?
Done. Though the only solution for the failing test coming to my mind was skipping the test on windows, where apparently setting readonly from java code doesn't work that easily. But  I guess it's fine if the test doesn't run on all platforms as relying on native sqlite error codes has the benefit that we can also rely on their platform abstraction efforts.

> Can you also change the assertions in the test to assertJ like in the rest of the project ?
Done

gotson:
> can I somehow retrigger the checks? I'm not that familiar with github ..

because you force pushed, i have to approve again. Don't force push, it's not easy for reviewers to understand what has changed.

src/test/java/org/sqlite/ConnectionTest.java
```diff
@@ -409,4 +415,46 @@ public void overrideURIPragmaValuesWithProperties() throws Exception {
...
+                                        .isEqualTo(SQLiteErrorCode.SQLITE_CANTOPEN));
+        assertThat(Files.exists(nonExisting)).isFalse();
+    }
+
+    @SuppressWarnings("resource")
+    @Test
+    public void openNonExistingFileInReadOnlyDirectory(@TempDir Path tmpDir) {
+        // skip test on windows as there marking a folder as readonly doesn't seem to work with pure
+        // java means
+        Assumptions.assumeFalse(OSInfo.getOSName().equals("Windows"));
```
gotson, line 440: this could be replaced by `@DisabledOnOs`

src/test/java/org/sqlite/ConnectionTest.java
```diff
@@ -409,4 +415,46 @@ public void overrideURIPragmaValuesWithProperties() throws Exception {
...
+        assertThatExceptionOfType(SQLiteException.class)
+                .isThrownBy(() -> cfg.createConnection("jdbc:sqlite:" + nonExisting))
+                .satisfies(
+                        e ->
+                                assertThat(e.getResultCode())
+                                        .isEqualTo(SQLiteErrorCode.SQLITE_CANTOPEN));
+        assertThat(Files.exists(nonExisting)).isFalse();
+    }
+
+    @SuppressWarnings("resource")
```
gotson, line 435: why do you need this ?

hofmandl1: I'm developing with Intellij and get this warning 
image
Intellij also offers me to add createConnection to some list of ignored methods (not sure where that ends up). The annotation I used should also be respected by eclipse, so is a little bit more cross-IDE.
I've found a way to put it closer to the culprit. The annotation only works on fields, methods or statements which is why it needs introduction of an unused variable though ..

gotson (approved):
lgtm

PR #1274, opened 2025-05-19
OSInfo: Update link to Java bug #8005545

howlger:
In a comment, update the link to Java bug #8005545, as the current one leads to a webpage saying: "This bug is not available."


gotson:
thanks

Issue #1210, opened 2024-11-09
Inaccurate android detection when outside of termux environment


alexytomi:
**Describe the bug**
When attempting to use a Minecraft Java mod (Distant Horizons) under PojavLauncher which uses this library to store level data, it picks the drivers for Linux instead of Linux-Android, failing and crashing

**Environment:**
 - OS: Android 13 OneUI 5.1
 - CPU architecture: arm64
 - sqlite-jdbc version: sqlite-3.43.0.0

**Additional context**
Using -Dos.name seems to be something that should work, but due to Minecraft just outright crashing when it's set to Android, it's not an option for this situation.

The current workaround is to either load the Linux-Android libraries using another mod, putting it inside the JRE, or overwriting the Linux libraries with Linux-Android ones.

gotson:
this is how the detection works for Android now: https://github.com/xerial/sqlite-jdbc/blob/4a419981aab112b398ca0a998c9b053d07b7fcba/src/main/java/org/sqlite/util/OSInfo.java#L120-L130

Most likely the launcher you are using does not fit in those 2 cases.

alexytomi:
> this is how the detection works for Android now:
> 
> https://github.com/xerial/sqlite-jdbc/blob/4a419981aab112b398ca0a998c9b053d07b7fcba/src/main/java/org/sqlite/util/OSInfo.java#L120-L130
> 
> Most likely the launcher you are using does not fit in those 2 cases.

It doesn't fit in isAndroidTermux() but I can set -Djava.runtime.name=android yet that has no effect.

I also tried -Djava.runtime.name="Java for Android" yet it also has no effect

PR #1275, opened 2025-05-22
enhancement: android stuff

alexytomi:
This addresses #1210 and changes the values the tests use to be more accurate to real world.

I mocked up a crappy little android app to get the values. It's set to target API 14 (Android 4.0.1/2 Ice Cream Sandwich, getting it down to an even lower version is a headache.

image


src/main/java/org/sqlite/util/OSInfo.java
```diff
@@ -114,7 +115,18 @@ public static String getOSName() {
...
+        // see https://developer.android.com/ndk/guides/stable_apis#graphics
+        // We don't use libc/libm/libdl because that has changed what directory its pointing to and
+        // OEMs implement the symlink that allows backwards compatibility
+        // for apps that use the old path differently, which may cause this check to fail because
+        // common undocumented behaviour. See
+        // https://developer.android.com/about/versions/10/behavior-changes-all#bionic
+        File androidGLES = new File("/system/lib/libGLESv1_CM");
+        File android64GLES = new File("/system/lib64/libGLESv1_CM");
+
+        boolean isRunningAndroid = android64GLES.exists() || androidGLES.exists();
```
gotson, line 128: would be worth extracting this in a separate method, the same way we already have `isAndroidRuntime` and `isAndroidTermux`

src/test/java/org/sqlite/util/OSInfoTest.java
```diff
@@ -133,8 +133,8 @@ public void testIsAndroidTermux() throws Exception {
     }
 
     @Nested
-    @SetSystemProperty(key = "java.runtime.name", value = "Java for Android")
-    @SetSystemProperty(key = "os.name", value = "Linux for Android")
+    @SetSystemProperty(key = "java.runtime.name", value = "Android Runtime")
+    @SetSystemProperty(key = "os.name", value = "Linux")
```
gotson, line 137: what is the reason for this change ?

alexytomi: It's the values that android returns for those properties on android apps

src/main/java/org/sqlite/util/OSInfo.java
alexytomi: Done! :D

gotson:
thanks for the PR

PR #1286, opened 2025-06-11
Update JDBC3DatabaseMetaData to enable supported features

newdenim:
Enable `supportsAlterTableWithDropColumn` and `supportsAlterTableWithAddColumn` that should be already supported by the driver.

These are both supported from 3.35.0, so the metadata should return `true`, so tools that depend on it it can work correctly. [(discussion](https://github.com/xerial/sqlite-jdbc/discussions/1285))

*NOTE*: I did not check if other properties from the metadata are wrong.

Let me know if you have any comments.
Thanks!


gotson:
thanks !

PR #1301, opened 2025-07-08
Support JDBC4 CLOB and BLOB in PreparedStatement parameters

timothyjward:
There are TODO markers in the JDBC4 Prepared Statement implementation relating to BLOB and CLOB data. These can easily be implemented by delegating to the existing JDBC implementations. This may not be optimised, but it will allow the use of CLOB and BLOB, which is a significant improvement.


gotson:
thanks for the PR!

PR #1311, opened 2025-08-14
chore(ci): use native `jq` functionality to construct matrix

fearphage:
* Switched to using `jq`'s native functionality for constructing objects and arrays
* Replaced multiple calls to `sed` with a single `awk` call


fearphage:
@xerial @gotson Bump.

Issue #1318, opened 2025-08-28
DELETE ... LIMIT support does not seem to be enabled


sandrinr:
**Describe the bug**

SQLite has `DELETE ... LIMIT` support. However, the support needs to be enabled during compile time. The SQLite binary shipped with sqlite-jdbc and used on a x64_64 Linux setup, does not seem to have this support enabled.

**To Reproduce**

Execute:

```
DELETE FROM t LIMIT 1;
```

through sqlite-jdbc.

**Expected behavior**

SQLite does not return with an error and executes the delete operation as instructed.

**Logs**

SQLite reports the error:

```
[SQLITE_ERROR] SQL error or missing database (near "LIMIT": syntax error)
```

**Environment (please complete the following information):**
 - OS: Linux
 - CPU architecture: x86_64
 - sqlite-jdbc version: 3.50.2.0

**Additional context**

Using the same SQL statement with the `sqlite3` command shipped with Ubuntu 22.04 works as expected.

gotson:
if you want to submit a PR we can have a look at it

sandrinr:
I looked into it and found that changing this is not going to be trivial. I expected it to be more of a "flipping a switch" thing 😉 .

sqlite-jdbc uses the amalgamation build of SQLite. This build comes with the parser already generated. However, the switch to enable `DELETE ... LIMIT` support, `SQLITE_ENABLE_UPDATE_DELETE_LIMIT`, is situated in the parser source files. To have it enabled, sqlite-jdbc would need to employ the toolchain (especially the parser generator) required to build the full source to generate an amalgamation that includes a parser with this support enabled.

While this is certainly doable, it would make the sqlite-jdbc build more complicated which might not be desirable. Also, there is an SQL-standard-conform workaround, `DELETE ... WHERE rowid IN (SELECT rowid ... LIMIT ...)`, to achieve the same behavior.

Feel free to close this.

gotson:
indeed it seems the discussion was raised on SQLite forums as far as 4 years ago, but still not solution.

kohlschuetter:
I just ran into this as well, and I hope this can trivially be fixed.

On the Apple-provided sqlite3 build in macOS (at least on macOS 15.7 with sqlite3 version 3.43.2 2023-10-10), the `SQLITE_ENABLE_UPDATE_DELETE_LIMIT` feature is enabled. It's therefore very unexpected to find the feature working when testing in the command line, and failing when using the JDBC version.

[...]

PR #1325, opened 2025-09-18
Enable LIMIT support for UPDATE and DELETE

kohlschuetter:
Prominent vendors release sqlite3 with SQLITE_ENABLE_UPDATE_DELETE_LIMIT enabled, which adds support for a LIMIT clause to UPDATE and DELETE.

The feature can currently only be enabled by building the sqlite3 parser from source. However, sqlite3-jdbc currently builds from the amalgamation file.

Add support to the Makefile to create the amalgamation zip from the source zip, enabling support for this feature.

The feature is now on by default. It can be disabled by building with "make ENABLE_UPDATE_DELETE_LIMIT=0"

Fixes: https://github.com/xerial/sqlite-jdbc/issues/1318


gotson:
there's an error when building the native libs in CI: https://github.com/xerial/sqlite-jdbc/actions/runs/17847931008/job/50750975762?pr=1325

would you be able to have a look ?

kohlschuetter:
@gotson done. Apparently the shells in your GitHub actions don't support curly brace expansion. (Note: the unrelated/unchanged target "clean-tests" also contains curly braces and would therefore also fail)

There's another, probably unrelated build failure on Android; you may want to take a look there.

kohlschuetter:
The unrelated failure occurs on several platforms; also see https://github.com/xerial/sqlite-jdbc/actions/runs/17860264461/job/50788907525 (which is a `/native` run based off master without these changes)

gotson:
> Apparently the shells in your GitHub actions don't support curly brace expansion. (Note: the unrelated/unchanged target "clean-tests" also contains curly braces and would therefore also fail)

Thanks. Honestly not sure what version of bash is used on GHA, it's weird it's not supported. `clean-tests` is never used in CI so that never came up.



> There's another, probably unrelated build failure on Android; you may want to take a look there.

I'll have a look. Since we rely on Dockcross to build native libs, it happens that the Dockcross images changed since the last successful build, and could introduce some issues.

Once solved we can rebase your PR.

gotson:
Thanks a lot for the PR!

PR #1406, opened 2026-04-27
chore: remove execute bit on non-executable files

michael-o:
Those files don't need an execute bit, remove it.


gotson:
Hi @michael-o the dll are generated at every version update so that will come back

michael-o:
> Hi @michael-o the dll are generated at every version update so that will come back

I already thought so, although it makes no sense since DLLs don't need an execute bit. Let me revert that.

michael-o:
Execute bit on DDLs removed.

Issue #1403, opened 2026-04-17
Custom PRAGMA names support


h8leet:
Is it possible to apply PRAGMA with custom name?

For example, I need apply `PRAGMA plaintext_header_size = 32` which is available in SQlite3 from [SQLite3MultipleCiphers](https://github.com/utelle/SQLite3MultipleCiphers) project.

Existsing values of enum SQLiteConfig.Pragma doesn't contains PLAINTEXT_HEADER_SIZE.

The problem is that this PRAGMA is applied only when opening a database and its value cannot be changed later, so it is not possible to set it by a SQL query after establishing a connection. There are also other PRAGMA ​​in SQLite3MultipleCiphers that are not in sqlite-jdbc and they are not needed for plain SQLite3.

gotson:
> The problem is that this PRAGMA is applied only when opening a database and its value cannot be changed later, so it is not possible to set it by a SQL query after establishing a connection.

Could you clarify on that part ?

PRAGMA are always called using the `PRAGMA` instruction, and this requires a connection.

gotson:
By the way, did you check https://github.com/Willena/sqlite-jdbc-crypt ?

h8leet:
> Could you clarify on that part ?

My bad, It means that `PRAGMA plaintext_header_size` must be executed before any SQL statements are executed. As I understand it, the trigger for creating a database with the `plaintext_header_size` parameter is the execution of any SQL, as is PRAGMAs cipher and key (https://utelle.github.io/SQLite3MultipleCiphers/docs/configuration/config_sql_pragmas/#key-handling).

> By the way, did you check https://github.com/Willena/sqlite-jdbc-crypt ?

Yes, sqlite-jdbc-crypt has an extended set of PRAGMAs including `plaintext_header_size`. But at the moment the sqlite-jdbc-crypt is supported with a large delay due to the maintainer don't have a lot of spare time it currently (https://github.com/Willena/sqlite-jdbc-crypt/issues/158#issuecomment-3246909154)

Also after further analysis I think that sqlite-jdbc-crypt's additional PRAGMA won't help me. If `Pragma#PASSWORD` is used, the query "select 1 from sqlite_schema" is implicitly executed: https://github.com/xerial/sqlite-jdbc/blob/4cd221a72480a681f91f78f9d8a99372daf423ce/src/main/java/org/sqlite/SQLiteConfig.java#L208
[...]

PR #1413, opened 2026-04-28
feat: custom PRAGMA names support

h8leet:
Closes the issue https://github.com/xerial/sqlite-jdbc/issues/1403


gotson:
please explain your change

h8leet:
> please explain your change

After testing the changes, it became clear that to fully support custom PRAGMAs, it was also necessary to be able to customize the order in which PRAGMAs are applied. I'll explain the changes made after the necessary additional changes.

PR #1418, opened 2026-05-19
Upgrade Java 11 to Java 17

GG-Feng:
﻿## Summary

This PR upgrades the project from Java 11 to Java 17, applying modern Java language features and updating the build configuration accordingly.

## Changes

- Updated source/target from 11 to 17
- Applied instanceof pattern matching
- Added @Serial annotation to serialVersionUID fields
- Converted String.format to .formatted
- Replaced Paths.get with Path.of
- Replaced new Integer/Long with valueOf
- Converted multiline strings to text blocks
- Updated CI workflow JDK version

## Statistics

26 files changed, 116 insertions, 125 deletions

## Verification

- BUILD SUCCESS with JDK 17

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
       - uses: actions/setup-java@v5
         with:
           distribution: 'temurin'
-          java-version: 11
+          java-version: 17
           java-package: jdk
           cache: 'maven'
       - name: spotless:check
@@ -53,7 +53,7 @@ jobs:
       - uses: actions/setup-java@v5
         with:
           distribution: 'temurin'
-          java-version: 11
+          java-version: 17
           java-package: jdk
           cache: 'maven'
       - name: Download and install signatures jar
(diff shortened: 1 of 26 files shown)


gotson:
The project targets Java 8. Java 11 is only used for the Spotless plugin.

There are no plans to target more recent Java versions, Java 8 will remain the target.

GG-Feng:
Thanks for the clarification.
Apologies for the misunderstanding caused by my eagerness earlier — I did not mean that sqlite-jdbc itself should officially drop Java 8 compatibility or change its current baseline policy.
What we are actually exploring is a parallel JDK adaptation and maintenance effort.

The idea is roughly:
1：the upstream project can continue normal development on its current Java baseline
2：while we maintain separate adaptation branches for newer JDK versions
3：mainly for users who want to run/build in newer Java ecosystems

We noticed there were already some discussions in the community around newer JDK compatibility, so we started working on JDK upgrade adaptations for several OSS projects and collecting community feedback.

If there is sustained community interest, we are willing to help maintain and support 2–3 upgraded JDK variant branches long‑term, without requiring the main upstream branch to change its compatibility target.

So this PR was intended more as an exploration of newer‑JDK adaptation feasibility and optional maintenance branches, rather than a proposal to force the upstream project to raise its official Java baseline.

We can open a dedicated issue for further discussion if you prefer.

Sorry again for the confusion！

gotson:
> We can open a dedicated issue for further discussion if you prefer.

please do so yes

PR #1425, opened 2026-06-12
fix(jdbc): escape table name in getColumns metadata query

dxbjavid:
getColumns reads each matching table name back from sqlite_schema (through getTables) and builds an internal query from it, but the table name is concatenated straight into a single-quoted literal for the tblname column while every other identifier in the same loop already goes through escape(). A table whose name contains a single quote is a perfectly valid SQLite identifier, so the generated query ends up malformed and the embedded text is treated as SQL, which means a crafted table name can break or inject into the metadata query. I have wrapped the value in escape() so the quote is doubled like the surrounding columns. Added a small regression test under DBMetaDataTest that creates a table with a quote in its name and reads its columns back.


gotson:
thanks for the PR

dxbjavid:
Thanks for the review and merge! Appreciated

PR #1426, opened 2026-06-13
fix(jdbc): escape table type values in getTables metadata query

dxbjavid:
the types argument passed to getTables is interpolated straight into the TABLE_TYPE IN (...) list without escape(), unlike the other identifiers used in these metadata queries, so a type value containing a single quote breaks the generated SQL and can change what the query returns (for instance a value like "X') OR ('1'='1" bypasses the type filter and lists every table). passing each value through escape() keeps embedded quotes literal, in line with the recent getColumns change. i added a small regression test next to the existing metadata quote tests.


gotson:
thanks for the PR!

dxbjavid:
Thanks for the review and merge. Appreciated

PR #1427, opened 2026-06-15
fix(jdbc): escape catalog and schema in foreign key metadata queries

dxbjavid:
getCrossReference, getExportedKeys and getImportedKeys put the caller-supplied catalog, schema and table arguments straight into their metadata queries through quote(), which only wraps the value in single quotes and does no escaping, unlike the escape() helper used for the other identifiers nearby. So a catalog or schema containing a single quote (a perfectly legal argument to these DatabaseMetaData calls) breaks out of the string literal and the generated query fails or can be steered, which is a first-order SQL injection through the metadata layer. The fix routes each of these values through escape() before quote() and makes escape() null-safe so it slots in next to the existing null handling. I have added regression tests for the three methods using a single-quote in the catalog and schema.


gotson:
thanks for all those escape PRs!

dxbjavid:
Thank you for the review and merge. Much appreciated

PR #1430, opened 2026-06-17
fix(jdbc): escape resolved table name in getExportedKeys query

dxbjavid:
getExportedKeys builds its PKTABLE_NAME from the table name it reads back out of sqlite_schema, the case-correct form stored when the table was created, and that value goes into the query through quote() on its own without escape(). Every other identifier in these foreign-key queries already passes through escape(), and the recent catalog/schema change covered the caller-supplied arguments, but this resolved name was left raw. A table created with a single quote in its name (a legal quoted identifier) then breaks out of the literal when getExportedKeys is called for it, so the generated metadata query either errors or can be steered, the same second-order injection getColumns had. I noticed it while going back over the remaining quote() calls in the file after the earlier fixes landed. Routing target through escape() before quote() closes it and lines up with how the surrounding values are handled. I have added a regression test that creates a quoted-name parent table with a child foreign key pointing at it.

### DIFF
--- a/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java
@@ -1349,7 +1349,7 @@ public abstract class JDBC3DatabaseMetaData extends CoreDatabaseMetaData {
                 .append(" as PKTABLE_CAT, ")
                 .append(schema)
                 .append(" as PKTABLE_SCHEM, ")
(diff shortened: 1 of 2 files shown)


gotson:
thanks for the PR, can you check the code formatting that is failing ?

dxbjavid:
good catch, that was a comment line spilling past the line limit. wrapped it so spotless:check is happy now.

dxbjavid:
thanks for the merge

Issue #821, opened 2022-12-20
Possible race condition in SQLitePooledConnection


nicolatimeus:
Hi all,

I've noticed that at https://github.com/xerial/sqlite-jdbc/blob/3904e83bf0b2e9d9ab2019f8c8bf49f567fb74ac/src/main/java/org/sqlite/javax/SQLitePooledConnection.java#L106

the `ConnectionEventListener.connectionClosed()` method is called and then the subsequent lines may perform modifications on the released connection:

https://github.com/xerial/sqlite-jdbc/blob/3904e83bf0b2e9d9ab2019f8c8bf49f567fb74ac/src/main/java/org/sqlite/javax/SQLitePooledConnection.java#L110

https://github.com/xerial/sqlite-jdbc/blob/3904e83bf0b2e9d9ab2019f8c8bf49f567fb74ac/src/main/java/org/sqlite/javax/SQLitePooledConnection.java#L112

Should these modifications be moved before the invocation of `ConnectionEventListener.connectionClosed()`? Otherwise the client code may start reusing the released connection concurrently while they are being performed.

gotson:
Would you be able to provide an example project or unit test that would highlight the problem maybe ?

nicolatimeus:
Sure, I have created the following project:

https://github.com/nicolatimeus/sqlite-test/blob/main/src/main/java/com/github/nicolatimeus/App.java

On my machine execution stops after a few iterations with exceptions like the following:

```
org.sqlite.SQLiteException: [SQLITE_ERROR] SQL error or missing database (cannot rollback - no transaction is active)
        at org.sqlite.core.DB.newSQLException(DB.java:1135)
        at org.sqlite.core.DB.newSQLException(DB.java:1146)
        at org.sqlite.core.DB.throwex(DB.java:1106)
        at org.sqlite.core.DB.exec(DB.java:198)
        at org.sqlite.SQLiteConnection.rollback(SQLiteConnection.java:455)
        at org.sqlite.javax.SQLitePooledConnection$1.invoke(SQLitePooledConnection.java:110)
        at com.sun.proxy.$Proxy0.close(Unknown Source)
        at com.github.nicolatimeus.App.worker(App.java:71)
        at com.github.nicolatimeus.App.lambda$0(App.java:94)
        at java.lang.Thread.run(Thread.java:748)
```

gotson:
[...]

PR #1435, opened 2026-07-03
fix(pool): reset connection before notifying listeners on close

dxbjavid:
closing a pooled handle currently notifies the ConnectionEventListeners (which returns the connection to the pool) before it rolls back and resets the physical connection to auto-commit, so under concurrency another thread can borrow the pooled connection and start using the same physical connection while that reset is still in flight, which intermittently throws "cannot rollback - no transaction is active" as reported in #821. this reorders the close handler to reset the physical connection and mark the handle closed before notifying listeners, and makes close idempotent so the second close triggered when getConnection() reuses the handle does not reset again. added a small multi-threaded regression test that fails on master and passes with the change.


gotson:
thanks a lot for the PR!

dxbjavid:
TY!

PR #1440, opened 2026-07-18
fix(jdbc): map DATETIME column type to Types.TIMESTAMP

### LINKED ISSUE #1436: Inconsistent DataType returned for DateTime
**Describe the bug**
in JDBC3ResultSet#getColumnType (line 700), a column of valueType = SQLITE_TEXT and typeName = DATETIME returns Types.DATE, not Types.TIMESTAMP. Is it intentional ?

**To Reproduce**
CREATE TABLE mytable ( "fid" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, "date_creation" DATETIME);
ResultSet rs = st.executeQuery("SELECT * FROM mytable"))
rsmd = rs.getMetaData();
rsmd.getColumnType(2);
=> jdbcType=91 (DATE)

**Expected behavior**

**Logs**

 - sqlite-jdbc version 3.53.0.0

**Additional context**

kamilkrzywanski:
`ResultSetMetaData.getColumnType()` was mapping declared `DATETIME` columns to `Types.DATE`, same as plain `DATE`. That looks wrong: a `DATETIME` value can include a time component (`2024-01-15 13:45:30.123`), and `TIMESTAMP` already returned `Types.TIMESTAMP`. Clients that pick `getDate()` vs `getTimestamp()` from the reported JDBC type end up dropping the time.

I changed both the integer and text branches in `JDBC3ResultSet#getColumnType` so `DATETIME` is reported as `Types.TIMESTAMP`, and left `DATE` alone. Updated the existing type matrix assertion and added a regression test covering the empty-result case from #1436, text and integer storage, and DATE/DATETIME/TIMESTAMP side by side.

Fixes #1436

### DIFF
--- a/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
+++ b/src/main/java/org/sqlite/jdbc3/JDBC3ResultSet.java
@@ -642,11 +642,11 @@ public abstract class JDBC3ResultSet extends CoreResultSet {
                 return Types.BIGINT;
             }
 
-            if ("DATE".equals(typeName) || "DATETIME".equals(typeName)) {
+            if ("DATE".equals(typeName)) {
                 return Types.DATE;
             }
 
-            if ("TIMESTAMP".equals(typeName)) {
+            if ("DATETIME".equals(typeName) || "TIMESTAMP".equals(typeName)) {
                 return Types.TIMESTAMP;
             }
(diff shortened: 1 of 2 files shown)


gotson:
nice work, thanks!

PR #1441, opened 2026-07-18
fix(jdbc): throw SQLException from createConnection on invalid URL

### LINKED ISSUE #1374: NativeLibraryManager.createConnection silently returns null on invalid URL instead of throwing SQLException
**Describe the bug**
The static method `org.sqlite.JDBC.createConnection(String url, Properties prop)` now silently returns `null` when the URL protocol is not handled by the driver (e.g., not starting with `jdbc:sqlite:`), instead of throwing `SQLException` with the message "invalid database address" as it did before.

This change was introduced in commit 392b7e7d9bf1babf840e175979b2319400f050ac ("return null if unhandled url protocol").

While returning `null` from `Driver.connect()` is correct JDBC behavior (to allow `DriverManager` to try other drivers), applying the same logic to the **static convenience method** `JDBC.createConnection()` is a breaking change. Code that calls this method directly expects either a valid `Connection` or an exception — not a silent `null`.

As a result, libraries and applications that use `JDBC.createConnection` (common with SQLite to avoid `DriverManager` overhead) now get `NullPointerException`s deep in their connection-handling code.

**To Reproduce**

Minimal pure-Java example (only sqlite-jdbc dependency required):

```java
import org.sqlite.JDBC;
import java.sql.Connection;
import java.util.Properties;

public class Repro {
    public static void main(String[] args) throws Exception {
        String url = "jdbc:wrongprotocol:test.db";  // Invalid protocol
        Properties prop = new Properties();
        Connection conn = JDBC.createConnection(url, prop);
        // conn == null
        conn.close();  // → NullPointerException
    }
}
```

Real-world example from the ScalaSql library (https://github.com/com-lihaoyi/scalasql):

```scala
private def withConnection[T](f: DbClient.Connection => T): T = {
  val connection = dataSource.getConnection
  try f(new DbClient.Connection(connection, config, listeners))
  finally connection.close()  // ← NPE if connection is null
}
```

If the internal `DataSource` uses `JDBC.createConnection` (common pattern for SQLite), an invalid URL now causes NPE instead of early SQLException.

**Expected behavior**
- The static `JDBC.createConnection` should continue to throw `SQLException` on invalid URLs (original behavior), as it is a convenience method intended for direct use with SQLite URLs.
- Alternatively, if returning `null` is kept, it must be explicitly documented in Javadoc, and a warning should be logged. Existing callers should not break silently.

The method's contract (`throws SQLException`) suggests errors are reported via exceptions, not `null` returns.

**Logs**
Typical stack trace:

```
Caused by: java.lang.NullPointerException: Cannot invoke "java.sql.Connection.close()" because "connection" is null
    at scalasql.core.DbClient$DataSource.withConnection(DbClient.scala:136)
    ...
```

**Environment**
- OS: Any
- CPU architecture: Any
- sqlite-jdbc version: All versions after commit 392b7e7 (including latest 3.51.1.0)

**Additional context**
The change improves compatibility when using `DriverManager.getConnection` with multiple drivers registered, but it breaks direct callers of `JDBC.createConnection`. Popular Scala libraries like ScalaSql (com-lihaoyi/scalasql) are affected because they follow the common pattern of calling the driver's static create method internally.

No documentation or changelog entry mentions this behavioral change or the new possibility of `null` return.

Suggested fixes:
1. Restore throwing `SQLException` in `JDBC.createConnection` (preferred, non-breaking for existing code).
2. Or add clear Javadoc + deprecation note + recommend switching to `DriverManager`.

Thanks for maintaining the driver!

kamilkrzywanski:
`JDBC.createConnection()` has been returning `null` for URLs that are not `jdbc:sqlite:` ever since 392b7e7. That is the right behaviour for `Driver.connect()` so `DriverManager` can walk the registered drivers, but the static helper is used directly by callers (and by things like `SQLiteDataSource` / ScalaSql) that expect either a real connection or an exception. They currently fall over later with an NPE instead.

I split the two paths: `connect()` still returns `null` when the protocol is not ours, and `createConnection()` throws `SQLException("invalid database address: …")` again like it did before that change. Javadoc on the static method notes the difference so we do not glue them back together by accident.

Updated the old test that asserted `null` from `createConnection`, and added coverage for `Driver.connect` returning `null`, null URL, and a valid `jdbc:sqlite:` URL.

Fixes #1374

### DIFF
--- a/src/main/java/org/sqlite/JDBC.java
+++ b/src/main/java/org/sqlite/JDBC.java
@@ -76,6 +76,9 @@ public class JDBC implements Driver {
 
     /** @see java.sql.Driver#connect(java.lang.String, java.util.Properties) */
     public Connection connect(String url, Properties info) throws SQLException {
+        if (!isValidURL(url)) {
+            return null;
+        }
         return createConnection(url, info);
     }
 
@@ -92,15 +95,20 @@ public class JDBC implements Driver {
     /**
      * Creates a new database connection to a given URL.
      *
+     * <p>Unlike {@link #connect(String, Properties)}, this method throws {@link SQLException} when
+     * the URL is not a {@code jdbc:sqlite:} address, rather than returning {@code null}.
+     *
      * @param url the URL
      * @param prop the properties
      * @return a Connection object that represents a connection to the URL
-     * @throws SQLException
-     * @see java.sql.Driver#connect(java.lang.String, java.util.Properties)
+     * @throws SQLException if the URL is not a valid SQLite JDBC URL, or if a database access error
+     *     occurs
      */
     public static SQLiteConnection createConnection(String url, Properties prop)
             throws SQLException {
-        if (!isValidURL(url)) return null;
+        if (!isValidURL(url)) {
+            throw new SQLException("invalid database address: " + url);
+        }
 
         url = url.trim();
         return new JDBC4Connection(url, extractAddress(url), prop);
--- a/src/test/java/org/sqlite/JDBCTest.java
+++ b/src/test/java/org/sqlite/JDBCTest.java
@@ -48,8 +48,32 @@ public class JDBCTest {
     }
 
     @Test
-    public void shouldReturnNullIfProtocolUnhandled() throws Exception {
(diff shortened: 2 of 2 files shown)


gotson:
thanks for the PR!

PR #1449, opened 2026-08-06
Fix error in docs

donalmurtagh:


gotson:
this is not an error, the jar only contains natives.

PR #1454, opened 2026-08-19
fix: bound utf-8 reader to the terminating nul in extension-functions

dxbjavid:
**out-of-bounds read on a truncated utf-8 sequence in extension-functions**

the multibyte reader behind reverse/proper/charindex/leftstr/rightstr/strfilter/difference reads the trailing bytes a lead byte announces without checking for the string's terminating nul, so a value ending in a truncated sequence such as `reverse(cast(x'f0' as text))` makes it walk a few bytes past the allocation, a heap over-read reachable from ordinary sql. this stops the read as soon as the next byte is not a continuation byte, so a truncated character resolves to U+FFFD and well-formed utf-8 decodes exactly as before.

### DIFF
--- a/src/main/ext/extension-functions.c
+++ b/src/main/ext/extension-functions.c
@@ -286,9 +286,12 @@ static const int utf_mask[] = {
   xtra = xtra_utf8_bytes[c];                           \
   switch( xtra ){                                      \
     case 4: c = (int)0xFFFD; break;                    \
-    case 3: c = (c<<6) + *(zIn)++;                     \
-    case 2: c = (c<<6) + *(zIn)++;                     \
-    case 1: c = (c<<6) + *(zIn)++;                     \
+    case 3: if((*(zIn)&0xc0)!=0x80){c=0xFFFD;break;}   \
+            c = (c<<6) + *(zIn)++;                     \
+    case 2: if((*(zIn)&0xc0)!=0x80){c=0xFFFD;break;}   \
+            c = (c<<6) + *(zIn)++;                     \
+    case 1: if((*(zIn)&0xc0)!=0x80){c=0xFFFD;break;}   \
+            c = (c<<6) + *(zIn)++;                     \
     c -= xtra_utf8_bits[xtra];                         \
     if( (utf_mask[xtra]&c)==0                          \
         || (c&0xFFFFF800)==0xD800                      \


gotson:
The extension functions come from the Sqlite repo, would you happen to know if that issue is also present there, or whether it has been fixed and this fix is the port of theirs ?

dxbjavid:
good question. this isn't part of the core sqlite amalgamation. it's the contributed extension-functions.c (the LMH one from sqlite.org's contrib page), and the reader here was copied out of sqlite 3.3.13's internal utf8 handling, as the comment just above the tables notes.

core sqlite reworked that reader a long time ago. their version now takes an end pointer and stops on the first non-continuation byte, so the over-read isn't present upstream. the contrib file was never kept in sync, so the old unbounded copy is what still ships here. there's no upstream patch to this file to port, so this is a local fix, but it deliberately follows the same continuation-byte approach current sqlite uses.

gotson:
would it make more sense to try to update that file from upstream instead of patching ?

dxbjavid:
i looked at that first, but there isn't really a newer upstream to pull from for this one. extension-functions.c isn't versioned like the amalgamation is, it's a static file on sqlite.org's contrib page, and the copy there still carries the same unbounded reader from 3.3.13. so re-fetching it wouldn't bring the fix in.

the only way to get it "from upstream" would be to lift the current reader out of core sqlite, but that one has a different shape: it takes an end pointer and is wired through sqlite's own Utf8 machinery. porting it would mean rewriting the macro and touching every call site here, which is a much bigger and riskier change for the same result.

this patch stays inside the existing file and follows the same continuation-byte logic core sqlite settled on, so well-formed utf-8 decodes exactly as before and only the truncated case changes. that felt like the safer, more contained option, but if you'd rather i attempt the fuller resync i can look into it, just flagging it's considerably more invasive.

gotson:
> it's a static file on sqlite.org's contrib page, and the copy there still carries the same unbounded reader from 3.3.13. so re-fetching it wouldn't bring the fix in.

oh, i misunderstood your previous message then. I thought the upstream had a fix.

gotson:
i'm ok with the change, however i was wondering if there is any chance we could write a unit test that would fail this scenario, before the fix ?

dxbjavid:
added one. it runs each of the affected functions (reverse, proper, leftstr, rightstr, charindex, strfilter, difference) over values that end in a truncated multibyte sequence, one case per trailing-byte count, plus a well-formed multibyte string to confirm normal decoding is untouched.

one honest caveat though: the bug is a read of only a few bytes past the buffer, and a read like that doesn't reliably fault in an ordinary build, it just returns whatever happened to be adjacent. so against the unpatched reader the test fails deterministically only when the native lib is built with -fsanitize=address, which is how i originally caught it; in a plain build the assertions pass either way. it still works as a regression guard that exercises every call site and pins the U+FFFD behaviour, but i didn't want to imply it hard-fails without a sanitiser. happy to share the standalone asan harness i used if that'd be more useful.

gotson:
thanks !

dxbjavid:
TY!

